PR #1, opened 2022-03-15
kiagnose: Introduce basic skeleton

EdDev:
Introducing the project skeleton, including a placeholder for the library entry-point, the script to run the tests and CI to run them on every PR.

### DIFF
--- a/automation/make.sh
+++ b/automation/make.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+#
+# This file is part of the kiagnose project
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Copyright 2022 Red Hat, Inc.
+#
+
+set -e
+
+options=$(getopt --options "" \
+    --long unit-test,help\
+    -- "${@}")
+eval set -- "$options"
+while true; do
+    case "$1" in
+    --unit-test)
+        OPT_UNIT_TEST=1
+        ;;
+    --help)
+        set +x
+        echo "$0 [--unit-test]"
+        exit
+        ;;
+    --)
+        shift
+        break
+        ;;
+    esac
+    shift
+done
+
+if [ -z "${OPT_UNIT_TEST}" ]; then
+    OPT_UNIT_TEST=1
+fi
+
+if [ -n "${OPT_UNIT_TEST}" ]; then
+    go test -v ./kiagnose/...
+fi
--- a/kiagnose/mainflow_test.go
+++ b/kiagnose/mainflow_test.go
@@ -1,0 +1,34 @@
+/*
+ * This file is part of the kiagnose project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright 2022 Red Hat, Inc.
+ *
+ */
+
+package kiagnose_test
+
+import (
+	"testing"
+
+	"github.com/kiagnose/kiagnose/kiagnose"
+)
+
+func TestRun(t *testing.T) {
+    t.Run("kiagnose entry point (placeholder test)", func(t *testing.T) {
+        if err := kiagnose.Run(); err != nil {
+	    t.Errorf("failed execution of Run(): %v", err)
+	}
+    })
+}
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -9,12 +9,19 @@ on:
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
+env:
+  GO_VERSION: 1.17
+
 jobs:
-  noop:
-    name: noop 
+  unit-test:
+    name: Unit Test
     runs-on: ubuntu-latest
     steps:
     - name: Check out code
       uses: actions/checkout@v2
-    - name: Hello Word
-      run: echo "Hello World"
+    - name: Setup Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: ${{ env.GO_VERSION }}
+    - name: Run unit tests
+      run: ./automation/make.sh --unit-test
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,3 @@
+module github.com/kiagnose/kiagnose
+
+go 1.17
--- a/kiagnose/mainflow.go
+++ b/kiagnose/mainflow.go
@@ -1,0 +1,24 @@
+/*
+ * This file is part of the kiagnose project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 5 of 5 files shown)


kiagnose/mainflow_test.go
```diff
@@ -0,0 +1,34 @@
...
+package kiagnose_test
+
+import (
+	"testing"
+
+	"github.com/kiagnose/kiagnose/kiagnose"
+)
+
+func TestRun(t *testing.T) {
+    t.Run("kiagnose entry point (placehoder test)", func(t *testing.T) {
```
orelmisan, line 29: *placeholder

automation/make.sh
```diff
@@ -0,0 +1,49 @@
...
+    shift
+done
+
+if [ -z "${OPT_UTEST}" ]; then
+    OPT_UTEST=1
+fi
+
+if [ -n "${OPT_UTEST}" ]; then
+    go test -v ./kiagnose/...
+fi
```
orelmisan, line 49: AFAIU, this logic runs the unit tests even if the script was not run with the `--unit-test` command line arg.

orelmisan (commented):
Thank you for the PR.
Please see my inline comments.

automation/make.sh
EdDev: Yes, it is mentioned in the commit message.

orelmisan: So if the flag has no meaning, why have it?

EdDev: See [here](https://github.com/kiagnose/kiagnose/pull/3/files#diff-543234b4401a6b3d83250efab9463483f32029dc05a0ca02096edc270d281166R46).

EdDev: If no flags are provided, all checks run. This is useful for local execution.

if at least one flag is provided, then only the checks that have been explicitly set are ran.

automation/make.sh
```diff
@@ -0,0 +1,49 @@
...
+set -e
+
+options=$(getopt --options "" \
+    --long unit-test,help\
+    -- "${@}")
+eval set -- "$options"
+while true; do
+    case "$1" in
+    --unit-test)
+        OPT_UTEST=1
```
ormergi, line 28: I would use explicit names at least for the option name - `OPT_UNIT_TESTS`

automation/make.sh
```diff
@@ -0,0 +1,49 @@
+#
```
ormergi, line 1: Missing shebang?

ormergi (commented):
Hi, 
I see what you did there with the `automation/make.sh` script,
would you call it `automation` or `auto` instead? :slightly_smiling_face:

automation/make.sh
orelmisan: Good catch.

ormergi:
Please consider creating a container image that will include the necessary tools for building and testing the project (currently go runtime), it will be nice to have it at early stage of the project.

Here is an example from the PoC, we use alpine-go container for building 
https://github.com/orelmisan/k8s-checkup-framework/blob/main/build/build-image

If its a lot of work we can do it on a follow up PR.

automation/make.sh
EdDev: Sorry, I do not follow. If I set everything by default as enabled, then how will I be able to explicitly enable only one? Two?

lets assume we have 3 options and up.

please write a pseudo code to clarify your suggested logic here.

automation/make.sh
EdDev: I don’t think it is needed, it is pretty clear what it is, and making sentences for flags is an overkill. But it is not worth arguing about, I will change it per your request.

EdDev:
> Hi,
> I see what you did there with the `automation/make.sh` script,
> would you call it `automation` or `auto` instead? 🙂

I fail to understand your request. You are talking about the file name? If so, I think it fits a verb.

EdDev:
> If its a lot of work we can do it on a follow up PR.

You are welcome to propose it in a PR. It has no value imo at this time and I do not know at this stage how to integrate it in the CI.

I think it is fine to expect the developer to have the golang tooling on his dev env.

ormergi:
> I fail to understand your request. You are talking about the file name? If so, I think it fits a verb.

Yeah I meant for the file name.
Maybe `run`? sorry I cant think of a better name..

automation/make.sh
EdDev: Done

EdDev:
[change](https://github.com/kiagnose/kiagnose/compare/2cec39a06fc67d853f348b006aefb225d3f17eba..3f69d772fbf76aa3c00f7fcbdaa615aa87aa7729): Responded to review comments.

ormergi:
> You are welcome to propose it in a PR. It has no value imo at this time and I do not know at this stage how to integrate it in the CI.
> 
> I think it is fine to expect the developer to have the golang tooling on his dev env.

In practice its pretty troublesome when you work on multiple projects that uses different version of go (or any other tool for building and testing).
For example, while working on the PoC, the latency checkup required older version of Go due to one of
its dependencies, while the framework used 1.17, using different versions didn't work properly.

automation/make.sh
orelmisan: As a user of this script, I find the suggested behavior unexpected:
Usually, common tools have a set of default values and command line options which override the default value of a single configuration option.

Regarding my last comment, I don't think we should enable everything by default, just the options that a common developer is likely use most of the time.

EdDev:
> In practice its pretty troublesome when you work on multiple projects that uses different version of go (or any other tool for building and testing). For example, while working on the PoC, the latency checkup required older version of Go due to one of its dependencies, while the framework used 1.17, using different versions didn't work properly.

You can, if you want, to customize your dev env as you wish. You have many options to do this, from [gvm](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&cad=rja&uact=8&ved=2ahUKEwjhrqfh1sj2AhW6SvEDHdLEBQsQFnoECAsQAQ&url=https%3A%2F%2Fgithub.com%2Fmoovweb%2Fgvm&usg=AOvVaw1tQsQacPaf0uGMaGuZhNpf) up to using a local container for each and every project.

I would also find it a problem if you cannot build using one of the latest Go versions.
In any case, I suggest to discuss it on the issue opened for this and not on this PR.

ormergi:
> `make.sh` follows the `Makefile` usages, and it is not worse than `run.sh`.

I understand where the name came from, I am not a fan for calling it after a tool we dont use because we dont need.
what you think about `build`?
If it doesn't work for you dont mind it, its seems like a matter of taste so its not that important :slightly_smiling_face:

ormergi (dismissed):
Thanks for the changes.

EdDev:
[change](https://github.com/kiagnose/kiagnose/compare/3f69d772fbf76aa3c00f7fcbdaa615aa87aa7729..c4a6eb4cda5f1e7113cf75704b4c36fc067f4385): Used a more recent `setup-go` action version (2).

automation/make.sh
orelmisan: I suggest to have the option to enable/disable each option separately, for example:
Lets say you have three features - A, B and C. A and B are enabled by default and C is disabled.
Default case:
```bash
$ ./automation/make.sh
```
A and B will run, C will not.
Second example:
```bash
$ ./automation/make.sh --disable-A
```
B will run, A and C will not.

Third example:
```bash
$ ./automation/make.sh --disable-A --enable-C
```

B and C will run, A will not.

This is one style of options.
We can also do something like:
```bash
$ ./automation/make.sh --mode=ci
```

EdDev: Thank you for the examples.

I think it is complicated compared to boolean flags. You also present a proposal to a need that does not exist at the moment.
On CI I want to run individual checks.
Locally I want either to run all or to run a specific one.

If I set such defaults as you described, I need to remember and document it. I also complicate the the user if he wants to run just one thing.
This may be a good solution if the options evolve to a point were it make sense. I do not see it happening anytime soon.

If you have such a strong opinion on this topic, I suggest you send a PR to suggest your approach. But I would do it when at least 3 options exist.

automation/make.sh
```diff
@@ -0,0 +1,50 @@
...
+    --long unit-test,help\
+    -- "${@}")
+eval set -- "$options"
+while true; do
+    case "$1" in
+    --unit-test)
+        OPT_UNIT_TEST=1
+        ;;
+    --help)
+        set +x
```
orelmisan, line 32: The `set +x` statement is redundant under a `--help` option which should print the usage message and exit.
On the other hand, it could be useful to put it under a `-v` or `--verbose` options, but this can be added in a follow up PR.

automation/make.sh
orelmisan: There is another alternative:
We could mimic the behavior of `make`  by providing a "target" / "targets" as an argument and not as options.
This:
```bash
$ ./automation/make.sh
```

shall be equal to this:
```bash
$ ./automation/make.sh all
```

The user could provide a single target:
```bash
$ ./automation/make.sh unittest
```

Or a list of targets:
```bash
$ ./automation/make.sh unittest lint
```

WDYT?

ormergi: Note that this is exactly what this script do, mimic `make` frontend just so you have single starting point.
I also like "cleaner" names for the "targets" (w/o `--` prefix) .

orelmisan (approved):
I believe that if our script is called make.sh it should behave like `make` does.
Having said that, I don't think that this should hold this PR any longer, because it could be changed later on.
Thank you.

PR #2, opened 2022-03-15
tests: Use testify assertions

EdDev:
[Testify](https://github.com/stretchr/testify) is a lightweight library for assertions in go.
It adds the minimum needed beyond the Go standard testing.

With this change, the go.sum and the vendoring of the dependencies is introduced.

~Depends on #1~


ormergi:
Are there other alternative assertion "suite" we should consider?

orelmisan:
@ormergi There is also [gomega](https://onsi.github.io/gomega/).

orelmisan (approved):

ormergi (approved):
Lets give it a try, thanks for the PR

PR #8, opened 2022-03-20
Introduce mainflow skeleton 

ormergi:
Introduce kiagnose mainflow skeleton for launching and managing
checkups lifecycle (e.g: setup, teardown, result reporting).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,7 @@ module github.com/kiagnose/kiagnose
 
 go 1.17
 
-require github.com/stretchr/testify v1.7.0
+require github.com/stretchr/testify v1.7.1
 
 require (
 	github.com/davecgh/go-spew v1.1.0 // indirect
--- a/go.sum
+++ b/go.sum
@@ -3,8 +3,8 @@ github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSs
 github.com/pmezard/go-difflib v1.0.0 h1:4DBwDE0NGyQoBHbLQYPwSUPoCMWR5BEzIk/f1lZbAQM=
 github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
 github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
-github.com/stretchr/testify v1.7.0 h1:nwc3DEeHmmLAfoZucVR881uASk0Mfjw8xYJ99tb5CcY=
-github.com/stretchr/testify v1.7.0/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
+github.com/stretchr/testify v1.7.1 h1:5TQK59W5E3v0r2duFAb7P95B6hEeOyEnHRa8MjYSMTY=
+github.com/stretchr/testify v1.7.1/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
 gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405 h1:yhCVgyC4o1eVCa2tZl7eS0r+SDo693bJlVdllGtEeKM=
 gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
 gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c h1:dUUwHk2QECo/6vqA44rthZ8ie2QXMNeKRTHCNY2nXvo=
--- a/kiagnose/mainflow.go
+++ b/kiagnose/mainflow.go
@@ -19,6 +19,45 @@
 
 package kiagnose
 
-func Run() error {
+import (
+	"fmt"
+)
+
+type launcher interface {
+	Setup() error
+	Run() error
+	Teardown() error
+	Report() error
+}
+
+func Run(launcher launcher) (runErr error) {
+	defer func() {
+		if reportErr := launcher.Report(); reportErr != nil {
+			if runErr != nil {
+				runErr = fmt.Errorf("%v, %v", runErr, reportErr)
+			} else {
+				runErr = reportErr
+			}
+		}
+	}()
+
+	if err := launcher.Setup(); err != nil {
+		return err
+	}
+
+	defer func() {
+		if teardownErr := launcher.Teardown(); teardownErr != nil {
+			if runErr != nil {
+				runErr = fmt.Errorf("%v, %v", runErr, teardownErr)
+			} else {
+				runErr = teardownErr
+			}
+		}
+	}()
+
+	if err := launcher.Run(); err != nil {
+		return err
+	}
+
 	return nil
 }
--- a/kiagnose/mainflow_test.go
+++ b/kiagnose/mainflow_test.go
@@ -20,6 +20,7 @@
 package kiagnose_test
 
 import (
+	"errors"
 	"testing"
 
 	assert "github.com/stretchr/testify/require"
@@ -28,7 +29,88 @@ import (
 )
 
 func TestRun(t *testing.T) {
-	t.Run("kiagnose entry point (placeholder test)", func(t *testing.T) {
-		assert.NoError(t, kiagnose.Run())
+	t.Run("should run successfully", func(t *testing.T) {
+		assert.NoError(t, kiagnose.Run(launcherStub{}))
 	})
+
+	t.Run("should fail when report is failing", func(t *testing.T) {
+		assert.ErrorContains(t, kiagnose.Run(launcherStub{failReport: errorReport}), errorReport.Error())
+	})
+
+	t.Run("should fail when setup is failing", func(t *testing.T) {
+		assert.ErrorContains(t, kiagnose.Run(launcherStub{failSetup: errorSetup}), errorSetup.Error())
+	})
+
+	t.Run("should fail when setup and report are failing", func(t *testing.T) {
+		err := kiagnose.Run(launcherStub{
+			failSetup:  errorSetup,
+			failReport: errorReport,
+		})
+		assert.ErrorContains(t, err, errorSetup.Error())
+		assert.ErrorContains(t, err, errorReport.Error())
+	})
+
+	t.Run("should fail when run is failing", func(t *testing.T) {
+		assert.ErrorContains(t, kiagnose.Run(launcherStub{failRun: errorRun}), errorRun.Error())
+	})
+
+	t.Run("should fail when teardown is failing", func(t *testing.T) {
+		assert.ErrorContains(t, kiagnose.Run(launcherStub{failTeardown: errorTeardown}), errorTeardown.Error())
+	})
+
+	t.Run("should fail when run and report are failing", func(t *testing.T) {
+		err := kiagnose.Run(launcherStub{
+			failRun:    errorRun,
+			failReport: errorReport})
+		assert.ErrorContains(t, err, errorRun.Error())
+		assert.ErrorContains(t, err, errorReport.Error())
+	})
+
+	t.Run("should fail when teardown and report are failing", func(t *testing.T) {
+		err := kiagnose.Run(launcherStub{
+			failTeardown: errorTeardown,
+			failReport:   errorReport})
+		assert.ErrorContains(t, err, errorTeardown.Error())
+		assert.ErrorContains(t, err, errorReport.Error())
+	})
+
+	t.Run("should fail when run, teardown and report are failing", func(t *testing.T) {
+		err := kiagnose.Run(launcherStub{
+			failRun:      errorRun,
+			failTeardown: errorTeardown,
+			failReport:   errorReport})
+		assert.ErrorContains(t, err, errorRun.Error())
+		assert.ErrorContains(t, err, errorTeardown.Error())
+		assert.ErrorContains(t, err, errorReport.Error())
+	})
+}
+
+var (
+	errorSetup    = errors.New("setup error")
+	errorRun      = errors.New("run error")
+	errorTeardown = errors.New("teardown error")
+	errorReport   = errors.New("report error")
+)
+
+type launcherStub struct {
+	failSetup    error
+	failRun      error
+	failTeardown error
+	failReport   error
+}
+
+func (s launcherStub) Setup() error {
+	return s.failSetup
+}
+
+func (s launcherStub) Run() error {
+	return s.failRun
+}
+
+func (s launcherStub) Teardown() error {
+	return s.failTeardown
+}
+
+func (s launcherStub) Report() error {
+	return s.failReport
 }
--- a/vendor/github.com/stretchr/testify/assert/assertion_compare.go
+++ b/vendor/github.com/stretchr/testify/assert/assertion_compare.go
@@ -3,6 +3,7 @@ package assert
 import (
 	"fmt"
 	"reflect"
+	"time"
 )
 
 type CompareType int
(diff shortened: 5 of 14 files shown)


orelmisan (commented):
Thank you for the PR!

kiagnose/mainflow.go
```diff
@@ -19,6 +19,52 @@
...
+)
+
+type Launcher interface {
+	Setup() error
+	Run() error
+	Teardown() error
+	Report() error
+}
+
+func Run(launcher Launcher) error {
```
EdDev, line 33: I like less the usage of an error stack here, I would prefer to see the errors being wrapped in the regular manner: An error of one step be wrapped by the error of the next one.

In addition, per what I managed to figure out from the code, the `Report()` is expected to be call always, while the `Teardown` is to be called only if it passed the `Setup()`.
While I think it is fine, mimicking a constructor behavior, it is worth noting that it also means that `Setup` is responsible of cleaning up whatever he did already. If it has a large amount of side effects, it may become an issue later on.

But the problem I do see at this stage is that the code does not try to terminate in a control manner when panics are raised. This is true for both the `Teardown` and `Report` calls.

Consider this skeleton:
```
func Run(launcher Launcher) (runerr error) {
    defer func() {
        err := launcher.Report()
        if err != nil or runerr != nil:
            runerr = fmt.Errorf("mainflow failed: %v, %v", runerr, err)
        }
    }()
    if err = launcher.Setup(); err != nil {
        return err
    }
    defer func() { /*call Teardown() like Report() earlier*/ }()
    
    if err := launcher.Run(); err != nil {
        return err
    }
    return nil
}
```

kiagnose/mainflow_test.go
```diff
@@ -28,7 +30,140 @@ import (
...
 
 func TestRun(t *testing.T) {
-	t.Run("kiagnose entry point (placeholder test)", func(t *testing.T) {
-		assert.NoError(t, kiagnose.Run())
+	t.Run("should run successfully", func(t *testing.T) {
+		assert.NoError(t, kiagnose.Run(launcherStub{
+			failSetup:    false,
+			failRun:      false,
+			failTeardown: false,
+			failReport:   false,
```
EdDev, line 38: This is the default, why specify it?

kiagnose/mainflow_test.go
```diff
@@ -28,7 +30,140 @@ import (
...
+		}))
 	})
+	t.Run("should fail when report is failing", func(t *testing.T) {
+		assert.ErrorIs(t, kiagnose.Run(launcherStub{
+			failSetup:    false,
+			failRun:      false,
+			failTeardown: false,
```
EdDev, line 45: same

kiagnose/mainflow_test.go
```diff
@@ -28,7 +30,140 @@ import (
...
+	errorRun      = errors.New("run error")
+	errorTeardown = errors.New("teardown error")
+	errorReport   = errors.New("report error")
+)
+
+type launcherStub struct {
+	failSetup    bool
+	failRun      bool
+	failTeardown bool
+	failReport   bool
```
EdDev, line 140: You can use the errors themselves, saving you an `if` each time...

```
launcherStub{
    failSetup:    errorSetup,
}
```

kiagnose/mainflow.go
```diff
@@ -19,6 +19,28 @@
 
 package kiagnose
 
-func Run() error {
-	return nil
+import (
+	"github.com/kiagnose/kiagnose/pkg/multierror"
+)
+
+type Launcher interface {
```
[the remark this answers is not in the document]
ormergi: Yes, Setup() should cover loading env vars up to creating all entities required for the checkup (w/o starting it).

EdDev (changes requested):
I have tried to simplify the handling and take care of cases where we have panics occurring (the current code does not handle them at all).

kiagnose/mainflow.go
```diff
@@ -19,6 +19,28 @@
...
+	Setup() error
+	Run() error
+	Teardown() error
+	Report() error
```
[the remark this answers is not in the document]
EdDev: What else can we report exactly if it is not the status?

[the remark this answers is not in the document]
ormergi: At this point of time I prefer to have the least of detailed interface, as other requirements may arise
with the implementation (e.g: add logic for the `mainflow`, expose variables, return other variables then `error`).

ormergi (commented):
Resolve code-review comments: 
- gitignore typo
- mainflow unit tests formatting and simplifying stubs

kiagnose/mainflow.go
ormergi: Thanks for the heads up regarding cleaning up after Setup and handling panics, I will do the necessary changes.

Other than that, now that I am looking at it again, I realize that there is a problem with calling to Report() when Setup() is failing.
Setup() should include the logic for initializing the reporter (e.g: load k8s client, load configMap/CRD to update and verify it exist ).
I think its necessary to have another func for initializing the reporter:
```
func Run(launcher Launcher) (runerr error) {
    if err := launcher.Init(); err != nil {
        return err
    }
    defer func() {
        err := launcher.Report()
        if err != nil or runerr != nil:
            runerr = fmt.Errorf("mainflow failed: %v, %v", runerr, err)
        }
    }()
    ...
}
```
WDYT?

EdDev: This will complicate things IMO.
You could just do a noop operation if the setup has failed to initialize the reporter.
This can be done internally or exposed:
- Internally means you do not see it at all, which may be problematic to test (you need to check this). The API remains as before.
- Externally means you have another object (e.g. `Reporter`) and you let the `Setup` set it up:
  `launcher.Setup(report.New())`
  Then `launcher.Report()` will delegate it to the inputted reporter, which may do something or not, depending of its population by the setup.

EdDev: With the internal option, you do not call `Report` if the setup fails, you leave it to the `Setup()` implementation to handle it, together with cleaning up what it manage to do already (i.e. partial setup cleanup is under its responsibility, including reporting).

ormergi:
Changes: resolve code-review [comment](https://github.com/kiagnose/kiagnose/pull/8#discussion_r831219880)

ormergi:
Changes: reduce errors scope exposure

kiagnose/mainflow_test.go
```diff
@@ -28,7 +30,93 @@ import (
...
+	t.Run("should fail when teardown is failing", func(t *testing.T) {
+		assert.ErrorIs(t, kiagnose.Run(launcherStub{failTeardown: errorTeardown}), errorTeardown)
+	})
+
+	t.Run("should fail with wrapped run-error and reflect report error when both are failing", func(t *testing.T) {
+		err := kiagnose.Run(launcherStub{
+			failRun:    errorRun,
+			failReport: errorReport})
+		assert.Regexp(t, regexp.MustCompile(errorRun.Error()), err.Error())
+		assert.Regexp(t, regexp.MustCompile(errorReport.Error()), err.Error())
```
EdDev, line 72: I find this too odd.
It is playing with test to check if something happened, which does not seem to be "strong".

Also, using a regex for a substring inclusion seems an over kill, a reader may think there is something most sophisticated here than it really is.

Lets see out options:
- Use dedicated typed errors and wrap them using `%w`: This is nice for testing, but it also exposes the errors as API/s.
- Do a simple substring check (`strings.Contains`)
- Use the testify dedicated assert (https://pkg.go.dev/github.com/stretchr/testify/assert#ErrorContains).

Can you try the last one?

kiagnose/mainflow_test.go
```diff
@@ -28,7 +30,93 @@ import (
...
+
+	t.Run("should fail when run is failing", func(t *testing.T) {
+		assert.ErrorIs(t, kiagnose.Run(launcherStub{failRun: errorRun}), errorRun)
+	})
+
+	t.Run("should fail when teardown is failing", func(t *testing.T) {
+		assert.ErrorIs(t, kiagnose.Run(launcherStub{failTeardown: errorTeardown}), errorTeardown)
+	})
+
+	t.Run("should fail with wrapped run-error and reflect report error when both are failing", func(t *testing.T) {
```
EdDev, line 67: Lets try to simplify it for the reader: `should fail when run & report is failing`.

kiagnose/mainflow_test.go
```diff
@@ -28,7 +30,93 @@ import (
...
+	t.Run("should fail when setup is failing", func(t *testing.T) {
+		assert.ErrorIs(t,
+			kiagnose.Run(launcherStub{
+				failSetup: errorSetup,
+			}),
+			errorSetup,
+		)
+	})
+
+	t.Run("should fail with wrapped setup-error and reflect report error when both are failing", func(t *testing.T) {
```
EdDev, line 50: Lets try to simplify it for the reader: `should fail when setup & report is failing`.

kiagnose/mainflow_test.go
```diff
@@ -28,7 +30,93 @@ import (
...
+
+	t.Run("should fail with wrapped run-error and reflect report error when both are failing", func(t *testing.T) {
+		err := kiagnose.Run(launcherStub{
+			failRun:    errorRun,
+			failReport: errorReport})
+		assert.Regexp(t, regexp.MustCompile(errorRun.Error()), err.Error())
+		assert.Regexp(t, regexp.MustCompile(errorReport.Error()), err.Error())
+	})
+
+	t.Run("should fail with wrapped teardown-error and reflect report error when both are failing", func(t *testing.T) {
```
EdDev, line 75: Lets try to simplify it for the reader: `should fail when teardown & report is failing`.

kiagnose/mainflow_test.go
```diff
@@ -28,7 +30,93 @@ import (
...
+		err := kiagnose.Run(launcherStub{
+			failTeardown: errorTeardown,
+			failReport:   errorReport})
+		assert.Regexp(t, regexp.MustCompile(errorTeardown.Error()), err.Error())
+		assert.Regexp(t, regexp.MustCompile(errorReport.Error()), err.Error())
+	})
+
+	t.Run("should fail with wrapped run-error and reflect teardown and report errors when each of them is failing", func(t *testing.T) {
```
EdDev, line 83: Lets try to simplify it for the reader: `should fail when run, teardown & report is failing`.

EdDev (changes requested):

kiagnose/mainflow_test.go
ormergi: I also dont like assert on an error content, initially I did tried to wrap the errors 
but `ErrorIs` (which make use of `error.Is`) work in a way that it wont detect more then
one wrapped error, for example:
In case that Run(), Teardown() and Report() were failing:
```
assert.ErrorIs(t, err, runErr) 
assert.ErrorIs(t, err, teardownErr) <------- assertion fails
assert.ErrorIs(t, err, reportErr) <--------- assertion fails
```
Thus, I used other assertions to ensure the info of other errors is there.
Note that the errors content is also controlled by the tests, meaning the test 
does ensure that the flow works as we expect.

I will try to optimize the assertions and get rid of the regex

ormergi: https://go.dev/play/p/NmCcpqK2yaZ

.gitignore
```diff
@@ -13,3 +13,9 @@
 
 # Dependency directories (remove the comment below to include it)
 # vendor/
+
+# Golang
```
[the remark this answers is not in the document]
ormergi: Done

kiagnose/mainflow.go
```diff
@@ -19,6 +19,52 @@
 
 package kiagnose
 
-func Run() error {
-	return nil
+import (
+	"fmt"
+)
+
+type Launcher interface {
```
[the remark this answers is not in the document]
ormergi: No need to expose it, fixed

kiagnose/mainflow_test.go
ormergi: Done

kiagnose/mainflow_test.go
ormergi: Done

kiagnose/mainflow_test.go
ormergi: Done

ormergi:
`[Change](https://github.com/kiagnose/kiagnose/compare/23eb323fa7d76522efacf7031a97914ce90b4d95..594f97a66f264d8832391625452d47cc3460ba36)`: Fix error messages formatting

go.mod
```diff
@@ -2,7 +2,7 @@ module github.com/kiagnose/kiagnose
 
 go 1.17
 
-require github.com/stretchr/testify v1.7.0
+require github.com/stretchr/testify v1.7.1
```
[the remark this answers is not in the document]
ormergi: I update it since `ErrorContians` assert is introduced only since 1.7.1

go.sum
```diff
@@ -5,6 +5,8 @@ github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZN
 github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
 github.com/stretchr/testify v1.7.0 h1:nwc3DEeHmmLAfoZucVR881uASk0Mfjw8xYJ99tb5CcY=
 github.com/stretchr/testify v1.7.0/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
```
EdDev, line 7: Hmm.. if you run `tidy`, does this one go away?

kiagnose/mainflow.go
```diff
@@ -19,6 +19,45 @@
...
+	Run() error
+	Teardown() error
+	Report() error
+}
+
+func Run(launcher launcher) (runErr error) {
+	defer func() {
+		if reportErr := launcher.Report(); reportErr != nil {
+			if runErr != nil {
+				runErr = fmt.Errorf("%w, %v", runErr, reportErr)
```
EdDev, line 37: I missed the reason why you used `%w`.
I do not think we decided to expose the errors. If you do expose them, we will need to document them.

I suggest we avoid this for now, we can change it later per need. And you do have now means to check the errors in tests without it.

---

There is another problem I see now, the "mainflow failed: " prefix is repeating in the stacked errors.
I think we can follow this logic on each error handling we have:
- If there is no need to add an error to an existing error stack, then just do: `runErr = <new error>` (e.g. `runErr = reportErr`).
- If there is a need to add an error to an existing stack of errors, then do: `runErr = fmt.Errorf("%v, %v", runErr, <new error>)`.

To complete this, at the top add a defer that just adds the prefix if an error exists.

WDYT?

go.sum
ormergi: Yep, thanks for the heads up.

kiagnose/mainflow.go
ormergi: My intent was test the mainflow by asserting errors by their types and not content,
it turned out that `error.Is` wont work as I expected.
Now that the tests asserts on the error content anyway there is no value for wrapping them
no longer (`%w`). Sorry for the mess I will get rid of it.

Regarding the `mainflow failed` prefix, at first I handled it as part of defer as you suggest,
then it seemed like I am piggy backing that defer func with not related logic to its original
purpose (which is performing the report).

EdDev: > Regarding the mainflow failed prefix, at first I handled it as part of defer as you suggest,
then it seemed like I am piggy backing that defer func with not related logic to its original
purpose (which is performing the report).

I do not understand. There is now a problem that the prefix is added multiple times and the addition is repeated instead of being done once. Am I missing something?

EdDev (approved):
Looks great, thanks!

.gitignore
```diff
@@ -13,3 +13,9 @@
 
 # Dependency directories (remove the comment below to include it)
 # vendor/
+
+# GoLand
+.idea/
```
phoracek, line 18: These are not project-specific but you-specific. I recommend setting up https://stackoverflow.com/questions/7335420/global-git-ignore.

orelmisan (approved):
Thank you, looks great!

EdDev (dismissed):

EdDev (approved):

PR #9, opened 2022-03-28
Run checkup on its own namespace

ormergi:
A checkup is treated as an untrusted application, it should
be executed in an isolated way without interfering with other
workloads / components in the cluster or change anything.
When finished all related objects should be deleted, leaving
the cluster as it was before.

Checkup componentes includes a new namespace where the checkup will
run with and where all related objects will be created, such as:
- ServiceAccount
    That is granted with permissions from Roles and ClusterRoles
    that were specified at the checkup spec.
- ConfigMap
    Where the checkup results will be stored.
- Roles RoleBindings and ClusterRoleBindings
    In order to grant the permissions to the workspace SA.
- Batch Job
    That will be used to run the checkup container one time
    with the needed permissions.

This PR introduce the logic for creating and deleting the checkup components. 
And also: 
- client package for wrapping k8s client, it introduce an interface that enables
  replacing the client with a fake client in tests.
- Initial implementation of the checkup Spec which represent the use input, 
  it is expected to be populated with the data from the user ConfigMap.


ormergi:
Lint github action failed https://github.com/kiagnose/kiagnose/runs/5837175951?check_suite_focus=true#step:4:25
It pass on my local env.

EdDev (commented):
I liked the isolation of the client, nice!

There are some comments inline.
You also used some `utils` folder/package, which you probably know I cannot accept. :)

kiagnose/internal/checkup/checkup.go
```diff
@@ -19,21 +19,237 @@
...
+
+const (
+	NamespaceName                       = "checkup-workspace"
+	ResultsConfigMapNameEnvVarName      = "RESULT_CONFIGMAP_NAME"
+	ResultsConfigMapNameEnvVarNamespace = "RESULT_CONFIGMAP_NAMESPACE"
+	ResultsConfigMapWriterRoleName      = "results-configmap-writer"
+	ResultsConfigMapName                = "checkup-results"
+)
+
+func New(spec *Spec) *Checkup {
```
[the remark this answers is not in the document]
ormergi: Done

kiagnose/internal/client/client.go
```diff
@@ -0,0 +1,46 @@
...
+	"k8s.io/client-go/kubernetes"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	rbacv1client "k8s.io/client-go/kubernetes/typed/rbac/v1"
+	"k8s.io/client-go/rest"
+)
+
+type Client interface {
+	CoreV1() corev1client.CoreV1Interface
+	RbacV1() rbacv1client.RbacV1Interface
+}
```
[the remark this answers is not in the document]
ormergi: Done

kiagnose/internal/client/client.go
```diff
@@ -0,0 +1,46 @@
...
+
+func Create() (Client, error) {
```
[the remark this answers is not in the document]
ormergi: I changed it to return concrete object

kiagnose/internal/utils/configmap/configmap.go
```diff
@@ -0,0 +1,39 @@
...
+	"log"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+func Create(client corev1client.CoreV1Interface, cm *corev1.ConfigMap) (*corev1.ConfigMap, error) {
+	createdConfigMap, err := client.ConfigMaps(cm.Namespace).
+		Create(context.Background(), cm, metav1.CreateOptions{})
```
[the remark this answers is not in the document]
ormergi: Changed it to one liner

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -0,0 +1,222 @@
...
+	description string
+	checkup     *checkup.Checkup
+}
+
+func TestNewCheckup(t *testing.T) {
+	checkupCreateTestCases := []checkupCreateTestCase{
+		{
+			description: "should create checkup",
+			checkup:     newTestCheckup(),
+		},
```
EdDev, line 50: I would suggest to minimize it to:
```
{"should create checkup", checkup.New("framework:v1", nil, nil, nil)}
```

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -0,0 +1,222 @@
...
+		{
+			description: "should create checkup with additional Roles",
+			checkup:     newTestCheckupWithRoles(),
+		},
```
EdDev, line 54: and here to:
```
{"create checkup with Roles", checkup.New("framework:v1", nil, nil, roles())}
```

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -0,0 +1,222 @@
...
+
+	"github.com/kiagnose/kiagnose/kiagnose/internal/checkup"
+)
+
+type checkupCreateTestCase struct {
+	description string
+	checkup     *checkup.Checkup
+}
+
+func TestNewCheckup(t *testing.T) {
```
EdDev, line 45: I am a bit confused here.
It seems that you have exposed many of the inner structure of `checkup`. But I am not clear why.
All tests below are mechanic, they check that things got constructed.

But if we go back to the actual need from `checkup`, we wanted to just expose `Setup`, `Run` and `Teardown`.
So why all these intermediate details matter, or asking differently, why there is a need to expose everything?

kiagnose/internal/client/client.go
```diff
@@ -0,0 +1,58 @@
...
+	"k8s.io/client-go/kubernetes"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	rbacv1client "k8s.io/client-go/kubernetes/typed/rbac/v1"
+	"k8s.io/client-go/rest"
+)
+
+type Client struct {
+	corev1 corev1client.CoreV1Interface
+	rbacv1 rbacv1client.RbacV1Interface
+}
```
EdDev, line 32: After going this again, this client is doing basically two things:
- Minimizing the `kubernetes.Interface` to only the things needed at the moment.
- Abstracting away `rest.InClusterConfig()`.

I think the second point is good, especially if there is no other place we can do it so it will be shared with multiple packages.

I do not see the usefulness of the first point. Could you please show where it helped in the production or test code?

kiagnose/internal/checkup/checkup.go
```diff
@@ -19,21 +19,244 @@
...
+			Name: clusterRoleName + "to-sa"},
+		Subjects: []rbacv1.Subject{subject},
+		RoleRef: rbacv1.RoleRef{
+			Kind:     clusterRoleKind,
+			APIGroup: rbacv1.GroupName,
+			Name:     clusterRoleName},
+	}
+}
+
+func newCheckupJob(name, namespace, serviceAccountName, image string, envs []corev1.EnvVar) *batchv1.Job {
```
orelmisan, line 244: `activeDeadlineSeconds` field is missing from the Job's spec (the Job's timeout).

orelmisan (commented):
I've reviewed the second commit's production code, please see the inline comments.

kiagnose/internal/checkup/namespaces/namespaces.go
```diff
@@ -0,0 +1,73 @@
...
+		return nil, err
+	}
+	log.Printf("Namespace '%s' successfully created", namespace.Name)
+
+	return namespace, nil
+}
+
+// DeleteAndWait deletes a namespace and waits until delete
+func DeleteAllObjects(client corev1client.CoreV1Interface, nsName string) error {
+	_, err := client.RESTClient().Delete().Resource("*").Namespace(nsName).Do(context.Background()).Get()
```
EdDev, line 48: We are not suppose to use the `RESTClient`, right?

kiagnose/internal/checkup/namespaces/namespaces.go
```diff
@@ -0,0 +1,73 @@
...
+
+// waitForDeletion waits until the namespace will be removed from the cluster
+func waitForDeletion(client corev1client.CoreV1Interface, nsName string, timeout time.Duration) error {
+	log.Printf("waiting for namespace '%s' to dispose", nsName)
+	return wait.PollImmediate(time.Second, timeout, func() (bool, error) {
+		_, err := client.Namespaces().Get(context.Background(), nsName, metav1.GetOptions{})
+		if errors.IsNotFound(err) {
+			return true, nil
+		}
+		return false, nil
```
EdDev, line 71: I'm not sure I understand this.
If there is no error or if there is an error that is not about the missing namespace, you retry.

But if the error is `!errors.IsNotFound(err)` something is wrong (e.g. communication error). Even if we should keep on trying, I would expect a log of this to be posted.

kiagnose/internal/checkup/namespaces/namespaces.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	log.Printf("namespace '%s' successfully deleted", nsName)
+
+	return waitForDeletion(client, nsName, timeout)
+}
+
+// waitForDeletion waits until the namespace will be removed from the cluster
+func waitForDeletion(client corev1client.CoreV1Interface, nsName string, timeout time.Duration) error {
+	log.Printf("waiting for namespace '%s' to dispose", nsName)
+	return wait.PollImmediate(time.Second, timeout, func() (bool, error) {
```
EdDev, line 66: You want to retry every second for 5 min (default)? This is a bit extreme, don't you think?

We probably need to use a watcher for this, but this can be an improvement for later on. Consider opening an issue to track it (if you agree with this).

kiagnose/internal/checkup/namespaces/namespaces.go
```diff
@@ -0,0 +1,73 @@
...
+	namespace, err = client.Namespaces().Create(context.Background(), namespace, metav1.CreateOptions{})
+	if err != nil {
+		return nil, err
+	}
+	log.Printf("Namespace '%s' successfully created", namespace.Name)
+
+	return namespace, nil
+}
+
+// DeleteAndWait deletes a namespace and waits until delete
```
EdDev, line 46: This is not in sync with the function name.
It is also unclear why it is under `namespaces` package. Perhaps elaborate in the text.

EdDev (changes requested):

EdDev:
Also, please make most packages singular (and not plural, like `namespace`).

kiagnose/internal/checkup/namespaces/namespaces.go
ormergi: This is a leftover I will remove it, sorry about the noise.

kiagnose/internal/checkup/namespaces/namespaces.go
ormergi: I am open to other suggestions for the retry time, what you have in mind?

As for using  a watcher, I am not sure whats going to happen when the namespace is disposed
and the watcher still being used, I will give it a try and see how it goes.

kiagnose/internal/checkup/namespaces/namespaces.go
ormergi: Yep this is pretty much it, any other error then not-exist one is silent.
I will log the error messages.

kiagnose/internal/checkup/checkup.go
```diff
@@ -19,21 +19,244 @@
...
+		image:               image,
+		envVars:             inputEnvVars,
+		job:                 newCheckupJob(checkupJobName, NamespaceName, serviceAccountName, image, envVars),
+	}
+}
+
+func newNamespace(name string) *corev1.Namespace {
+	return &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
```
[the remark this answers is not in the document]
orelmisan: So we can have multiple checkups running simultaneously.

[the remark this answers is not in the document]
orelmisan: Each checkup could run in its own random Namespace (all Namespaces will share a common prefix and a unique suffix).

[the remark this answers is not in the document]
EdDev: Yes, sounds right to me.
Only the framework ns is constant, the checkup jobs are unique.
It is also the only way to assure this will work safely.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -0,0 +1,272 @@
...
+
+func newTestEnvVars() []corev1.EnvVar {
+	return []corev1.EnvVar{
+		{Name: "env-var-1", Value: "env-var-1-value"},
+		{Name: "env-var-1", Value: "env-var-1-value"}}
+}
+
+type testsClient struct {
+	fakeClient *fake.Clientset
+}
```
EdDev, line 213: Couldn't you just extended it? (I did not try it)
`type testClient struct { fake.Clientset }`

kiagnose/internal/checkup/namespace/namespace.go
```diff
@@ -0,0 +1,67 @@
...
+	return namespace, nil
+}
+
+// DeleteAndWait deletes a namespace and waits until delete
+func DeleteAndWait(client corev1client.CoreV1Interface, nsName string, timeout time.Duration) error {
+	err := client.Namespaces().Delete(context.Background(), nsName, metav1.DeleteOptions{})
+	if err != nil {
+		return err
+	}
+	log.Printf("namespace '%s' successfully deleted", nsName)
```
EdDev, line 52: Deleting has been successfully requested at this point, nothing more.
Actual deletion occurred after the wait period.

kiagnose/internal/checkup/namespace/namespace.go
```diff
@@ -0,0 +1,67 @@
...
+
+	return waitForDeletion(client, nsName, timeout)
+}
+
+// waitForDeletion waits until the namespace will be removed from the cluster
+func waitForDeletion(client corev1client.CoreV1Interface, nsName string, timeout time.Duration) error {
+	log.Printf("waiting for namespace '%s' to dispose", nsName)
+	return wait.PollImmediate(time.Second, timeout, func() (bool, error) {
```
EdDev, line 60: Polling each second is too fast IMO, consider increasing to 3 or 5.
If there is a version of polling with exponential backoff it may work best.

kiagnose/internal/checkup/rbac/rbac.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func DeleteClusterRoleBinding(client rbacv1client.RbacV1Interface, name string) error {
+	if err := client.ClusterRoleBindings().Delete(context.Background(), name, metav1.DeleteOptions{}); err != nil {
+		return err
+	}
+	log.Printf("ClusterRoleBinding '%s' successfully deleted\n", name)
+
+	return nil
+}
```
EdDev, line 76: I think this is not used.

kiagnose/internal/checkup/rbac/rbac.go
```diff
@@ -0,0 +1,139 @@
...
+		if err := client.ClusterRoleBindings().Delete(context.Background(), clsuterRoleBinding, metav1.DeleteOptions{}); err != nil {
+			danglingClusterRoleBindings = append(danglingClusterRoleBindings, clsuterRoleBinding)
+		}
+	}
+
+	if len(danglingClusterRoleBindings) > 0 {
+		return fmt.Errorf("failed to delete ClusterRoleBindings: %s", danglingClusterRoleBindings)
+	}
+
+	return nil
```
EdDev, line 92: Unlike the namespace, here you do not wait for it to be gone.
Is this intentional?

kiagnose/internal/checkup/checkup.go
```diff
@@ -52,6 +54,10 @@ type Checkup struct {
 	job                 *batchv1.Job
 }
 
+func (c *Checkup) SetTeardownTimeout(duration time.Duration) {
+	c.teardownTimeout = duration
+}
+
```
EdDev, line 60: You should place it somewhere after the factory or near the `Teardown` method.

kiagnose/internal/checkup/checkup.go
```diff
@@ -273,5 +279,29 @@ func (c *Checkup) Run() error {
...
+	if err := rbac.DeleteClusterRoleBindings(c.client.RbacV1(), c.clusterRoleBindings); err != nil {
+		errs = append(errs, err)
+	}
+
+	if err := namespace.DeleteAndWait(c.client.CoreV1(), c.namespace.Name, c.teardownTimeout); err != nil {
+		errs = append(errs, err)
+	}
+
+	if len(errs) > 0 {
+		return concentrateErrors(errs)
```
EdDev, line 293: It would be beneficial to mention it is a teardown failure.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -83,7 +83,6 @@ func TestCheckupSetup(t *testing.T) {
 			assertConfigMapWriterRoleBindingCreated(t, c)
 		})
 	}
-
```
EdDev: Unrelated change

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -189,23 +274,28 @@ func (c *testsClient) listNamespaces() ([]corev1.Namespace, error) {
...
+		clusterRoleBindingsList := objects.(*rbacv1.ClusterRoleBindingList)
+		return clusterRoleBindingsList.Items, err
+	}
+	return nil, nil
 }
 
 func (c *testsClient) listObjectsByKind(group, resourceName, resourceKind string) (runtime.Object, error) {
 	gvr := schema.GroupVersionResource{Group: group, Version: "v1", Resource: resourceName}
 	gvk := schema.GroupVersionKind{Group: group, Version: "v1", Kind: resourceKind}
-
```
EdDev: unrelated change

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -270,3 +360,15 @@ func assertNoObjectExists(t *testing.T, testClient *testsClient) {
 	assert.NoError(t, err)
 	assert.Equal(t, []rbacv1.ClusterRoleBinding{}, actualClusterRoleBindings)
 }
+
+func assertClusterRoleBindingNotExists(t *testing.T, testClient *testsClient) {
```
EdDev, line 364: I suggest you keep the previous naming pattern: ``assertNoClusterRoleBindingExists`.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -270,3 +360,15 @@ func assertNoObjectExists(t *testing.T, testClient *testsClient) {
...
+	ns, err := testClient.listClusterRoleBindings()
+	assert.NoError(t, err)
+	assert.Empty(t, ns)
+}
+
+func assertNamespaceNotExists(t *testing.T, testClient *testsClient) {
```
EdDev, line 370: Same, better keep the previous name pattern.

EdDev (changes requested):
Great work Or.

Although I have a lot of comments, this PR looks really nice and the unit tests are very thorough.

Please also try and go over the commit messages, they need some more work.

kiagnose/internal/checkup/checkup.go
```diff
@@ -19,21 +19,201 @@
...
+	var clusterRoleBindings []rbacv1.ClusterRoleBinding
+	for i := range clusterRoles {
+		clusterRoleBindings = append(clusterRoleBindings, newClusterRoleBinding(clusterRoles[i].Name, subject))
+	}
+
+	var envVars []corev1.EnvVar
+	envVars = append(envVars,
+		corev1.EnvVar{Name: ResultsConfigMapNameEnvVarName, Value: ResultsConfigMapName},
+		corev1.EnvVar{Name: ResultsConfigMapNameEnvVarNamespace, Value: NamespaceName})
+	envVars = append(envVars, inputEnvVars...)
```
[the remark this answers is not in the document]
ormergi: It is not possible to pass both single object and a slice (using ...) to append https://go.dev/play/p/Rk2sMg_Zhym
I will change envVars to be initialized with each envVar and append the slice with the rest.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/fd225ba37a6fb0dc8655775559c2b52912fee1b7..72a0be582ccab1954434a2050d3a45b368399f39) following "Introduce checkup components" review:
- Rename checkup.New() argument and local's names to be more consistent
- Initialize slices with elements instead of calling append twice.
- newCheckupJob() timeout argument is changed to int64 instead of time.Duration

kiagnose/internal/checkup/checkup_test.go
ormergi: I moved to previous commit.

kiagnose/internal/checkup/checkup_test.go
ormergi: Done

ormergi (commented):
[Changes](https://github.com/kiagnose/kiagnose/compare/bf689de988236a27af67e878f978912c57f31e55..b077fb93f916c260948f03919eaff9e57701ec6f) following "Delete checkup components on teardown" review

ormergi:
Commits messages were updated

kiagnose/internal/checkup/checkup.go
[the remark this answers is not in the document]
EdDev: This should have been added to the design to ease this discussion, we just missed it.

This is not only related to concurrency, but to proper execution. Failure for any reason to remove a namespace should not block other executions, can we agree on that?
In my view, this is no different from randomizing namespaces when running e2e tests (see k8s e2e framework).
There should be no dependencies between one run and another, be it serial or in parallel.

kiagnose/internal/checkup/checkup_test.go
ormergi: Moved to previous commit.

kiagnose/internal/checkup/checkup.go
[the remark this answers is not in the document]
orelmisan: This will probably cause a change to checkup.New(), because the namespace name will be known only after its creation.

kiagnose/internal/checkup/checkup.go
```diff
@@ -19,21 +19,198 @@
...
+	checkupPodSpec := corev1.PodTemplateSpec{
+		ObjectMeta: metav1.ObjectMeta{},
+		Spec: corev1.PodSpec{
+			ServiceAccountName:            serviceAccountName,
+			RestartPolicy:                 corev1.RestartPolicyNever,
+			TerminationGracePeriodSeconds: &terminationGracePeriod,
+			Containers:                    []corev1.Container{checkupContainer},
+		},
+	}
+	backoffLimit := int32(0)
```
orelmisan, line 212: Please consider treating `defaultTerminationGracePeriod` and `backoffLimit` in the same manner (as consts).

kiagnose/internal/checkup/checkup_test.go
EdDev: Is there a problem to expose it? We are in a test, what do you see problematic by doing so?

kiagnose/internal/checkup/checkup.go
```diff
@@ -203,7 +221,50 @@ func newCheckupJob(name, namespaceName, serviceAccountName, image string, active
...
+
+	if c.roleBindings, err = rbac.CreateRoleBindings(c.client.RbacV1(), c.roleBindings); err != nil {
+		return fmt.Errorf("%s: %v", errMessage, err)
+	}
+
+	defer func() {
+		if err != nil {
+			_ = rbac.DeleteClusterRoleBindings(c.client.RbacV1(), c.clusterRoleBindings)
+		}
+	}()
```
orelmisan, line 263: Why is this differed function is declared before `CreateClusterRoleBindings()`?

orelmisan (commented):
Hi @ormergi, I've reviewed your second and third commits, its a very nice work.

kiagnose/internal/checkup/checkup.go
ormergi: In the `defaultTerminationGracePeriod` case the linter considered it as magic number so I added a const for it,
for the `backoffLimit` it doesnt consider `0` as magic number (probably because its the default value).
If the liner would pass it I would define single variable with `:=`.

kiagnose/internal/checkup/checkup.go
```diff
@@ -19,21 +19,198 @@
...
+func newCheckupJob(name, namespaceName, serviceAccountName, image string, activeDeadlineSeconds int64, envs []corev1.EnvVar) *batchv1.Job {
+	const containerName = "checkup"
+
+	checkupContainer := corev1.Container{
+		Name:            containerName,
+		ImagePullPolicy: corev1.PullAlways,
+		Image:           image,
+		Env:             envs,
+	}
+	const defaultTerminationGracePeriod = 5
```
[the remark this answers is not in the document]
ormergi: In general I prefer to avoid shortening physical unit names, it is more consistent with other API's (e.g time.Duration func's use full names, k8s API's use full names for physical units), the code would look like this:
```
const defaultTerminationGracePeriodSeconds = 5
terminationGracePeriodSeconds := int64(defaultTerminationGracePeriodSeconds)
checkupPodSpec := corev1.PodTemplateSpec{
	ObjectMeta: metav1.ObjectMeta{},
	Spec: corev1.PodSpec{
	TerminationGracePeriodSeconds: &terminationGracePeriodSeconds,
        ...
```
I think its long and a bit too bulky,  note that the unit name can be understood from the code
as it is now, the usage is right next to the definition and used only one time, WDYT?

ormergi (commented):
Changes following "Introduce checkup components" review

kiagnose/internal/checkup/rbac/rbac.go
ormergi: > What can happen if one job finishes and immediately another is triggered?

If by  job finishes and immediately another is triggered you mean a checkup is executed and as soon as it finish
it is being executed again, it should work fine no collisions whatsoever since the clean up is performed.

In case a checkup is executed before another is finished and both use the same `ClusterRole`, the second
checkup will fail on creating the `ClusterRoleBinding` since its already executed.
Following the discussion on https://github.com/kiagnose/kiagnose/pull/9/commits/cc7301b0c031525f1ce077d9246d90c1dd721c86#r844968232
it means that ClusterRoleBinding names should also be generated.

> Lets assume the deletion of the cluster roles is in progress and not yet finished.
Will this cause a problem? Collision? Are these objects unique?

In case the ClusterRole is being deleted the checkup will probably failed at some point
since it wont have the permissions it expects to have.

kiagnose/internal/checkup/checkup.go
```diff
@@ -20,15 +20,29 @@
...
+
+	"github.com/kiagnose/kiagnose/kiagnose/internal/checkup/configmap"
+	"github.com/kiagnose/kiagnose/kiagnose/internal/checkup/namespace"
+	"github.com/kiagnose/kiagnose/kiagnose/internal/checkup/rbac"
 )
 
+type client interface {
+	CoreV1() corev1client.CoreV1Interface
+	RbacV1() rbacv1client.RbacV1Interface
+}
```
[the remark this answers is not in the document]
ormergi: In order to test `Checkup` an interface of the client is necessary.
I would also work with `kubernetes.Interface` (which default K8s clientset implements)
but you would need to figure out which of the interfaces `kubernetes.Interface` brings
a component use.
Defining our own interface for the client makes it visible and it is clear what are the dependencies.

kiagnose/internal/checkup/rbac/rbac.go
```diff
@@ -0,0 +1,137 @@
...
+		if err != nil {
+			createErr = err
+			break
+		}
+		createdBindings = append(createdBindings, *createdCrb)
+	}
+
+	if createErr != nil {
+		if err := deleteClusterRoleBindings(client, createdBindings); err != nil {
+			return nil, fmt.Errorf("failed to clean up ClusterRoleBindings: %v", err)
```
EdDev, line 58: It first failed to create, then alslo failed to clean up. I think you should add both to avoid confusion.

kiagnose/internal/checkup/checkup.go
```diff
@@ -19,21 +19,193 @@
...
+	ServiceAccountName             = "checkup-sa"
+	ResultsConfigMapName           = "checkup-results"
+	ResultsConfigMapWriterRoleName = "results-configmap-writer"
+	JobName                        = "checkup-job"
+
+	ResultsConfigMapNameEnvVarName      = "RESULT_CONFIGMAP_NAME"
+	ResultsConfigMapNameEnvVarNamespace = "RESULT_CONFIGMAP_NAMESPACE"
+)
+
+func New(image string, timeout time.Duration, envVars []corev1.EnvVar, clusterRoles []rbacv1.ClusterRole, roles []rbacv1.Role) *Checkup {
```
orelmisan: @ormergi can you please receive (and save) the following types?
```
clusterRoles []*rbacv1.ClusterRole
roles []*rbacv1.Role
```
When working with the K8s API, later on, we use them as pointers.

ormergi: Done

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/2e87b48e3bed4309a819c5d3199b14a95e771013..c66f8f67d2ffad85e8ef68ae225b06a2e46d8e7e) use pointers when referring to  reference types (e.g Role, ClusterRoles, etc..)

kiagnose/internal/checkup/checkup.go
orelmisan: Thank you.

kiagnose/internal/checkup/rbac/rbac.go
ormergi: Done

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -123,6 +123,78 @@ func TestCheckupSetupShould(t *testing.T) {
 	})
 }
 
+func TestCheckupTeardown(t *testing.T) {
```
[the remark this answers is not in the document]
ormergi: Done

kiagnose/internal/checkup/checkup.go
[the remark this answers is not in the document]
EdDev: I would just do this to reduce the number of names: `var terminationGracePeriod int64 = 5`.

Generally speaking, adding `Sec` to the tail will not harm anything.
At the same time, using any other name will be suffice as well (probably even `t`), just because it is just a mean to pass an address in a very localized area.

Being a nit, it does not deserve too much attention IMO.

EdDev (commented):
Second commit.

> Introduce checkup components

checkup: Construct a checkup instance

> A checkup is treated as an untrusted application, it should
> be executed in an isolated way without interfering with other
> workloads / components in the cluster or change anything.
> When finished all related objects should be deleted, leaving
> the cluster as it was before.

Nice introduction, however, it seems not related to the actual commit.
Probably more related to the teardown one.
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              

> Checkup componentes includes a new namespace where all related
> objects will be created, such as:

The checkup object includes the following components:
...

> This commit introduce the logic for generating each object that is
> required for the checkup according to the user spec.

This change implements the constructions of the checkup object and its members.

EdDev (commented):
Commit 3:

> Create checkup componentes on setup

checkup: Create components

> This commit introduce the logic for creating each of checkup
> componentes (e.g: Namespace, Job, etc..) in the cluster.

- This change ...
- typo: components

> Note about the 'launcher', since Setup() is the initialize the checkup
> it won't trigger Teardown().
> In case of failure checkup.Setup(), attempts to cleanup the object that
> already been created.
> The clean up is done by deleting the Namespace and eventually all the objects
> inside it [1].

The creation of the components is performed in the `Setup` method.
In case of failure, it will attempt to cleanup any intermediate objects.

(the part about teardown not being called when setup fails is not implemented in this change, so I would not mention it at all)

> In tests Kubernetes fake client is being used, and should be normalized
> to prevent inconsistent results when listing objects that been created
> and deleted, for example:
> With a "clean" client, where no objects were create, listing Namespaces
> will return 'nil'.
> But if a Namespace already been created and deleted, listing Namespaces
> result would be an empty Namespaces slice.

This is very confusing and hard to read.
How about something like:

```
The kubernetes fake client used in tests records objects that are created and deleted. In order to assert the results correctly, a normalization is required to some of the empty containers (slices, maps) the fake tracks. Any `nil` value is converted to an initialized empty container, allowing a semantic equality to pass.
```

kiagnose/internal/checkup/checkup.go
```diff
@@ -259,6 +261,34 @@ func (c *Checkup) Run() error {
...
 func (c *Checkup) Teardown() error {
+	var errs []error
+
+	if err := rbac.DeleteClusterRoleBindings(c.client.RbacV1(), c.clusterRoleBindings, c.teardownTimeout); err != nil {
+		errs = append(errs, err)
+	}
+
+	if err := namespace.DeleteAndWait(c.client.CoreV1(), c.namespace.Name, c.teardownTimeout); err != nil {
+		errs = append(errs, err)
+	}
```
EdDev, line 277: This shows that we should not mix between the delete and wait parts.
It is more efficient to delete all objects first and then wait for them afterwards.

This can be optimized in a following PR.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -277,3 +381,15 @@ func assertNoObjectExists(t *testing.T, testClient *testsClient) {
...
+	ns, err := testClient.listClusterRoleBindings()
+	assert.NoError(t, err)
+	assert.Empty(t, ns)
+}
+
+func assertNoNamespaceExists(t *testing.T, testClient *testsClient) {
+	ns, err := testClient.listNamespaces()
+	assert.NoError(t, err)
+	assert.Empty(t, ns)
+}
```
EdDev, line 395: There is some duplication (and difference in assertion) with `assertNoObjectExists`.
Can you use these new assertions in the previous one?

EdDev (commented):
Commit 4

> Delete checkup components on teardown

checkup: Delete components at teardown

kiagnose/internal/checkup/checkup.go
```diff
@@ -20,21 +20,35 @@
...
+	teardownTimeout     time.Duration
 	namespace           *corev1.Namespace
 	serviceAccount      *corev1.ServiceAccount
 	resultConfigMap     *corev1.ConfigMap
-	roles               []rbacv1.Role
-	roleBindings        []rbacv1.RoleBinding
-	clusterRoleBindings []rbacv1.ClusterRoleBinding
+	roles               []*rbacv1.Role
+	roleBindings        []*rbacv1.RoleBinding
+	clusterRoleBindings []*rbacv1.ClusterRoleBinding
```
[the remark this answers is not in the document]
ormergi: On the other hand, one could ask why these functions should be exported when reviewing the original commit :slightly_smiling_face:

kiagnose/internal/checkup/checkup.go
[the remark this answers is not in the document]
ormergi: I added seconds to the const, and changed the intermediate var to compact name,
hope it will make everyone happy :slightly_smiling_face:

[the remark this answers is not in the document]
ormergi: I added seconds to the const, and changed the intermediate var to compact name,
hope it will make everyone happy :slightly_smiling_face:

kiagnose/internal/checkup/checkup.go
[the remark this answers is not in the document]
orelmisan: @ormergi can you please open an issue about randomizing the Namespace's name?

kiagnose/internal/checkup/checkup.go
```diff
@@ -19,21 +19,192 @@
...
+			Name:      name,
+			Namespace: namespaceName,
+		},
+		Rules: []rbacv1.PolicyRule{
+			newConfigMapWriterPolicyRule(configMapName),
+		},
+	}
+}
+
+func newConfigMapWriterPolicyRule(cmName string) rbacv1.PolicyRule {
```
orelmisan, line 144: This function is called only once. Please consider placing its content inline.
IMO it will improve the Role definition's readability.

kiagnose/internal/checkup/checkup.go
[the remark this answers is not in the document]
ormergi: I have oppend an issue to track this https://github.com/kiagnose/kiagnose/issues/17

kiagnose/internal/checkup/checkup.go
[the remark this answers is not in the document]
ormergi: Moved to the previous commit.

kiagnose/internal/checkup/checkup.go
```diff
@@ -19,21 +19,192 @@
...
+}
+
+func newRoleBinding(roleName, namespaceName string, subject rbacv1.Subject) rbacv1.RoleBinding {
+	return rbacv1.RoleBinding{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "RoleBinding",
+			APIVersion: rbacv1.GroupName,
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      roleName,
```
orelmisan, line 160: Shouldn't this name represent the relationship between the Role and the ServiceAccount?

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -0,0 +1,279 @@
...
+	clusterRoleBindingKind     = "ClusterRoleBinding"
+	rolesResource              = "roles"
+	rolesBindingResource       = "rolebindings"
+	serviceAccountResource     = "serviceaccounts"
+	configMapResource          = "configmaps"
+
+	rbacResourceGroup = "rbac.authorization.k8s.io"
+
+	testImage   = "framework:v1"
+	testTimeout = time.Minute * 10
```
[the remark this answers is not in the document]
ormergi: Note that in general this is how you set 10 minutes time.Duration, 
for those tests 1 minute is good enough so I removed the `* 10` part.

kiagnose/internal/checkup/checkup.go
```diff
@@ -19,21 +19,192 @@
...
-func New() *Checkup {
-	return &Checkup{}
+func newClusterRoleBinding(clusterRoleName string, subject rbacv1.Subject) rbacv1.ClusterRoleBinding {
+	return rbacv1.ClusterRoleBinding{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "ClusterRoleBinding",
+			APIVersion: rbacv1.GroupName,
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: clusterRoleName},
```
orelmisan, line 177: Shouldn't this name represent the relationship between the ClusterRole and the ServiceAccount?

kiagnose/internal/checkup/checkup.go
```diff
@@ -19,21 +19,192 @@
...
+			Name:     clusterRoleName},
+	}
+}
+
+func newCheckupJob(name, namespaceName, serviceAccountName, image string, activeDeadlineSeconds int64, envs []corev1.EnvVar) *batchv1.Job {
+	const containerName = "checkup"
+
+	checkupContainer := corev1.Container{
+		Name:            containerName,
+		ImagePullPolicy: corev1.PullAlways,
```
orelmisan, line 191: Why was this image pull policy selected? (and not for example `IfNotPresent`)
https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy

kiagnose/internal/checkup/checkup.go
ormergi: I have opened an issue for tracking https://github.com/kiagnose/kiagnose/issues/18

kiagnose/internal/checkup/checkup_test.go
ormergi: Done

ormergi (commented):
Also updated the commit messages title's and body according to code-review comments.

ormergi:
@EdDev @orelmisan please take another look

orelmisan (commented):
Hi @ormergi, Please see the inline comments regarding the second commit.

kiagnose/internal/checkup/checkup.go
```diff
@@ -19,21 +19,192 @@
...
+const (
+	NamespaceName                  = "checkup-workspace"
+	ServiceAccountName             = "checkup-sa"
+	ResultsConfigMapName           = "checkup-results"
+	ResultsConfigMapWriterRoleName = "results-configmap-writer"
+	JobName                        = "checkup-job"
+
+	ResultsConfigMapNameEnvVarName      = "RESULT_CONFIGMAP_NAME"
+	ResultsConfigMapNameEnvVarNamespace = "RESULT_CONFIGMAP_NAMESPACE"
+)
```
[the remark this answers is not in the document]
ormergi: I placed the const part just before New() for visibility because its the place where it is being used.
Note that the checkup struct also use an interface that is defined just before it, meaning the const block
will be at the top of the file and might make it hard to follow.

I just realized that there are const that are being used only at New(), I moved them inside (job and env var names consts).

kiagnose/internal/checkup/checkup.go
```diff
@@ -19,21 +19,192 @@
...
+
+func New(image string, timeout time.Duration, envVars []corev1.EnvVar, clusterRoles []*rbacv1.ClusterRole, _ []*rbacv1.Role) *Checkup {
```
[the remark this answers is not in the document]
ormergi: We discussed it offline and realized that it will require forwarder changes to the code and tests and it is not 
a must for the MVP to work, if no one objects I prefer to do it on a follow up PR so we can get done
with this one.
I have opened an issue to tack this https://github.com/kiagnose/kiagnose/issues/20

kiagnose/internal/checkup/checkup.go
ormergi: I am planning to handle Roles on a follow up PR this will probably changed as well, I have opened an issue to tack this #20.

kiagnose/internal/checkup/checkup.go
EdDev: I do not understand this comment. Please be explicit on what you think is wrong and/or provide a suggestion.

ormergi: I am planning to handle Roles on a follow up PR this will probably changed as well, I have opened an issue to tack this https://github.com/kiagnose/kiagnose/issues/20.

kiagnose/internal/checkup/checkup.go
EdDev: I do not understand this comment. Please be explicit on what you think is wrong and/or provide a suggestion.

ormergi: Not sure what is the convention, since its going to be ephemeral object I left the name as is.
Please let me know if you have suggestion how it should be named.

kiagnose/internal/checkup/checkup.go
ormergi: Currently the framework user cannot control this, if the image pull policy it will be set to `IfNotPresent` by default,
there might be a case where new revisions of a checkup image exists but wont be fetched, ending up with 
the framework keep executing the old revision of a checkup image.

kiagnose/internal/checkup/checkup.go
orelmisan: We could use something like <RoleName>-<ServiceAccount_name> or vice versa.
I'll look for a convention.

kiagnose/internal/checkup/checkup.go
orelmisan: We could use something like `<ClusterRoleName>-<ServiceAccount_name>` or vice versa.
I'll look for a convention.

kiagnose/internal/checkup/checkup.go
```diff
@@ -19,21 +19,192 @@
...
+	checkupRoles := []rbacv1.Role{newConfigMapWriterRole(ResultsConfigMapWriterRoleName, NamespaceName, ResultsConfigMapName)}
+
+	subject := rbacv1.Subject{Kind: rbacv1.ServiceAccountKind, Name: ServiceAccountName, Namespace: NamespaceName}
+	var checkupRoleBindings []rbacv1.RoleBinding
+	for i := range checkupRoles {
+		checkupRoleBindings = append(checkupRoleBindings, newRoleBinding(checkupRoles[i].Name, NamespaceName, subject))
+	}
```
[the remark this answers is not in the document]
ormergi: Done

kiagnose/internal/checkup/checkup.go
```diff
@@ -19,21 +19,192 @@
...
+	var checkupClusterRoleBindings []rbacv1.ClusterRoleBinding
+	for i := range clusterRoles {
+		checkupClusterRoleBindings = append(checkupClusterRoleBindings, newClusterRoleBinding(clusterRoles[i].Name, subject))
+	}
```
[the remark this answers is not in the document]
ormergi: Done

ormergi:
I pushed new [changes](https://github.com/kiagnose/kiagnose/compare/09f3644aa6220d0a7a63a24d4bc5b8f8e168dfa2..19c1fbc807dade6c48d256548abe01b53738fa99) following 
https://github.com/kiagnose/kiagnose/pull/9#discussion_r851725702
https://github.com/kiagnose/kiagnose/pull/9#discussion_r851725752
And also https://github.com/kiagnose/kiagnose/pull/9#discussion_r851726591
to somthing nicer.

kiagnose/internal/checkup/checkup.go
EdDev: We are not suppose to set it to a tag like `latest` but to a digest. In such a case, it has no meaning IMO and saves you from problems like connectivity to the registry.
What is the setting in other large projects? E.g. kubevirt.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -52,7 +52,7 @@ const (
 	rbacResourceGroup = "rbac.authorization.k8s.io"
 
 	testImage   = "framework:v1"
-	testTimeout = time.Minute * 10
+	testTimeout = time.Minute
```
EdDev, line 55: nit: You changed it in the wrong commit.

kiagnose/internal/checkup/checkup.go
EdDev: You are talking about adding a posix to the binding name?
If it is to use some addition it is from the subject passes in, otherwise the caller should construct the name.

But this is only relevant if you expect more than one binding to be associated to the same ClusterRole. Is this the case?

kiagnose/internal/checkup/checkup.go
ormergi: Yes it is the best practice but it doesn't mean that every one follows it, I set this pull policy so that the framework will do best effort to get the right image.
On Kubevirt the image pull policy is controllable via `Kubevirts` custom resource 
https://github.com/kubevirt/kubevirt/blob/main/staging/src/kubevirt.io/api/core/v1/types.go#L1629

EdDev: And what is kubevirt default value set there?

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -0,0 +1,279 @@
...
+
+func (c *testsClient) listObjectsByKind(group, resourceName, resourceKind string) (runtime.Object, error) {
+	gvr := schema.GroupVersionResource{Group: group, Version: "v1", Resource: resourceName}
+	gvk := schema.GroupVersionKind{Group: group, Version: "v1", Kind: resourceKind}
+	return c.Tracker().List(gvr, gvk, "")
+}
+
+func assertNamespaceCreated(t *testing.T, testClient *fake.Clientset) {
+	gvr := schema.GroupVersionResource{Group: "", Version: "v1", Resource: namespaceResource}
+	actualNs, err := testClient.Tracker().Get(gvr, "", checkup.NamespaceName)
```
orelmisan, line 324: Can this be performed using the "Regular" `corev1().Namespaces().Get()`?
Why is it necessary to use `testClient.Tracker().Get()`

orelmisan (commented):
Hi @ormergi, I've reviewed most of the third commit.

kiagnose/internal/checkup/checkup.go
orelmisan: Consider a scenario:
You have a ClusterRole named `kubvirt-reader`.
You have checkup A and checkup B who needs to run simultaneously and use kubvirt-reader .
With the current implementation, the second checkup to run will fail to create the ClusterRoleBinding because it already exists.

orelmisan: Furthermore, a cluster admin who will run:
```
kubectl get clusterrolebindings
``` 

Will see a ClusterRoleBinding with the same name as the Role.

EdDev: Ok, so it is a name randomization issue and we can defer it with all the other ones.
Ref: #17

kiagnose/internal/checkup/checkup.go
ormergi: @EdDev  
> And what is kubevirt default value set there?
It is not clear from the docs what Kubevirt does by default, I think its `IfNotPresent`.

We can use the default policy meaning to no specify anything, it looks good enough
https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -0,0 +1,279 @@
...
+
+	assert.NoError(t, err)
+	assert.Equal(t, checkup.NewNamespace(checkup.NamespaceName), actualNs)
```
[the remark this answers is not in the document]
EdDev: It checks that the client has the expected data, and that data is created from the production code.
If you claim that we need to duplicate this creation of the data in the test, it means we couple it strongly to the implementation (i.e. if the content of the namespace creation in this case changes, we need to change the test).

Coupling the tests in this manner requires a very good reasoning.
Do you have one?

Perhaps this can help look at this from a different angle: We are not interested to check if the structure contains the correct data, we are interested to check if the logic of adding it, deleting it and so on works as expected.
One could claim that we also need to assert the exact correct data we want to reach the API, but that sounds like a useless test to me. We do not change the data, if we did, we should have checked it explicitly and then such a test would have made sense.

kiagnose/internal/checkup/namespace/namespace.go
```diff
@@ -0,0 +1,74 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+func Create(client corev1client.CoreV1Interface, namespace *corev1.Namespace) (*corev1.Namespace, error) {
+	var err error
+
+	namespace, err = client.Namespaces().Create(context.Background(), namespace, metav1.CreateOptions{})
```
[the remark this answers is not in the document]
EdDev: You will have to elaborate and provide a suggestion. I do not really understand what you mean by this.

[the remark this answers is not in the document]
orelmisan: I have provided a suggestion.
I suggested to remove the `var err error` and replace it with the short variable declaration stated above.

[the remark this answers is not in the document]
orelmisan: BTW, all other calls to K8s client in the code are using the short variable declaration:
```go
someObject, err := someAPICall()
```

kiagnose/internal/checkup/checkup_test.go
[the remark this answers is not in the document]
orelmisan: > We are not interested to check if the structure contains the correct data, we are interested to check if the logic of adding it, deleting it and so on works as expected.

Fine by me.
Thanks for the clarification.

kiagnose/internal/checkup/namespace/namespace.go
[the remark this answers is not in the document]
EdDev: Ohh so you mean to ask:
> Why do we need `var err error` here?

Well, I guess the confusion is on which line you commented and the indirect ref to this `err` thing.

I guess the intention was to use the same variable that came as an argument.
That is not really needed, and it can also renamed to something short like `ns`.
Whatever works for you guys, I am fine with all options, seems like a nit to me.

kiagnose/internal/checkup/namespace/namespace.go
```diff
@@ -0,0 +1,74 @@
...
+
+	return namespace, nil
+}
+
+// DeleteAndWait delete and waits for the given namespace to dispose.
+func DeleteAndWait(client corev1client.CoreV1Interface, nsName string, timeout time.Duration) error {
+	if err := client.Namespaces().Delete(context.Background(), nsName, metav1.DeleteOptions{}); err != nil {
+		return err
+	}
+	log.Printf("deleted namespace '%s' request sent", nsName)
```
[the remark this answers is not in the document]
ormergi: Done (all over)

kiagnose/internal/checkup/rbac/rbac.go
```diff
@@ -0,0 +1,186 @@
...
+
+func CreateServiceAccount(client corev1client.CoreV1Interface, sa *corev1.ServiceAccount) (*corev1.ServiceAccount, error) {
+	createdSa, err := client.ServiceAccounts(sa.Namespace).Create(context.Background(), sa, metav1.CreateOptions{})
+	if err != nil {
+		return nil, err
+	}
+	log.Printf("ServiceAccount '%s/%s' successfully created", sa.Namespace, sa.Name)
+
+	return createdSa, nil
+}
```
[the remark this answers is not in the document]
ormergi: What does it matter if its from another API group?
Note that `rbac` package provides helpers for working with k8s RBAC objects, 
ServiceAccount is part of it eventough its not from the same API group as other 
objects.

kiagnose/internal/checkup/namespace/namespace.go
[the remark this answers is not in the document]
orelmisan: > I guess the intention was to use the same variable that came as an argument.

I haven't noticed it in the first read.
This is a nit, sorry for not marking it correctly.

kiagnose/internal/checkup/checkup_test.go
ormergi: Yes it can work, the problem is that if you inject some mocking logic (e.g: return error when an object is created)
and the test triggers it as part of its assertion it will never pass and wont make sense.

For example: 
A test that checks, that checkup Setup() fails when it fails  to create a Namespace.
First we inject an error to the client when Namespace is created, then verify that no such object exists.
If we use `corev1().Namespaces().Get()` it will fail due to the injected error,
but if we use `testClient.Tracker().Get()` it will work as expected.

kiagnose/internal/checkup/namespace/namespace.go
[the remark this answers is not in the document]
ormergi: Done

kiagnose/internal/checkup/checkup_test.go
ormergi: Done

kiagnose/internal/checkup/checkup_test.go
orelmisan: Thank you for the explanation.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/534b2bc3b8922a311cca7ba917ff1f5268fb2c20..52f25853b751a9e622f1e1ce93ebc4d58b78f1e8): added unit test for creating correspond ClusterRoleBinding to additional ClusterRoles

ormergi:
Changes [[1]](https://github.com/kiagnose/kiagnose/compare/534b2bc3b8922a311cca7ba917ff1f5268fb2c20..52f25853b751a9e622f1e1ce93ebc4d58b78f1e8) [[2]](https://github.com/kiagnose/kiagnose/compare/534b2bc3b8922a311cca7ba917ff1f5268fb2c20..52f25853b751a9e622f1e1ce93ebc4d58b78f1e8): added unit test for creating correspond ClusterRoleBinding to additional ClusterRoles

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -283,11 +387,18 @@ func assertClusterRoleBindingsCreated(t *testing.T, testClient *fake.Clientset,
 // associated object, Thus other objects that were created inside the checkup Namespace
 // are not being checked explicitly.
 func assertNoObjectExists(t *testing.T, testClient *testsClient) {
-	actualNamespaces, err := testClient.listNamespaces()
+	assertNoNamespaceExists(t, testClient)
+	assertNoClusterRoleBindingExists(t, testClient)
+}
+
+func assertNoClusterRoleBindingExists(t *testing.T, testClient *testsClient) {
+	ns, err := testClient.listClusterRoleBindings()
```
EdDev, line 395: This `ns` is a copy/paste leftover probably.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -0,0 +1,293 @@
...
+	subject := rbacv1.Subject{Kind: rbacv1.ServiceAccountKind, Name: checkup.ServiceAccountName, Namespace: checkup.NamespaceName}
+	expectedRoleBinding := checkup.NewRoleBinding(checkup.ResultsConfigMapWriterRoleName, checkup.NamespaceName, subject)
+
+	assert.Equal(t, expectedRoleBinding, actualRoleBinding)
+}
+
+func assertClusterRoleBindingsCreated(t *testing.T, testClient *fake.Clientset, clusterRoles []*rbacv1.ClusterRole) {
+	gvr := schema.GroupVersionResource{Group: rbacv1.GroupName, Version: "v1", Resource: clusterRoleBindingResource}
+	gvk := schema.GroupVersionKind{Group: rbacv1.GroupName, Version: "v1", Kind: "ClusterRoleBinding"}
+	object, err := testClient.Tracker().List(gvr, gvk, "")
```
EdDev, line 371: I think you can use `testClient.listClusterRoleBindings()`

kiagnose/internal/checkup/checkup_test.go
ormergi: Done

kiagnose/internal/checkup/checkup_test.go
ormergi: Done

orelmisan (approved):
Thank you!

EdDev (approved):
Great, thanks!

PR #10, opened 2022-03-30
kiagnose: Restructure skeleton and wire concrete objects

EdDev:
In order to wire concrete implementation objects, the basic skeleton
of the project is updated.

- The exposed project API is now limited to a single `Run()` function.
- Individual packages have been created for:
  - launcher: Responsible for the upper level logic flow.
  - checkup: Responsible for the checkup workload logic.
  - reporter: Responsible for reporting the results of the run.

The mainflow logic is now moved into the `launcher` object.
The checkup and reporter are introduced with an empty implementation.

With this change, following changes can add checkup and reporter
implementations.


EdDev:
Please note that we still have some missing parts, like loading the configuration.
I guess it fits to be called by the `launcher` and have its own package & object.
Anyway, it can be added/wired in following PR/s per need.

kiagnose/internal/checkup/checkup.go
```diff
@@ -0,0 +1,43 @@
...
+	return nil
+}
+
+func (s *Checkup) Teardown() error {
+	return nil
+}
+
+func (s Checkup) Report() error {
+	return nil
+}
```
orelmisan, line 43: Is this method's planned responsibility is to get the results from the ConfigMap object that is in the checkup's namespace?

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+type checkupStub struct {
+	failSetup    error
+	failRun      error
+	failTeardown error
+}
+
+type reporterStub struct {
+	failReport error
+}
```
orelmisan, line 114: Can we put the type next to its methods?

orelmisan (commented):
@EdDev Thanks for the PR.
Looks good.
I have two questions in the inline comments.

kiagnose/internal/checkup/checkup.go
EdDev: This is wrong, it should not be here. `Report` is for the reporter.
Done

kiagnose/internal/launcher/launcher_test.go
EdDev: Done

EdDev:
[change](https://github.com/kiagnose/kiagnose/compare/b58638f47eeef8e64c37e5769d9532e2f1eff6b5..a56acd97dfb203cb444f11bd5b7cdb8674c45591): Answered review comments.

orelmisan:
@EdDev Thanks for the changes.
Do you want to address reading the checkup's results from the checkup's results ConfigMap in this PR?

orelmisan:
> Do you want to address reading the checkup's results from the checkup's results ConfigMap in this PR?

Is it the responsibility of the `reporter`?

EdDev:
> @EdDev Thanks for the changes. Do you want to address reading the checkup's results from the checkup's results ConfigMap in this PR?

This PR does not add functionality beyond what exists now in terms of API.
it just structures it differently so it can be wired to real implementation.

Reading results and reporting them is up to the checkup and reporter responsability.
One of the next steps is to implement the setup and teardown of the checkup.
Another to introduce the input config so the checkup will know what to do.
Same for reporting back part.

> Is it the responsibility of the reporter?

The reporter responsibility as we see today is to post results to the outer API (e.g. ConfigMap or future CR Status). Where this info resides and how it get to have it is not known at the moment. I guess it needs to be figure out when it will need to be implemented.

orelmisan (approved):

ormergi (approved):
Looks good, its easier to connect the dots now :sweat_smile:

PR #13, opened 2022-04-05
Use arnested/go-version-action

### LINKED ISSUE #5: Use go-version-action
The [go-version-action](https://github.com/marketplace/actions/go-version-action) provides means to use the Go version that is specified in the `go.mod` file.
It is considered the minimum Go version that should be used to build the module.

arnested:
Fixes #5.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -9,20 +9,28 @@ on:
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
-env:
-  GO_VERSION: 1.17
-
 jobs:
+  go-versions:
+    name: Lookup go versions
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.versions.outputs.matrix }}
+      minimal: ${{ steps.versions.outputs.minimal }}
+    steps:
+      - uses: actions/checkout@v2
+      - uses: arnested/go-version-action@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/check.yaml
```diff
@@ -22,20 +29,24 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v2
         with:
-          go-version: ${{ env.GO_VERSION }}
+          go-version: ${{ needs.go-versions.outputs.minimal }}
```
EdDev, line 32: @arnested , The check execution does not seem to take the right version.
https://github.com/kiagnose/kiagnose/runs/5873546432?check_suite_focus=true#step:3:7

EdDev (commented):
Thank you very much for the contribution!
I think there is something missing at the `lint` job as the go version used at the run looks wrong.

.github/workflows/check.yaml
arnested: Good catch, @EdDev!

I was wondering how that could happen, and then I realized I forgot the `needs: go-versions` on the  `lint` job.

I have pushed a fix.

arnested:
Good catch, @EdDev!

I was wondering how that could happen, and then I realized I forgot the needs: go-versions on the lint job.

I have pushed a fix.

EdDev (approved):
Thank you!

Issue #4, opened 2022-03-15
Containerize tests and build dependencies


ormergi:
Using the exact same tooling and flows both at CI and development environments 
is a good practice that prevent inconsistent builds, and improve overall development experience.

Example for using alpine-go container for building go project:
https://github.com/orelmisan/k8s-checkup-framework/blob/main/build/build-image
https://github.com/orelmisan/k8s-checkup-framework/blob/main/build/Dockerfile

EdDev:
Go provides consistent builds through its existing tooling (e.g `mod.sum`).

I think containerizing needs a stronger reason here, it is after-all not free, there is an additional cost to maintain such an image.`
Usually it comes when a 3rd party component is needed (e.g. a service when integration tests are running).

orelmisan:
The need here is for the go tool-chain to be at a consistent version (for example [1.17.8](https://go.dev/dl/)) in the developer's local environment and the CI.
There are official go images at [docker-hub](https://hub.docker.com/_/golang), the maintenance cost from our side is to bump the versions of the selected image according to a certain policy.
The source code and the output directories could be bind-mounted to the container created from this image.
AFAIU there is no need for us to create a dedicated image for this purpose.

ormergi:
The idea is to use the exact same tooling both for CI and development environment
in a way that it will be possible to move to another CI system in the future.
Using the same and toolings and flows both for all environments, will prevent inconsistent 
behavior on different env's ("It works on my laptop", "This test passed few moments a go").

Using local tolling for managing Go versions is nice, it do requires some time to set up
and jump between one version to another.
Note that fixing Go version is just another benefit of having such container, its not the main reason.
Naturally there will be more dependencies for this project other than Go runtime (e.g: golangci-lint)

As for the cost of creating such container image, I think only the time takes to create it should be considered.
Maintaining (e.g: add another dependency) is pretty straightforward.

orelmisan:
This could also be done using a multi-stage build, where the binary is built using an image with all the tooling, and then copied to a leaner image with only the necessary dependencies.

PR #15, opened 2022-04-05
Checkup config: read and parse checkup's config from ConfigMap object

orelmisan:
This add the ability to read and parse checkup's configuration from a ConfigMap object.


orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/66bff36aa0f3db018bcfc23d180186e755ab5bbb..edaaa90f089d16eb2f227ae7bdefa8ee0cba5bb1): Fixed a problem discovered by the linter.

kiagnose/mainflow.go
```diff
@@ -20,12 +20,43 @@
...
+	if err != nil {
+		return err
+	}
+	log.Printf("ConfigMap fullname: \"%s/%s\"", configMapNamespace, configMapName)
+
 	l := launcher.New(checkup.New(), reporter.New())
 	return l.Run()
 }
+
+func readConfigMapFullNameFromEnv() (namespace, name string, err error) {
```
EdDev, line 53: This should not be here, but hidden in some package.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/81957fe3100b8474c3dd428df0bb0f9be635f4e7..7f40b7036a5466a7bec415ee999f346c897e091b): Changed a commit message.

orelmisan:
Changes: Optimized the unit test code.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/65691c0adf658ea9aa0c19573d97431a00913552..5ba741e6a9ce2583af23c85a1141141f72f73022): Fixed linter problems.

kiagnose/internal/checkup/spec_test.go
```diff
@@ -0,0 +1,218 @@
...
+)
+
+func TestNewSpecLoaderShould(t *testing.T) {
+	const missingEnvVarErr = "missing required environment variable(s)"
+
+	t.Run("run successfully when required environment variables exists", func(t *testing.T) {
+		os.Setenv(checkup.ConfigMapNamespaceEnvVarName, configMapNamespace)
+		defer os.Unsetenv(checkup.ConfigMapNamespaceEnvVarName)
+		os.Setenv(checkup.ConfigMapNameEnvVarName, configMapName)
+		defer os.Unsetenv(checkup.ConfigMapNameEnvVarName)
```
EdDev, line 49: I prefer our unit tests to be "pure", in the sense they are not to depend on the OS.

Per what I actually see in these tests, it is also not needed. The production code does not call `os.*`, so I am unclear why it is used at all.

kiagnose/internal/checkup/spec_test.go
```diff
@@ -0,0 +1,218 @@
...
+func checkNewSpecLoaderSuccess(t *testing.T) {
+	env := os.Environ()
+	fakeClient := fake.NewSimpleClientset()
+
+	specLoader, err := checkup.NewSpecLoader(fakeClient, env)
+	assert.NoError(t, err)
+	assert.NotNil(t, specLoader)
+}
+
+func checkNewSpecLoaderFailure(t *testing.T, expectedError string) {
```
EdDev, line 189: I find this helper problematic for several reasons:
- It hides the main test operation, mixing both execution of the tested production code and the assertions.
- That you expect a specific error text, which does not seem to be a public API (i.e. it can change and therefore break all these tests).
- That you use `os.Environ()` and therefore depend on the OS, when there is nothing that requires it.
- That you create a fake client, but you do not really have to.

kiagnose/internal/checkup/spec_test.go
```diff
@@ -0,0 +1,218 @@
...
+	env := os.Environ()
+	fakeClient := fake.NewSimpleClientset()
+
+	specLoader, err := checkup.NewSpecLoader(fakeClient, env)
+	assert.ErrorContains(t, err, expectedError)
+	assert.Nil(t, specLoader)
```
EdDev, line 195: It is a common practice to ignore the returned value in case of error.
No one should trust the value in case of an error, so we do not need to check it.

kiagnose/mainflow.go
```diff
@@ -20,12 +20,50 @@
 package kiagnose
 
 import (
+	"log"
+	"os"
+
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
```
EdDev, line 27: I prefer not to see such a dependency at this top level.
Barring it deep under an inner package makes it cleaner IMO. (e.g. [this](https://github.com/kiagnose/kiagnose/pull/9/files#diff-c810fb751f76dc7919d84b25d9611683296c4e3648b45d588ba7db3af1954038R20))

kiagnose/mainflow.go
```diff
@@ -20,12 +20,50 @@
...
 	"github.com/kiagnose/kiagnose/kiagnose/internal/reporter"
 )
 
 func Run() error {
+	clientset, err := getK8sClientset()
+	if err != nil {
+		return err
+	}
+
+	env := os.Environ()
```
EdDev, line 40: This is an implementation detail I prefer to be deep inside the package that implements it.

kiagnose/mainflow.go
```diff
@@ -20,12 +20,50 @@
...
+	specLoader, err := checkup.NewSpecLoader(clientset, env)
+	if err != nil {
+		return err
+	}
+
+	checkupSpec, err := specLoader.Load()
+	if err != nil {
+		return err
+	}
```
EdDev, line 49: I would have preferred to see a separation between the `checkup` package and the loading/importing of the configuration.
To have such a separation, I imagine a totally independent package that is instantiated and then instructed to load the configuration, similar to what you just did.

I am unclear why I need a `*Loader`, I think from a usage perspective, this can be enough:
```
conf, err := config.New(clientset)
if err != nil { return err }

confData, err := conf.Load()
if err != nil { return err }
```

In the `config` package I expect to see a `config` file that has the general/common parts and in a `configmap` the things specific to configmaps. Same for `envvar` and other implementation specifics.
Considering that we do want to more in the future away from them, we want to change as little as possible at the business logic or common areas.

EdDev (changes requested):

kiagnose/internal/checkup/spec_test.go
orelmisan: This was removed.

kiagnose/mainflow.go
orelmisan: Done.

kiagnose/mainflow.go
orelmisan: This was removed.

orelmisan:
Changes: Fixed code review comments + linter errors.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/baae1eaeddbcede3194a2bbc676c2c7509cec323..fd48f7f2987eb4a710e672574b78f36e29621257): Moved types form loader_test.go to outside of Test*() functions.
Also changed their field names according to an offline code-review comment by @ormergi.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/fd48f7f2987eb4a710e672574b78f36e29621257..cf1d4c7071850ada0d22df9da2b04a5fa79df61b): Changed Config.Params to Config.EnvVars of type `[]corev1.EnvVar`.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/e0d7a0f6299fb4fd640e46192b4f7b0bc3210846..bcb79833db1d963b3cc7de46893568817dba6841): Added `config.ErrRoleNameIsIllegal`.

kiagnose/internal/config/loader.go
```diff
@@ -0,0 +1,211 @@
...
+
+	return nil
+}
+
+func parseListSeparatedByNewlines(rawString string) []string {
+	trimmedString := strings.TrimSpace(rawString)
+	return strings.Split(trimmedString, "\n")
+}
+
+func splitRoleName(fullName string) (namespace, name string, err error) {
```
EdDev, line 199: Why are you using named return values? This is not a common practice and is usually reserved for cases where you must (e.g. using a defer that wants to change the returned value).

kiagnose/internal/config/loader.go
```diff
@@ -0,0 +1,211 @@
...
+	ErrImageFieldIsMissing   = errors.New("image field is missing")
+	ErrTimeoutFieldIsMissing = errors.New("timeout field is missing")
+	ErrTimeoutFieldIsIllegal = errors.New("timeout field is illegal")
+	ErrRoleNameIsIllegal     = errors.New("illegal Role name")
+)
+
+type Loader struct {
+	client             kubernetes.Interface
+	configMapNamespace string
+	configMapName      string
```
EdDev, line 59: Looking at this after looking at all the rest of the code, it looks to me like these 3 values are the only ones needed for a loader. All the rest are results of intermediate steps which do not really need to be preserved in this object state.

kiagnose/internal/config/loader_test.go
```diff
@@ -0,0 +1,284 @@
...
+		{
+			TypeMeta:   metav1.TypeMeta{Kind: "Role", APIVersion: rbacv1.GroupName},
+			ObjectMeta: metav1.ObjectMeta{Name: "role1", Namespace: "default"},
+		},
+		{
+			TypeMeta:   metav1.TypeMeta{Kind: "Role", APIVersion: rbacv1.GroupName},
+			ObjectMeta: metav1.ObjectMeta{Name: "role2", Namespace: "default"},
+		},
+	}
+)
```
EdDev, line 82: It seems to me like all globals here are used in specific tests (or groups) and not shared between many.
So it would be safer and more easily read if placed near the usage as possible.

Please try to follow this with other variables, I have may missed others.

kiagnose/internal/config/loader_test.go
```diff
@@ -0,0 +1,284 @@
...
+	configMapData map[string]string
+	expectedError string
+}
+
+func TestNewConfigLoaderShould(t *testing.T) {
+	t.Run("run successfully when required environment variables exists", func(t *testing.T) {
+		fakeClient := fake.NewSimpleClientset()
+		loader, err := config.NewLoader(fakeClient, goodEnv)
+		assert.NoError(t, err)
+		assert.NotNil(t, loader)
```
EdDev, line 109: I think that checking that it is not a nil is not very useful, could you check against something you expect to see in the output that has some explicit meaning?
Such asserts have value during development, but less value at the end (unless you can check for some explicit result).

kiagnose/internal/config/loader_test.go
```diff
@@ -0,0 +1,284 @@
...
+			expectedErrorContains: expectedErrorPrefix,
+		},
+		{
+			description:           "fail when both ConfigMap's environment variables are missing",
+			envVars:               map[string]string{},
+			expectedErrorContains: expectedErrorPrefix,
+		},
+	}
+
+	fakeClient := fake.NewSimpleClientset()
```
EdDev, line 132: You can inline this, but is it even needed for these tests?

kiagnose/internal/config/loader_test.go
```diff
@@ -0,0 +1,284 @@
...
+
+type loadFailureTestCase struct {
+	description   string
+	configMapData map[string]string
+	expectedError string
+}
+
+func TestNewConfigLoaderShould(t *testing.T) {
+	t.Run("run successfully when required environment variables exists", func(t *testing.T) {
+		fakeClient := fake.NewSimpleClientset()
```
EdDev, line 106: Can you pass a `nil` instead of a fake? You do not really need the fake in these tests.

kiagnose/internal/config/loader_test.go
```diff
@@ -0,0 +1,284 @@
...
+			loader, err := config.NewLoader(fakeClient, goodEnv)
+			assert.NoError(t, err)
+			assert.NotNil(t, loader)
+			_, err = loader.Load()
+			assert.ErrorContains(t, err, testCase.expectedError)
+		})
+	}
+}
+
+func newFakeClientset(namespace,
```
EdDev, line 238: This seems to be a clientset with stuff in it. Can you please name it to express that?

kiagnose/internal/config/loader_test.go
```diff
@@ -0,0 +1,284 @@
...
+			Namespace: namespace,
+		},
+		Data: data,
+	}
+}
+
+func newExpectedConfig(image, timeout string,
+	envVars []corev1.EnvVar,
+	clusterRules []*rbacv1.ClusterRole,
+	roles []*rbacv1.Role) *config.Config {
```
EdDev, line 275: I would prefer one liner, even if it is long. It is hard to read it like this.

Same for the others.

EdDev (changes requested):

kiagnose/internal/config/loader.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	if err := l.loadRoles(); err != nil {
+		return nil, err
+	}
+
+	return l.checkupConfig, nil
+}
+
+func (l *Loader) getConfigMapFullNameFromEnv(env map[string]string) (err error) {
```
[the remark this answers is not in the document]
orelmisan: 1. The named returned value was a leftover, thanks for catching it.
2. Agreed.

kiagnose/internal/config/loader.go
orelmisan: Our linter requires it:
```
./automation/make.sh 
kiagnose/internal/config/loader.go:179:1: unnamedResult: consider giving a name to these results (gocritic)
func splitRoleName(fullName string) (string, string, error) {
^
```

kiagnose/mainflow.go
orelmisan: Done.

kiagnose/internal/config/loader.go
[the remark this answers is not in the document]
orelmisan: Done.

kiagnose/internal/config/loader.go
```diff
@@ -0,0 +1,211 @@
...
+	configMapName      string
+	checkupConfig      *Config
+	clusterRoleNames   []string
+	roleNames          []string
+}
+
+func NewLoader(client kubernetes.Interface, env map[string]string) (*Loader, error) {
+	l := &Loader{
+		client:        client,
+		checkupConfig: &Config{},
```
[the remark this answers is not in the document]
orelmisan: This was changed, and taken out of config.Loader.

kiagnose/internal/config/loader.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+	}
+
+	if data[ClusterRolesKey] != "" {
+		l.clusterRoleNames = parseListSeparatedByNewlines(data[ClusterRolesKey])
+	}
+
+	if data[RolesKey] != "" {
+		l.roleNames = parseListSeparatedByNewlines(data[RolesKey])
+	}
```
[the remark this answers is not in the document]
orelmisan: Done.

kiagnose/internal/config/loader.go
orelmisan: Done.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/bcb79833db1d963b3cc7de46893568817dba6841..ea90db57115a224a2328eae590b1f5684e32ef9a): Fixed comments from code review in production code.

kiagnose/internal/config/loader_test.go
orelmisan: Replaced it with nil.

kiagnose/internal/config/loader_test.go
orelmisan: Replaced it with nil.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/ea90db57115a224a2328eae590b1f5684e32ef9a..cbe4d70fbee84e205150a79cc1ee17385fbf6fc8): Fixed comments from code review in unit test code.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/cbe4d70fbee84e205150a79cc1ee17385fbf6fc8..3df113b0ca02f8742dc59b9b13d37d2932a2f8e8): Added a test case - fail when ConfigMap object doesn't exists.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/3bafbf849f1e0ab20ce67258d9c9e6ad338650d1..f9bdef8ead13148dffaecf433eb81abe87bd5590): Moved low level K8s client usage to a dedicated function.
Try to get ClusterRole and Role objects, only after successful parsing of their names.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/f9bdef8ead13148dffaecf433eb81abe87bd5590..68792f2fa76a70b2ddcd45295ec313b30c4d49f9): Extracted the parsing logic from config.Loader to a new entity.

orelmisan:
Change: Fixed code review comments regarding unit test code.

kiagnose/internal/config/loader.go
```diff
@@ -0,0 +1,141 @@
...
+			return nil, err
+		}
+
+		clusterRoles = append(clusterRoles, clusterRole)
+	}
+
+	return clusterRoles, nil
+}
+
+func (l *Loader) getRoles(roleNames []string) ([]*rbacv1.Role, error) {
```
EdDev, line 96: nit: You do not need the get really, it is redundant.

kiagnose/internal/config/loader.go
```diff
@@ -0,0 +1,141 @@
...
+	var roles []*rbacv1.Role
+
+	for _, roleName := range roleNames {
+		splitName, err := splitRoleName(roleName)
```
EdDev, line 100: nit: `fullRoleName` ?

kiagnose/internal/config/parser.go
```diff
@@ -0,0 +1,187 @@
...
+	configMapName      string
+	configMapRawData   map[string]string
+	image              string
+	timeout            time.Duration
+	params             map[string]string
+	clusterRoleNames   []string
+	roleNames          []string
+}
+
+func newConfigMapParser(client kubernetes.Interface, env map[string]string) (*configMapParser, error) {
```
EdDev, line 66: Can you also name the file with relation to the ConfigMap? e.g. `cmparser`

kiagnose/internal/config/loader.go
```diff
@@ -0,0 +1,141 @@
...
+		parser: parser,
+	}
+
+	return l, nil
+}
+
+func (l *Loader) Load() (*Config, error) {
+	if err := l.parser.Get(); err != nil {
+		return nil, err
+	}
```
EdDev, line 55: It does not seem to be a responsibility of a parser to get the object. A parser parses, it is not suppose to know how to fetch the object.

But I think this can be improved in parallel to the next step, so for now I tend to just let it roll.

kiagnose/internal/config/parser.go
```diff
@@ -0,0 +1,187 @@
...
+		configMapNamespace: namespace,
+		configMapName:      name,
+		params:             map[string]string{},
+	}, nil
+}
+
+func (cmp *configMapParser) Get() error {
+	configMap, err := cmp.client.CoreV1().ConfigMaps(cmp.configMapNamespace).Get(context.Background(), cmp.configMapName, metav1.GetOptions{})
+	if err != nil {
+		return err
```
EdDev, line 90: I think you are not testing this path.

kiagnose/internal/config/loader.go
```diff
@@ -0,0 +1,141 @@
...
+
+	for _, roleName := range roleNames {
+		splitName, err := splitRoleName(roleName)
+		if err != nil {
+			return nil, err
+		}
+
+		role, err := l.client.RbacV1().Roles(splitName.Namespace).Get(context.Background(), splitName.Name, metav1.GetOptions{})
+		if err != nil {
+			return nil, err
```
EdDev, line 107: I think you are not testing this path.

kiagnose/internal/config/loader.go
```diff
@@ -0,0 +1,141 @@
...
+	return checkupConfig, nil
+}
+
+func (l *Loader) getClusterRoles(clusterRoleNames []string) ([]*rbacv1.ClusterRole, error) {
+	var clusterRoles []*rbacv1.ClusterRole
+
+	for _, name := range clusterRoleNames {
+		clusterRole, err := l.client.RbacV1().ClusterRoles().Get(context.Background(), name, metav1.GetOptions{})
+		if err != nil {
+			return nil, err
```
EdDev, line 87: I think you are not testing this path.

EdDev (commented):
I had some comments but I am ok with proceeding in this state as is and fixing/improving as a followup PR.

Review and check if you want to take some of the suggestions now.
And most importantly, please check this is working ok against a real `ConfigMap` if you can.

kiagnose/internal/config/loader.go
```diff
@@ -0,0 +1,220 @@
...
+
+	checkupConfig.Roles, err = l.getRoles(roleNames)
+	if err != nil {
+		return nil, err
+	}
+
+	return checkupConfig, nil
+}
+
+func (l *Loader) getClusterRoles(clusterRoleNames []string) ([]*rbacv1.ClusterRole, error) {
```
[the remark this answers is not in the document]
orelmisan: `l.clusterRoles` looks off to me.

kiagnose/internal/config/loader.go
orelmisan: I agree that its not the parser's responsibility.

kiagnose/internal/config/parser.go
orelmisan: It is tested in a scenario when the ConfigMap doesn't exists.
When we'll have E2E tests, they should also test read permission problems.

kiagnose/internal/config/loader.go
orelmisan: It is tested in a scenario when the Role doesn't exists.
When we'll have E2E tests, they should also test read permission problems.

kiagnose/internal/config/loader.go
orelmisan: It is tested in a scenario when the ClusterRole doesn't exists.
When we'll have E2E tests, they should also test read permission problems.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/0931821973a9544544444c33b6c711e5f2bb97ec..fc259ce518132787efff1cf669409783843339f3): Fixed code review comments.

kiagnose/internal/config/parser.go
EdDev: The unit tests do not simulate the error occurring here, which means this condition is not evaluated and covered.

kiagnose/internal/config/loader.go
EdDev: This still needs to be covered by unit tests.

kiagnose/internal/config/loader.go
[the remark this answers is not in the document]
EdDev: I do not like `get`, but this is a nit, I will not insist.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/fc259ce518132787efff1cf669409783843339f3..6bf2db6e17030bfe16049fe1144e30c97d12315d): Moved ConfigMap namespace and name back to loader.go.

kiagnose/internal/client/client.go
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package client
+
+import (
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+func GetK8sClientset() (*kubernetes.Clientset, error) {
```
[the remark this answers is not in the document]
orelmisan: Done.

EdDev (approved):
Thank you!

kiagnose/internal/config/loader.go
orelmisan: The call to `Get()` returns an error if there is no ClusterRole with the requested name.

kiagnose/internal/config/loader.go
orelmisan: The call to `Get()` returns an error if there is no Role with the requested `namspace/name`.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/23ab02e01cef696d6b1c3d0c603e17751abe48c0..6db629219d4484072c5541b1c9726d49edcc5e6b): Rebase and integration with PR #9.

kiagnose/internal/config/loader.go
```diff
@@ -0,0 +1,165 @@
...
+
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	ConfigMapNamespaceEnvVarName = "CONFIGMAP_NAMESPACE"
+	ConfigMapNameEnvVarName      = "CONFIGMAP_NAME"
+)
+
+type Loader struct {
```
ormergi, line 39: It would make Loader{} easier to change (e.g moving to CRD) if it would hold an instance of a parser,
it could also make it easier to test as you could replace it with a stub on tests.

kiagnose/internal/config/loader_test.go
```diff
@@ -0,0 +1,305 @@
...
+			sort.Slice(actualConfig.EnvVars, func(i, j int) bool {
+				return actualConfig.EnvVars[i].Name < actualConfig.EnvVars[j].Name
+			})
+
+			assert.Equal(t, testCase.expectedConfig, actualConfig)
+		})
+	}
+}
+
+func TestLoaderLoadShouldFail(t *testing.T) {
```
ormergi, line 165: Please add tests to cover client errors e.g: when it fails to fetch Role's ClusterRole's or ConfigMap's
similar to what we do at `checkup` tests.

kiagnose/internal/config/loader_test.go
```diff
@@ -0,0 +1,305 @@
...
+		t.Run(testCase.description, func(t *testing.T) {
+			fakeClient := newFakeClientWithObjects(configMapNamespace, configMapName, testCase.configMapData, nil, nil)
+
+			loader, err := config.NewLoader(fakeClient, goodEnv)
+			assert.NoError(t, err)
+
+			_, err = loader.Load()
+			assert.ErrorContains(t, err, testCase.expectedError)
+		})
+	}
```
ormergi, line 222: I think those tests should have their own test suite at `cmparser` package, 
it would make it easier to test Loader{} as you could replace it the parser with a stub.
Will you give it a try?

kiagnose/mainflow.go
```diff
@@ -20,17 +20,32 @@
...
+
 	"github.com/kiagnose/kiagnose/kiagnose/internal/checkup"
 	"github.com/kiagnose/kiagnose/kiagnose/internal/client"
+	"github.com/kiagnose/kiagnose/kiagnose/internal/config"
 	"github.com/kiagnose/kiagnose/kiagnose/internal/launcher"
 	"github.com/kiagnose/kiagnose/kiagnose/internal/reporter"
 )
 
-func Run() error {
+func Run(env map[string]string) error {
```
ormergi, line 30: I think mainflow.Run() should read env by it self instead of receiving it as parameter,
it should at least validate ConfigMap name and namespace env vars are valid as it is critical 
requirement for it to work.
It is not the scope of the "config loader" to validate it.

kiagnose/internal/config/loader.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+func (l *Loader) getConfigMapData() (map[string]string, error) {
+	configMap, err := l.client.CoreV1().ConfigMaps(l.configMapNamespace).Get(context.Background(), l.configMapName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	return configMap.Data, nil
+}
```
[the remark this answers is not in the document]
orelmisan: There is now a configmap package containing both functions.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/6db629219d4484072c5541b1c9726d49edcc5e6b..088d2010d0fcb2ddb72db853ea5c54b7e2e03b4f): Added a configmap package to unify actions against a ConfigMap object.

kiagnose/internal/config/loader.go
```diff
@@ -0,0 +1,165 @@
...
+
+	for _, roleName := range roleNames {
+		fullRoleName, err := splitRoleName(roleName)
+		if err != nil {
+			return nil, err
+		}
+
+		role, err := l.client.RbacV1().Roles(fullRoleName.Namespace).Get(context.Background(), fullRoleName.Name, metav1.GetOptions{})
+		if err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
orelmisan: Done.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/088d2010d0fcb2ddb72db853ea5c54b7e2e03b4f..d24bc743c40da32d9b9058d1ef1e5188b6b18aed): Put the logic of `splitRoleName()` inline.

kiagnose/internal/config/loader_test.go
```diff
@@ -0,0 +1,305 @@
...
+			assert.NoError(t, err)
+
+			actualConfig, err := loader.Load()
+			assert.NoError(t, err)
+
+			sort.Slice(actualConfig.EnvVars, func(i, j int) bool {
+				return actualConfig.EnvVars[i].Name < actualConfig.EnvVars[j].Name
+			})
+
+			assert.Equal(t, testCase.expectedConfig, actualConfig)
```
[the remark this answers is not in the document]
orelmisan: IIUC assert.ElementsMatch() works on slices.
Here we have a slice inside a struct so it won't help in this case because we are comparing structs.

kiagnose/internal/config/loader_test.go
```diff
@@ -0,0 +1,305 @@
...
+		})
+	}
+}
+
+func TestLoaderLoadShouldFail(t *testing.T) {
+	t.Run("when ConfigMap doesn't exist", testLoaderFailureWhenConfigMapIsMissing)
```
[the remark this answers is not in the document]
orelmisan: Done.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/d24bc743c40da32d9b9058d1ef1e5188b6b18aed..230c3553275929c26f313afcec5ba8b57c7f281f): Inlined testLoaderFailureWhenConfigMapIsMissing()

kiagnose/internal/config/loader.go
orelmisan: The tests in loader_test.go are also covering 100% of the parser code.

kiagnose/internal/config/loader.go
```diff
@@ -0,0 +1,165 @@
...
+	ConfigMapNameEnvVarName      = "CONFIGMAP_NAME"
+)
+
+type Loader struct {
+	client             kubernetes.Interface
+	configMapNamespace string
+	configMapName      string
+}
+
+func NewLoader(client kubernetes.Interface, env map[string]string) (*Loader, error) {
```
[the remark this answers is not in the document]
orelmisan: As we have discussed offline, the environment variables validation logic was moved to Loader.Load().
NewLoader() doesn't return an error any more.

kiagnose/internal/config/loader_test.go
orelmisan: This is covered by three tests:
1. [Failure to get a ConfigMap](https://github.com/kiagnose/kiagnose/pull/15/files#diff-9c52ee6d71999c83d9e358fc8c604002291553b5351a4c710ac89fdc1b662206R157).
2. [Failure to get a ClusterRole](https://github.com/kiagnose/kiagnose/pull/15/files#diff-9c52ee6d71999c83d9e358fc8c604002291553b5351a4c710ac89fdc1b662206R192).
3. [Failure to get a Role](https://github.com/kiagnose/kiagnose/pull/15/files#diff-9c52ee6d71999c83d9e358fc8c604002291553b5351a4c710ac89fdc1b662206R197).

kiagnose/internal/config/loader_test.go
orelmisan: The tests in loader_test.go are also covering 100% of the parser code.

kiagnose/mainflow.go
orelmisan: The validation was moved to Loader.Load().

orelmisan:
Change: Fixed commit messages.

kiagnose/internal/configmap/configmap.go
```diff
@@ -25,9 +25,19 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 )
 
+func GetData(client kubernetes.Interface, namespace, name string) (map[string]string, error) {
```
EdDev, line 32: The commit says it moves the the package up, but in fact you also introduce a new method.

I would just move it, reasoning that it also is to be shared with the config package.
Then, when you actually need that method, to introduce it.

BTW, when you add it, it will fit nicer below the `Create` (just a nit).

EdDev (commented):
First commit

kiagnose/internal/rbac/rbac.go
```diff
@@ -32,10 +32,29 @@ import (
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	rbacv1client "k8s.io/client-go/kubernetes/typed/rbac/v1"
 )
 
+func GetClusterRole(client kubernetes.Interface, name string) (*rbacv1.ClusterRole, error) {
```
EdDev, line 40: Same here, better introduce it when you actually use it.

EdDev (commented):
Second commit.

kiagnose/internal/config/loader.go
```diff
@@ -0,0 +1,141 @@
...
+	if !exists {
+		return nil, fmt.Errorf(envVarErrFmtString, ConfigMapNamespaceEnvVarName)
+	}
+
+	configMapName, exists := l.env[ConfigMapNameEnvVarName]
+	if !exists {
+		return nil, fmt.Errorf(envVarErrFmtString, ConfigMapNameEnvVarName)
+	}
+
+	var err error
```
EdDev, line 61: I think you need this only further below. Try to place it close to it being needed.

kiagnose/internal/config/loader.go
```diff
@@ -0,0 +1,141 @@
...
+
+	rawData, err := configmap.GetData(l.client, configMapNamespace, configMapName)
+	if err != nil {
+		return nil, err
+	}
+
+	parser := newConfigMapParser(rawData)
+	err = parser.Parse()
+	if err != nil {
```
EdDev, line 70: Can be inlined

EdDev (commented):
commit 3

kiagnose/internal/configmap/configmap.go
orelmisan: Done.

kiagnose/internal/rbac/rbac.go
orelmisan: Done.

kiagnose/internal/config/loader.go
orelmisan: Without this, one linter fails, without the previous one, a different linter fails.

kiagnose/internal/config/loader.go
EdDev: In this case, I think it is actually very easy to keep the parser internal and not expose it as an API.
By doing so, the tests are not aware it even exists and therefore should not be affected if it changes (or replaced).

There are cases where this is not possible (e.g. the k8s client) due to external dependencies, but here it is.
It is also easy to check all permutations through the `Loader` API, at least with the current input/output logic options.

orelmisan:
Change: Changed commit messages.

kiagnose/internal/config/loader_test.go
EdDev: The client errors seem to be covered by just making sure the fetched object is not in the fake client.
It was illusive to me too at first.

kiagnose/internal/config/loader.go
EdDev: Why would `if err := parser.Parse(); err != nil {...}` fail?

orelmisan:
```
if err := parser.Parse(); err != nil {
	return nil, err
}
```

```
./automation/make.sh 
kiagnose/internal/config/loader.go:68:5: shadow: declaration of "err" shadows declaration at line 61 (govet)
        if err := parser.Parse(); err != nil {
           ^
```

EdDev: Pfff... just a matter of naming (and line 61 should have been removed).
Not going to block on this nit.

EdDev (approved):
Thank you!

kiagnose/internal/config/loader_test.go
```diff
@@ -0,0 +1,290 @@
...
+			panic("failed to create ClusterRole")
+		}
+	}
+
+	for i := range clusterRoles {
+		_, err := client.RbacV1().Roles(roles[i].Namespace).Create(context.Background(), roles[i], metav1.CreateOptions{})
+		if err != nil {
+			panic("failed to create Role")
+		}
+	}
```
ormergi, line 235: I think this can be simplified by passing `[]ClusterRoles` and `[]Roles` directly to `fake.NewSimpleClientset()`.

kiagnose/internal/config/loader_test.go
```diff
@@ -0,0 +1,290 @@
...
+			configMapData: map[string]string{config.ImageKey: imageName, config.TimeoutKey: timeoutValue, config.ClusterRolesKey: "NA\n"},
+			expectedError: "clusterroles.rbac.authorization.k8s.io",
+		},
+		{
+			description:   "when Role doesn't exist",
+			configMapData: map[string]string{config.ImageKey: imageName, config.TimeoutKey: timeoutValue, config.RolesKey: "default/role999\n"},
+			expectedError: "roles.rbac.authorization.k8s.io",
+		},
+		{
+			description:   "when Role name is illegal",
```
ormergi, line 202: Do you think it make sense to cover more cases of `RolesKey`, such as `/myrole` `myns/` `ns//role` `""`?

What you think about adding similar test for `ClusterRoleKey` does it make sense?

kiagnose/mainflow.go
```diff
@@ -41,7 +39,6 @@ func Run(env map[string]string) error {
 		return err
 	}
 
-	log.Printf("Checkup configuration: %+v", checkupConfig)
-	l := launcher.New(checkup.New(c, "", 0, nil, nil, nil), reporter.New())
```
ormergi, line 34: Looks like a leftover

ormergi:
@orelmisan 
Thanks for the changes it look much better!

I have some more comments, most of them are suggestions to simplify the code even more,
Please take a look and see if it make sense to do.

kiagnose/mainflow.go
EdDev: ?

EdDev (commented):
I answered here in an attempt to separate the nits from the ones which merit attention once detected.
Looking forward to see this finalized.

kiagnose/internal/config/cmparser.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func newConfigMapParser(configMapRawData map[string]string) *configMapParser {
+	return &configMapParser{
+		configMapRawData: configMapRawData,
+		params:           map[string]string{},
+	}
+}
+
+func (cmp *configMapParser) Parse() error {
```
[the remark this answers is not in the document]
orelmisan: The parser internal state is different than the Config struct.
The image and timeout fields are the same but:
1. The parser holds raw map of strings for params, while Config holds []EnvVar.
2. The parser holds names of ClusterRole and Role objects, while Config holds a slices of ClusterRole and Role.

kiagnose/internal/config/cmparser.go
```diff
@@ -0,0 +1,150 @@
...
+	ClusterRolesKey    = "spec.clusterRoles"
+	RolesKey           = "spec.roles"
+)
+
+var (
+	ErrConfigMapDataIsNil    = errors.New("configMap Data field is nil")
+	ErrImageFieldIsMissing   = errors.New("image field is missing")
+	ErrTimeoutFieldIsMissing = errors.New("timeout field is missing")
+	ErrTimeoutFieldIsIllegal = errors.New("timeout field is illegal")
+	ErrRoleNameIsIllegal     = errors.New("illegal Role name")
```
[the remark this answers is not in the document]
orelmisan: This allows us to publish (some) of the errors that this package produces.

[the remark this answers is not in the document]
ormergi: I am not talking about formatting strings, just extracting the message to const and create the error where needed:
```
const errNotValid = "not valid"
...
func foo () err {
    return errors.New("errNotValid")
}
```
Currently we do not provide errors as part of the API, isnt it?

kiagnose/internal/config/loader.go
```diff
@@ -0,0 +1,141 @@
...
+	client kubernetes.Interface
+	env    map[string]string
+}
+
+func NewLoader(client kubernetes.Interface, env map[string]string) *Loader {
+	return &Loader{client: client, env: env}
+}
+
+func (l *Loader) Load() (*Config, error) {
+	const envVarErrFmtString = "missing required environment variable %q"
```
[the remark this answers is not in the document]
orelmisan: Done.

kiagnose/internal/config/loader.go
```diff
@@ -0,0 +1,141 @@
...
+	err = parser.Parse()
+	if err != nil {
+		return nil, err
+	}
+
+	checkupConfig := &Config{
+		Image:   parser.Image(),
+		Timeout: parser.Timeout(),
+		EnvVars: paramsToEnvVars(parser.Params()),
+	}
```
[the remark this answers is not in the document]
orelmisan: Done.

kiagnose/internal/config/cmparser.go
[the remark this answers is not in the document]
ormergi: > Considering that the parsing includes multiple functions, each requiring to pass the raw data, the usage of receivers seems reasonable to me.

Just to be clear, changing Parse() to static also means getting rid of the helpers receivers so each will be static as well.

I am not insisting on this change.

kiagnose/internal/config/loader.go
```diff
@@ -0,0 +1,141 @@
...
+
+	if checkupConfig.Roles, err = l.getRoles(parser.RoleNames()); err != nil {
+		return nil, err
+	}
+
+	return checkupConfig, nil
+}
+
+func (l *Loader) getClusterRoles(clusterRoleNames []string) ([]*rbacv1.ClusterRole, error) {
```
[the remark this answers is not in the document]
orelmisan: Done.

kiagnose/internal/config/loader.go
```diff
@@ -0,0 +1,141 @@
...
+			return nil, err
+		}
+
+		clusterRoles = append(clusterRoles, clusterRole)
+	}
+
+	return clusterRoles, nil
+}
+
+func (l *Loader) getRoles(roleNames []string) ([]*rbacv1.Role, error) {
```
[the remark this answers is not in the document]
orelmisan: Done.

kiagnose/internal/config/loader_test.go
```diff
@@ -0,0 +1,290 @@
...
+			envVars:               map[string]string{},
+			expectedErrorContains: expectedErrorPrefix,
+		},
+	}
+
+	for _, testCase := range failureTestCases {
+		t.Run(testCase.description, func(t *testing.T) {
+			loader := config.NewLoader(nil, testCase.envVars)
+			_, err := loader.Load()
+			assert.ErrorContains(t, err, testCase.expectedErrorContains)
```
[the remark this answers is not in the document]
ormergi: Errors may change over time meaning it will be necessary to change the tests accordingly, 
I think we should avoid assenting on error message content when possible.

Using `assert.Error()` will simplify the tests since there will be no need to pass `expectedErrorContains`
on each test case.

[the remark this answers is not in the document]
orelmisan: Here we want to differentiate between different errors, not just to assert that an error had occurred.

kiagnose/internal/config/loader_test.go
```diff
@@ -0,0 +1,290 @@
...
+		{
+			description:   "when Role name is illegal",
+			configMapData: map[string]string{config.ImageKey: imageName, config.TimeoutKey: timeoutValue, config.RolesKey: "illegal name\n"},
+			expectedError: config.ErrRoleNameIsIllegal.Error(),
+		},
+	}
+
+	for _, testCase := range failureTestCases {
+		t.Run(testCase.description, func(t *testing.T) {
+			fakeClient := newFakeClientWithObjects(configMapNamespace, configMapName, testCase.configMapData, nil, nil)
```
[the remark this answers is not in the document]
orelmisan: Done.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/fcff5c7929d9f62e7d54623a7d0479367d6560bd..7197d634f0a7cf4d9a7e338ff614d2426ca700b4): Answered code review comments.

kiagnose/internal/config/loader_test.go
```diff
@@ -0,0 +1,290 @@
...
+				EnvVars:      expectedEnvVars(param1Key, param1Value, param2Key, param2Value),
+				ClusterRoles: expectedClusterRoles(),
+				Roles:        expectedRoles(),
+			},
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.description, func(t *testing.T) {
+			fakeClient := newFakeClientWithObjects(configMapNamespace, configMapName, testCase.configMapData, testCase.clusterRoles, testCase.roles)
```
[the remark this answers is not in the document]
orelmisan: I've had problems making this compile.

kiagnose/internal/config/loader_test.go
```diff
@@ -0,0 +1,290 @@
...
+func expectedClusterRoles() []*rbacv1.ClusterRole {
+	return []*rbacv1.ClusterRole{
+		{
+			TypeMeta:   metav1.TypeMeta{Kind: "ClusterRole", APIVersion: rbacv1.GroupName},
+			ObjectMeta: metav1.ObjectMeta{Name: "cluster_role1"},
+		},
+		{
+			TypeMeta:   metav1.TypeMeta{Kind: "ClusterRole", APIVersion: rbacv1.GroupName},
+			ObjectMeta: metav1.ObjectMeta{Name: "cluster_role2"},
+		},
```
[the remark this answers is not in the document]
orelmisan: Can you please elaborate?

kiagnose/internal/config/loader_test.go
```diff
@@ -0,0 +1,290 @@
...
+	return []*rbacv1.Role{
+		{
+			TypeMeta:   metav1.TypeMeta{Kind: "Role", APIVersion: rbacv1.GroupName},
+			ObjectMeta: metav1.ObjectMeta{Name: "role1", Namespace: "default"},
+		},
+		{
+			TypeMeta:   metav1.TypeMeta{Kind: "Role", APIVersion: rbacv1.GroupName},
+			ObjectMeta: metav1.ObjectMeta{Name: "role2", Namespace: "default"},
+		},
+	}
```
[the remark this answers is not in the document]
orelmisan: Can you please elaborate?

kiagnose/internal/config/loader_test.go
orelmisan: I've had problems making it compile.

kiagnose/internal/config/loader_test.go
orelmisan: There is a full ticket regarding configuration validation.
IMO this should be handled there.

kiagnose/mainflow.go
orelmisan: This is a code that was deleted.
What is wrong with it?

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/7197d634f0a7cf4d9a7e338ff614d2426ca700b4..de035b2fcca73c3908c64e56d18e50ab190f19e6): Fixed a small bug in `newFakeClientWithObjects()`

kiagnose/mainflow.go
ormergi: Correct me if I am wrong but this change was introduced on commit #4 and and deleted at commit #5.

kiagnose/internal/config/loader_test.go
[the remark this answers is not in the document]
ormergi:
```
{TypeMeta: metav1.TypeMeta{Kind: "ClusterRole", APIVersion: rbacv1.GroupName},ObjectMeta: metav1.ObjectMeta{Name: "cluster_role1"}},
```

kiagnose/internal/config/loader_test.go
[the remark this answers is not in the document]
ormergi:
```
{TypeMeta: metav1.TypeMeta{Kind: "Role", APIVersion: rbacv1.GroupName}, ObjectMeta: metav1.ObjectMeta{Name: "role2", Namespace: "default"}}
```

kiagnose/mainflow.go
ormergi: Correct me if I am wrong but this change (line 44) was introduced on commit #4 and and deleted at commit #5.

kiagnose/internal/config/cmparser.go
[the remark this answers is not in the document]
ormergi: @EdDev  I am not talking about formatting strings, just extracting the message to const and create the error where needed:
```
const errNotValid = "not valid"
...
func foo () err {
    return errors.New("errNotValid")
}
```
@orelmisan the errors are only used at tests, and they can use the error message instead.
I think this package should not expose the errors until there is a need for it.

kiagnose/internal/config/loader_test.go
ormergi: Sorry for the noise, I checked it again and it works when passing *RoleList and *ClusterRoleList instead of []*Role and []*ClusterRoleList.

kiagnose/mainflow.go
orelmisan: Commit 4 is to introduce the Config loading to mainflow.
Commit 5 is to integrate the config loading with the checkup.

Is there something do you think we should do here?

kiagnose/internal/config/loader_test.go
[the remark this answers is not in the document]
orelmisan: IMO we should check that the error we expect is the actual error we are getting.

kiagnose/internal/config/loader_test.go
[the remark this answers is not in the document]
orelmisan: IMO it makes it less readable.

kiagnose/internal/config/loader_test.go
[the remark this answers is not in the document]
orelmisan: IMO it makes it less readable.

kiagnose/mainflow.go
ormergi: Remove this line https://github.com/kiagnose/kiagnose/pull/15/commits/79420429daaf952650e57f41cb4f37c7a47ace58#diff-4ec73023f96b17dcaab6fb900294476b4a985df1fac89d1545944edc5fc6e163R44

kiagnose/internal/config/cmparser.go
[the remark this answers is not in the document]
ormergi: @orelmisan I didnt meant to use `configMapParser` as the result of Parse() not `Config` struct.

kiagnose/mainflow.go
orelmisan: It won't compile without it.
Unless we do:
```go
_, err := configLoader.Load()
```

kiagnose/internal/config/cmparser.go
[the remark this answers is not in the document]
orelmisan: IIUC you want the configMapParser (lets say ParseResults) as the result of Parse().
We can do it in a future PR.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/de035b2fcca73c3908c64e56d18e50ab190f19e6..4cd6195c2b5ec591ef11d3f2fad0210d3e9642ea): Fixed a review comment.

kiagnose/mainflow.go
orelmisan: @ormergi Done.

ormergi (approved):
Thanks for the changes, good work!

There are these two comments https://github.com/kiagnose/kiagnose/pull/15#discussion_r852673792
https://github.com/kiagnose/kiagnose/pull/15#discussion_r852173990, if you think it will take more work then it seems we can do it on a follow up PR.

kiagnose/internal/config/cmparser.go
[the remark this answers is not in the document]
ormergi: Fair enough

kiagnose/internal/config/cmparser.go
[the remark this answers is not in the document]
ormergi: @orelmisan @EdDev thanks for explaining.

Looking at this again, I was confused since on tests the assertions is done on the error message content and not on its type, I guess we can improve them later on.

kiagnose/internal/config/loader.go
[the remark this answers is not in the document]
ormergi: Sorry I wasn't clear about saying this is a nit.

kiagnose/internal/config/loader.go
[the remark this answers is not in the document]
ormergi: Great, thanks!

kiagnose/internal/config/loader.go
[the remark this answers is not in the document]
ormergi: Great, thanks it looks much better.

kiagnose/internal/config/loader.go
ormergi: @EdDev understood, thanks for explaining!

kiagnose/internal/config/loader_test.go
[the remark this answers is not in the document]
EdDev: The cleanest solution would have been to assert against an error type or instance.
You can introduce improvements to this in following changes if you like, at this stage I do not consider this a blocker.

kiagnose/internal/config/loader_test.go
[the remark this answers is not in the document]
EdDev: I agree this is an improvement, but I will take this change in even without it.
Feel free to post this change in a following PR.

kiagnose/internal/config/loader_test.go
EdDev: The code was introduced in this change, so tests should have been covering various scenarios of it here.
Said that, I will take the chance and take this change in even without this part.
Consider improving this in a following change.

EdDev (approved):
Thank you!

I have considered the still open issues from the reviews and I think it is safe to merge this as is.
Thank you @ormergi for the thorough review.

PR #21, opened 2022-04-18
cmd: Add main function

orelmisan:
This adds a main function to kiagnose.
This is a required step in order to build an image.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -51,3 +51,16 @@ jobs:
         go-version: ${{ matrix.version }}
     - name: Run unit tests
       run: ./automation/make.sh --unit-test
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    needs: go-versions
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+      - name: Setup Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ needs.go-versions.outputs.minimal }}
+      - name: Run build
+        run: ./automation/make.sh --build-core
--- a/automation/make.sh
+++ b/automation/make.sh
@@ -19,8 +19,10 @@
 
 set -e
 
+BINARY_NAME="kiagnose"
+
 options=$(getopt --options "" \
-    --long lint,unit-test,help\
+    --long lint,unit-test,build-core,help\
     -- "${@}")
 eval set -- "$options"
 while true; do
@@ -30,6 +32,9 @@ while true; do
         ;;
     --unit-test)
         OPT_UNIT_TEST=1
+        ;;
+    --build-core)
+        OPT_BUILD_CORE=1
         ;;
     --help)
         set +x
@@ -44,9 +49,10 @@ while true; do
     shift
 done
 
-if  [ -z "${OPT_LINT}" ] && [ -z "${OPT_UNIT_TEST}" ]; then
+if  [ -z "${OPT_LINT}" ] && [ -z "${OPT_UNIT_TEST}" ] && [ -z "${OPT_BUILD_CORE}" ]; then
     OPT_LINT=1
     OPT_UNIT_TEST=1
+    OPT_BUILD_CORE=1
 fi
 
 if [ -n "${OPT_LINT}" ]; then
@@ -59,4 +65,10 @@ fi
 
 if [ -n "${OPT_UNIT_TEST}" ]; then
     go test -v ./kiagnose/...
+fi
+
+if [ -n "${OPT_BUILD_CORE}" ]; then
+  echo "Trying to build \"${BINARY_NAME}\"..."
+  go build -v -o ./bin/${BINARY_NAME} ./kiagnose/cmd/
+  echo "Successfully built \"${BINARY_NAME}\""
 fi
--- a/kiagnose/cmd/main.go
+++ b/kiagnose/cmd/main.go
@@ -1,0 +1,34 @@
+/*
+ * This file is part of the kiagnose project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright 2022 Red Hat, Inc.
+ *
+ */
+
+package main
+
+import (
+	"log"
+	"os"
+
+	"github.com/kiagnose/kiagnose/kiagnose"
+)
+
+func main() {
+	env := EnvToMap(os.Environ())
+	if err := kiagnose.Run(env); err != nil {
+		log.Fatalf("%s had failed: %v", os.Args[0], err)
+	}
+}
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 # Binaries for programs and plugins
+bin/
 *.exe
 *.exe~
 *.dll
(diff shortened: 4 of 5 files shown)


kiagnose/cmd/main.go
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package main
+
+import (
+	"log"
+	"os"
+
+	"github.com/kiagnose/kiagnose/kiagnose"
+	"github.com/kiagnose/kiagnose/kiagnose/internal/environment"
```
EdDev, line 27: This is not suppose to compile. The whole point of having `internal` was to make sure no one can do this.

Anyway, I do not think it deserve to exist there, it can be placed in the cmd IMO and does not require direct testing as (almost) no logic is involved. If you still think it is worth having the unit tests, then just place them under `kiagnose` with the code.

kiagnose/cmd/main.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+func main() {
+	env := environment.EnvToMap(os.Environ())
+	if err := kiagnose.Run(env); err != nil {
+		log.Fatalln(err.Error())
```
EdDev, line 33: Maybe you should add some high level description that `kiagnose` execution failed.

EdDev (commented):
Thanks, nice and simple.

We need in the same PR the CI coverage as well.
That means:
- Scripts executing the build.
- CI running the script (build).

kiagnose/internal/environment/environment.go
```diff
@@ -0,0 +1,33 @@
...
+
+package environment
+
+import "strings"
+
+func EnvToMap(rawEnv []string) map[string]string {
+	env := map[string]string{}
+
+	for _, entry := range rawEnv {
+		splitKeyValue := strings.Split(entry, "=")
```
EdDev, line 28: I think there is a need to check that it of len 2, otherwise return an error or just ignore.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/556e84e6a0ae5be5ed88d47a5b1ee017dece36e5..98b9020aa0e0837fb88eec2a7ac839d24acbd7d5): Answered code review comments.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/98b9020aa0e0837fb88eec2a7ac839d24acbd7d5..fee2f67a86259fbde912610b88d12470aa347601): Removed magic number.

kiagnose/internal/environment/environment.go
orelmisan: Done.

kiagnose/cmd/main.go
orelmisan: I've added a description with the following format:
`$PROGRAM_NAME had failed: $ERR_DESCRIPTION`
Does that works for you?

kiagnose/cmd/main.go
orelmisan: Removed `environment_test.go`.
Moved `environment.go` under `cmd`.

automation/make.sh
```diff
@@ -31,6 +33,9 @@ while true; do
     --unit-test)
         OPT_UNIT_TEST=1
         ;;
+    --build-core)
+        BUILD_CORE=1
```
EdDev, line 37: Lets keep the same format pattern: `OPT_BUILD_CORE`

.github/workflows/check.yaml
```diff
@@ -51,3 +51,19 @@ jobs:
         go-version: ${{ matrix.version }}
     - name: Run unit tests
       run: ./automation/make.sh --unit-test
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    needs: go-versions
+    strategy:
+      matrix:
+        version: ${{ fromJSON(needs.go-versions.outputs.matrix) }}
```
EdDev, line 60: I think we should be ok with a single version. WDYT?
We already check the unit tests with multiple version of golang, so I do not think we need this one checked as well.

EdDev (changes requested):
Nice.

The reason it fails is because it tried to run checks that it should not.
Just need to fix the default condition:
```diff
-if  [ -z "${OPT_LINT}" ] && [ -z "${OPT_UNIT_TEST}" ]; then
+if  [ -z "${OPT_LINT}" ] && [ -z "${OPT_UNIT_TEST}" ] && [ -z "${OPT_BUILD_CORE}" ]; then
     OPT_LINT=1
     OPT_UNIT_TEST=1
+    OPT_BUILD_CORE=1
 fi
 ```

**Comment by EdDev** [comment]
Here are the changes needed:

- https://github.com/kiagnose/kiagnose/pull/23/commits/aeb669039f15b6947f1a00bd342d49cebbf1bf29
- https://github.com/kiagnose/kiagnose/pull/23/commits/729642cb03537ecfce7283bd8f69673b51002397#diff-78ef9dc8d6ce0002e26fc4adde9d00123acd7a7c214da456a828814850c7d9e2R57-R64

File: kiagnose/cmd/main.go [hunk 6]
```diff
@@ -0,0 +1,34 @@
...
+	"log"
+	"os"
+
+	"github.com/kiagnose/kiagnose/kiagnose"
+)
+
+func main() {
+	env := EnvToMap(os.Environ())
+	if err := kiagnose.Run(env); err != nil {
+		log.Fatalf("%s had failed: %v", os.Args[0], err)
```
ormergi, line 32: Why is it necessary to pass `os.Args[0],` cant we print "kiagnose exit with an error" or something like that?

automation/make.sh
ormergi, line 37: Note that this script run all the operations it provides when no args are being passed to it.
I think it should be initialized so it will be run by default when no args are being passed,
similar to `OPT_LIST`  and `OPT_UNIT_TEST`.

automation/make.sh
```diff
@@ -31,6 +33,9 @@ while true; do
     --unit-test)
         OPT_UNIT_TEST=1
         ;;
+    --build-core)
```
ormergi, line 36: Until we have other stuff to build I think you can call it just `build`.

.github/workflows/check.yaml
```diff
@@ -51,3 +51,19 @@ jobs:
         go-version: ${{ matrix.version }}
     - name: Run unit tests
       run: ./automation/make.sh --unit-test
+  build:
```
ormergi, line 54: Note that when running unit tests it also build the code, unless we use the binary (building a container the runs the binary for example) I dont think we need another Github action.

ormergi (commented):
Nice work

automation/make.sh
orelmisan: Done.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/7deae5a565eea52f20145b2cf33c63bc03e06f0b..1517af07dc70c8dbc94307da26d3d9c94d208d74): Answered code review comments.

kiagnose/cmd/main.go
orelmisan: It gives us the flexibility of not hard coding it.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/1517af07dc70c8dbc94307da26d3d9c94d208d74..b197242d7953ff71e7b194c3bc1004dcdd8340e3): Changed .github/workflows/check.yaml according to https://github.com/kiagnose/kiagnose/commit/729642cb03537ecfce7283bd8f69673b51002397#diff-78ef9dc8d6ce0002e26fc4adde9d00123acd7a7c214da456a828814850c7d9e2R57-R64

automation/make.sh
orelmisan: Done.

automation/make.sh
orelmisan: We have an image to build on PR #22 .

.github/workflows/check.yaml
EdDev: It does not build `main`.
If we will have a step that includes it, we can adjust as needed.

EdDev (dismissed):
Thanks!

EdDev:
@orelmisan , please rebase.

orelmisan:
Change: Rebase.

EdDev (approved):
Thank you!

PR #23, opened 2022-04-18
[WIP] Experimenting with running kubevirt on kind on CI

EdDev:


qinqon:
It's easier to make it run locally with scripts and call those scripts from github actinons, so te amount of github action "code" is as small as possible.

PR #26, opened 2022-04-19
actions, publish: Push the main core image

qinqon:
After a PR is merge the kiagnose core image has to be pushed so users
can consume it. This change push the core kiagnose image on PR merges at
quay.io/kiagnose/kiagnose:main.

Notes for reviewers:
This cannot be tested on the PR since secrets are not passed to jobs running at PRs.
Tested with [act](https://github.com/nektos/act)

### DIFF
--- a/automation/make.sh
+++ b/automation/make.sh
@@ -21,14 +21,17 @@ set -e
 
 CRI=${CRI:-podman}
 
-CORE_BINARY_NAME="kiagnose"
-CORE_IMAGE_REPO=${CORE_IMAGE_REPO:-quay.io}
-CORE_IMAGE_ORG=${CORE_IMAGE_ORG:-kiagnose}
+IMAGE_REGISTRY=${IMAGE_REGISTRY:-quay.io}
+IMAGE_ORG=${IMAGE_ORG:-kiagnose}
+
 CORE_IMAGE_NAME="kiagnose"
 CORE_IMAGE_TAG=${CORE_IMAGE_TAG:-devel}
+CORE_IMAGE="${IMAGE_REGISTRY}/${IMAGE_ORG}/${CORE_IMAGE_NAME}:${CORE_IMAGE_TAG}"
+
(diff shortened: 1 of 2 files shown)


automation/make.sh
```diff
@@ -22,6 +22,8 @@ set -e
 CRI=${CRI:-podman}
 
 CORE_BINARY_NAME="kiagnose"
+CORE_IMAGE_REPO=${CORE_IMAGE_REPO:-quay.io}
```
EdDev, line 25: Isn’t it a registry? The name hints it is a repo.

automation/make.sh
```diff
@@ -45,9 +46,12 @@ while true; do
     --build-core-image)
         OPT_BUILD_CORE_IMAGE=1
         ;;
+    --push-core-image)
+        OPT_PUSH_CORE_IMAGE=1
```
EdDev, line 50: It needs to be added to the defaults, but not triggered automatically.

EdDev (commented):
This is really nice.
Some point inline…

EdDev:
This also needs to be limited/filtered per changes in the cmd and kiagnose folders only. If possible, please excluding tests files will also be nice (but not a must).

qinqon:
> This also needs to be limited/filtered per changes in the cmd and kiagnose folders only. If possible, excluding tests files will also be nice (but not a must).

Not really needed, the time it takes does not affect PRs CI and the image sha is going to be the same so pushing will not overwrite anything.

EdDev:
> Not really needed, the time it takes does not affect PRs CI and the image sha is going to be the same so pushing will not overwrite anything, is better to just delegate container hash comparation on pushing.

Are you sure the build artifact is not affected? Remember that there is no cache between the runs, unlike when run locally.

Anyway, we can take it in and examine as it goes.

EdDev (approved):
Thank you!

PR #27, opened 2022-04-19
Start and wait for a checkup Job to finish

ormergi:
This PR introduce the logic for actually executing a checkup in a cluster.

With this PR changes the framework will create a batch Job to run the checkup image
and waits for it to finish.

A Job is considered as finished when its status has 'Completed' or 'Failed' condition.

### DIFF
--- a/kiagnose/internal/checkup/checkup.go
+++ b/kiagnose/internal/checkup/checkup.go
@@ -29,9 +29,11 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	batchv1client "k8s.io/client-go/kubernetes/typed/batch/v1"
 	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
 	rbacv1client "k8s.io/client-go/kubernetes/typed/rbac/v1"
 
+	"github.com/kiagnose/kiagnose/kiagnose/internal/checkup/job"
 	"github.com/kiagnose/kiagnose/kiagnose/internal/checkup/namespace"
 	"github.com/kiagnose/kiagnose/kiagnose/internal/config"
 	"github.com/kiagnose/kiagnose/kiagnose/internal/configmap"
@@ -41,6 +43,7 @@ import (
 type client interface {
 	CoreV1() corev1client.CoreV1Interface
 	RbacV1() rbacv1client.RbacV1Interface
+	BatchV1() batchv1client.BatchV1Interface
 }
 
 type Checkup struct {
@@ -52,6 +55,7 @@ type Checkup struct {
 	roles               []*rbacv1.Role
 	roleBindings        []*rbacv1.RoleBinding
 	clusterRoleBindings []*rbacv1.ClusterRoleBinding
+	jobTimeout          time.Duration
 	job                 *batchv1.Job
 }
 
@@ -60,17 +64,13 @@ const (
 	ServiceAccountName             = "checkup-sa"
 	ResultsConfigMapName           = "checkup-results"
 	ResultsConfigMapWriterRoleName = "results-configmap-writer"
+	JobName                        = "checkup-job"
+
+	ResultsConfigMapNameEnvVarName      = "RESULT_CONFIGMAP_NAME"
+	ResultsConfigMapNameEnvVarNamespace = "RESULT_CONFIGMAP_NAMESPACE"
 )
 
 func New(c client, checkupConfig *config.Config) *Checkup {
-	const (
-		jobName = "checkup-job"
-
-		resultsConfigMapNameEnvVarName      = "RESULT_CONFIGMAP_NAME"
-		resultsConfigMapNameEnvVarNamespace = "RESULT_CONFIGMAP_NAMESPACE"
-
-		defaultTeardownTimeout = time.Minute * 5
-	)
 	checkupRoles := []*rbacv1.Role{NewConfigMapWriterRole(ResultsConfigMapWriterRoleName, NamespaceName, ResultsConfigMapName)}
 
 	subject := newServiceAccountSubject(ServiceAccountName, NamespaceName)
@@ -80,11 +80,12 @@ func New(c client, checkupConfig *config.Config) *Checkup {
 	}
 
 	checkupEnvVars := []corev1.EnvVar{
-		{Name: resultsConfigMapNameEnvVarName, Value: ResultsConfigMapName},
-		{Name: resultsConfigMapNameEnvVarNamespace, Value: NamespaceName},
+		{Name: ResultsConfigMapNameEnvVarName, Value: ResultsConfigMapName},
+		{Name: ResultsConfigMapNameEnvVarNamespace, Value: NamespaceName},
 	}
 	checkupEnvVars = append(checkupEnvVars, checkupConfig.EnvVars...)
 
+	const defaultTeardownTimeout = time.Minute * 5
 	return &Checkup{
 		client:              c,
 		teardownTimeout:     defaultTeardownTimeout,
@@ -93,8 +94,9 @@ func New(c client, checkupConfig *config.Config) *Checkup {
 		resultConfigMap:     NewConfigMap(ResultsConfigMapName, NamespaceName),
 		roles:               checkupRoles,
 		roleBindings:        checkupRoleBindings,
+		jobTimeout:          checkupConfig.Timeout,
 		clusterRoleBindings: NewClusterRoleBindings(checkupConfig.ClusterRoles, ServiceAccountName, NamespaceName),
-		job: newCheckupJob(jobName,
+		job: NewCheckupJob(JobName,
 			NamespaceName,
 			ServiceAccountName,
 			checkupConfig.Image,
@@ -204,7 +206,7 @@ func newClusterRoleBinding(clusterRoleName string, subject rbacv1.Subject) *rbac
 	}
 }
 
-func newCheckupJob(name, namespaceName, serviceAccountName, image string, activeDeadlineSeconds int64, envs []corev1.EnvVar) *batchv1.Job {
+func NewCheckupJob(name, namespaceName, serviceAccountName, image string, activeDeadlineSeconds int64, envs []corev1.EnvVar) *batchv1.Job {
 	const containerName = "checkup"
 
 	checkupContainer := corev1.Container{
@@ -277,6 +279,12 @@ func (c *Checkup) Setup() error {
 }
 
 func (c *Checkup) Run() error {
+	var err error
+
+	if c.job, err = job.Create(c.client.BatchV1(), c.job); err != nil {
+		return err
+	}
+
 	return nil
 }
 
--- a/kiagnose/internal/checkup/checkup.go
+++ b/kiagnose/internal/checkup/checkup.go
@@ -285,6 +285,10 @@ func (c *Checkup) Run() error {
 		return err
 	}
 
+	if c.job, err = job.WaitForJobToFinish(c.client.BatchV1(), c.job, c.jobTimeout); err != nil {
+		return err
+	}
+
 	return nil
 }
 
--- a/kiagnose/internal/checkup/checkup_test.go
+++ b/kiagnose/internal/checkup/checkup_test.go
@@ -28,11 +28,13 @@ import (
 
 	assert "github.com/stretchr/testify/require"
 
+	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes/fake"
 	clienttesting "k8s.io/client-go/testing"
 
@@ -227,8 +229,9 @@ func TestCheckupTeardownShould(t *testing.T) {
 }
 
 type checkupRunTestCase struct {
-	description string
-	envVars     []corev1.EnvVar
+	description  string
+	envVars      []corev1.EnvVar
+	jobCondition *batchv1.JobCondition
 }
 
 func TestCheckupRunShouldCreateAJob(t *testing.T) {
@@ -239,6 +242,8 @@ func TestCheckupRunShouldCreateAJob(t *testing.T) {
 	for _, testCase := range checkupRunTestCases {
 		t.Run(testCase.description, func(t *testing.T) {
 			testClient := newNormalizedFakeClientset()
+			completeTrueJobCondition := &batchv1.JobCondition{Type: batchv1.JobComplete, Status: corev1.ConditionTrue}
+			testClient.injectJobWatchEvent(newJobWithCondition(checkup.NamespaceName, checkup.JobName, completeTrueJobCondition))
 			testCheckup := checkup.New(testClient, &config.Config{Image: testImage, Timeout: testTimeout, EnvVars: testCase.envVars})
 
 			assert.NoError(t, testCheckup.Setup())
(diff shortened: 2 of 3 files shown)


kiagnose/internal/checkup/checkup_test.go
```diff
@@ -225,6 +227,39 @@ func TestCheckupTeardownShould(t *testing.T) {
...
 }
 
+type checkupRunTestCase struct {
+	description string
+	envVars     []corev1.EnvVar
+}
+
+func TestCheckupRunShouldCreateJob(t *testing.T) {
+	checkupRunTestCases := []checkupRunTestCase{
+		{description: "with no addition env vars"},
```
orelmisan, line 239: *additional

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -225,6 +227,39 @@ func TestCheckupTeardownShould(t *testing.T) {
 	})
 }
 
+type checkupRunTestCase struct {
+	description string
+	envVars     []corev1.EnvVar
+}
+
+func TestCheckupRunShouldCreateJob(t *testing.T) {
```
orelmisan, line 237: nit: Please consider changing this function's name to `TestCheckupRunShouldCreateAJob`

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -225,6 +227,39 @@ func TestCheckupTeardownShould(t *testing.T) {
...
+			testCheckup := checkup.New(testClient, &config.Config{Image: testImage, Timeout: testTimeout, EnvVars: testCase.envVars})
+
+			assert.NoError(t, testCheckup.Run())
+
+			assertJobCreated(t, testClient, testImage, testTimeout, testCase.envVars)
+		})
+	}
+}
+
+func TestCheckupRunShouldFailWhen(t *testing.T) {
```
orelmisan, line 270: Is it possible to simulate the following errors using the fake client?
1. Failure to fetch the image.
2. Checkup exiting with a non zero exit code.

Do we want to simulate these kinds of errors in the unit tests?

orelmisan (commented):
Please see the inline comments for the first commit.

kiagnose/internal/checkup/checkup.go
```diff
@@ -285,6 +285,10 @@ func (c *Checkup) Run() error {
 		return err
 	}
 
+	if c.job, err = job.WaitForJobToFinish(c.client.BatchV1(), c.job, c.jobTimeout); err != nil {
+		return err
+	}
+
 	return nil
```
orelmisan, line 292: I guess that this will be the integration point with the component that will read the checkup results from the results ConfigMap.
We need to think on which errors can we try to read the results.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -225,6 +227,39 @@ func TestCheckupTeardownShould(t *testing.T) {
...
+	checkupRunTestCases := []checkupRunTestCase{
+		{description: "with no addition env vars"},
+		{description: "with additional env vars", envVars: newTestEnvVars()},
+	}
+	for _, testCase := range checkupRunTestCases {
+		t.Run(testCase.description, func(t *testing.T) {
+			testClient := newNormalizedFakeClientset()
+			testCheckup := checkup.New(testClient, &config.Config{Image: testImage, Timeout: testTimeout, EnvVars: testCase.envVars})
+
+			assert.NoError(t, testCheckup.Run())
```
orelmisan, line 247: How is the checkup running successfully without the `Setup()`?

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -225,6 +227,39 @@ func TestCheckupTeardownShould(t *testing.T) {
...
+}
+
+func TestCheckupRunShouldFailWhen(t *testing.T) {
+	t.Run("failed to create Job", func(t *testing.T) {
+		const expectedErr = "failed to create Job"
+		testClient := newNormalizedFakeClientset()
+		testClient.injectCreateErrorForResource(jobResource, expectedErr)
+		testCheckup := checkup.New(testClient, &config.Config{Image: testImage})
+
+		assert.ErrorContains(t, testCheckup.Run(), expectedErr)
```
orelmisan, line 277: How is the checkup running without calling setup()?

kiagnose/internal/checkup/checkup_test.go
ormergi: Done

kiagnose/internal/checkup/job/job.go
```diff
@@ -36,3 +41,54 @@ func Create(client batchv1client.BatchV1Interface, job *batchv1.Job) (*batchv1.J
 	log.Printf("Job '%s/%s' successfully created", job.Namespace, job.Name)
 	return job, nil
 }
+
+func WaitForJobToFinish(client batchv1client.BatchV1Interface, job *batchv1.Job, timeout time.Duration) (*batchv1.Job, error) {
+	const JobNameLabel = "job-name"
+
+	jobLabel := fmt.Sprintf("%s=%s", JobNameLabel, job.Name)
+	jobWatcher, err := client.Jobs(job.Namespace).Watch(context.Background(), metav1.ListOptions{LabelSelector: jobLabel})
```
orelmisan, line 49: Can you please consider using an informer?
It should have an advantage over raw watch, for example it can recover from a momentary communication failure.

kiagnose/internal/checkup/checkup_test.go
orelmisan: Can we differentiate between these problems?

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/f7331e95b4d7953b82ffc44cea0ed81a823eadc4..dba2c4e0f011978ffff344c657f8efd4e2269ce3): Addressed code review comments.
Also use context with timeout when waiting for the checkup job to finish.

kiagnose/internal/checkup/job/job.go
ormergi: As we concluded offline due to the complexity using an informer brings using bare watcher is good enough for now.
I posted new PR for using informer https://github.com/kiagnose/kiagnose/pull/32

kiagnose/internal/checkup/checkup_test.go
EdDev: @orelmisan , why do we need to differentiate between different errors? Is there a different action to take if the error differs?
I would even claim that you do not even covered here all the error types that can be raised, moreover, there may be others added (or removed) in the future.
As I see it, if there is no differentiation on the production code about what to do when a specific error is detected, there is no point in covering it in tests.

kiagnose/internal/checkup/checkup.go
EdDev: I think that the question was about how this "waiting" here integrates with the need to record its run status on the result ConfigMap.

So let me ask this in a more specific manner:
- If the wait ends without a timeout, is the job still around to query its status?

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -278,6 +312,16 @@ func newTestEnvVars() []corev1.EnvVar {
...
 }
 
+func newJobWithCondition(name, namespace string, condition batchv1.JobConditionType) *batchv1.Job {
+	return &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace},
+		Status: batchv1.JobStatus{
+			Conditions: []batchv1.JobCondition{
+				{Type: condition}}}}
```
EdDev, line 322: nit: Can you place this in one line?
It is hard to see the relation between opened and closed braces.. It will be nice to see blocks like this:
```
foo{
    boo{
        zoo{..{}}
    }
}
```

I wander if we can enforce this with a formatter.

kiagnose/internal/checkup/job/job.go
```diff
@@ -36,3 +40,55 @@ func Create(client batchv1client.BatchV1Interface, job *batchv1.Job) (*batchv1.J
 	log.Printf("Job '%s/%s' successfully created", job.Namespace, job.Name)
 	return job, nil
 }
+
+func WaitForJobToFinish(client batchv1client.BatchV1Interface, job *batchv1.Job, timeout time.Duration) (*batchv1.Job, error) {
+	const JobNameLabel = "job-name"
```
EdDev, line 45: Where is this coming from?

kiagnose/internal/checkup/job/job.go
```diff
@@ -36,3 +40,55 @@ func Create(client batchv1client.BatchV1Interface, job *batchv1.Job) (*batchv1.J
...
+
+	jobLabel := fmt.Sprintf("%s=%s", JobNameLabel, job.Name)
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	jobWatcher, err := client.Jobs(job.Namespace).Watch(ctx, metav1.ListOptions{LabelSelector: jobLabel})
+	if err != nil {
+		return nil, err
+	}
+	log.Printf("'%s/%s' Job watcher obtained", job.Namespace, job.Name)
```
EdDev, line 54: Why do we need this? If it was not obtained, we get an error, no?

kiagnose/internal/checkup/job/job.go
```diff
@@ -36,3 +40,55 @@ func Create(client batchv1client.BatchV1Interface, job *batchv1.Job) (*batchv1.J
...
+	for {
+		select {
+		case event := <-eventsCh:
+			job, ok := event.Object.(*batchv1.Job)
+			if !ok {
+				continue
+			}
+			if jobStatus, err := json.MarshalIndent(job.Status, "", " "); err == nil {
+				log.Printf("received job event '%s/%s': \n%v\n", job.Namespace, job.Name, string(jobStatus))
+			}
```
EdDev, line 77: If we will log this for every job change, it will be too verbose IMO.
I do not think this needs to exist in the production code.

kiagnose/internal/checkup/job/job.go
```diff
@@ -36,3 +40,55 @@ func Create(client batchv1client.BatchV1Interface, job *batchv1.Job) (*batchv1.J
...
+			return nil, ctx.Err()
+		}
+	}
+}
+
+func finished(job *batchv1.Job) bool {
+	for _, condition := range job.Status.Conditions {
+		if condition.Type == batchv1.JobComplete || condition.Type == batchv1.JobFailed {
+			return true
+		}
```
EdDev, line 91: `return condition.Type == batchv1.JobComplete || condition.Type == batchv1.JobFailed`

kiagnose/internal/checkup/job/job.go
orelmisan: Thank you for the effort.

kiagnose/internal/checkup/job/job.go
```diff
@@ -36,3 +40,55 @@ func Create(client batchv1client.BatchV1Interface, job *batchv1.Job) (*batchv1.J
...
+
+	return finishedJob, nil
+}
+
+func waitForJob(ctx context.Context, watcher k8swatch.Interface) (*batchv1.Job, error) {
+	eventsCh := watcher.ResultChan()
+	defer watcher.Stop()
+	for {
+		select {
+		case event := <-eventsCh:
```
EdDev, line 70: `watcher.Stop()` closes the channel?

EdDev (changes requested):

kiagnose/internal/checkup/checkup_test.go
orelmisan: Can we deduce it from another source?

This is important for loading (or not) the checkup's results from its ConfigMap.

kiagnose/internal/checkup/checkup.go
orelmisan: This could lead to reading an empty ConfigMap object.

kiagnose/internal/checkup/checkup_test.go
orelmisan: I just don't understand how the checkup is run without its supporting objects created by Setup().
Can you please elaborate?

kiagnose/internal/checkup/job/job.go
ormergi: Note that Job status conditions can be `Completed` `Failed` or `Suspended`, other than that it seems that its possible to add any other arbitrary condition to the Job status (according to doc status.conditions can be patched)
https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/job-v1/#JobStatus

ormergi (commented):
@orelmisan @EdDev please take another look I resolved some code review comments and answered questions.

kiagnose/internal/checkup/job/job.go
EdDev: Based on that information, this is not complete then. Right?

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -389,6 +433,22 @@ func assertClusterRoleBindingsCreated(t *testing.T, testClient testsClient, clus
 	assert.Subset(t, actualClusterRoleBindings, expectedClusterRoleBindings)
 }
 
+func assertJobCreated(t *testing.T, testClient *testsClient, image string, timeout time.Duration, envVars []corev1.EnvVar) {
+	gvr := schema.GroupVersionResource{Group: batchv1.GroupName, Version: "v1", Resource: jobResource}
+	actualJob, err := testClient.Tracker().Get(gvr, checkup.NamespaceName, checkup.JobName)
```
EdDev, line 500: Is it possible to use a simple `Get` here? Do you need the `Tracker()` in this check?

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -389,6 +433,22 @@ func assertClusterRoleBindingsCreated(t *testing.T, testClient testsClient, clus
 	assert.Subset(t, actualClusterRoleBindings, expectedClusterRoleBindings)
 }
 
+func assertJobCreated(t *testing.T, testClient *testsClient, image string, timeout time.Duration, envVars []corev1.EnvVar) {
```
EdDev, line 498: This helper does 3 things, hiding a lot of things from the tests, which in this case could have assisted the reader to understand better.

WDYT about this (at the test):
```
createdJob := ...
expectedJob := ...
assert.Equal(t, createdJob, expectedJob)
```

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -267,6 +295,12 @@ func TestCheckupRunShouldFailWhen(t *testing.T) {
 		assert.NoError(t, testCheckup.Teardown())
 		assertNoObjectExists(t, testClient)
 	})
+
+	t.Run("Job wont finish on time", func(t *testing.T) {
+		testCheckup := checkup.New(fake.NewSimpleClientset(), &config.Config{Image: testImage, Timeout: time.Nanosecond})
+
+		assert.ErrorContains(t, testCheckup.Run(), context.DeadlineExceeded.Error())
```
EdDev, line 302: It will be better to assert on the error type/indentity if possible. Is it possible without changing the production code?

EdDev (changes requested):

kiagnose/internal/checkup/job/job.go
ormergi: From what I tested a Job is considered to be finished when `Complete` or `Failed` conditions exists,
this is why we look for these conditions at Job status.

kiagnose/internal/checkup/job/job.go
ormergi: Done

ormergi (commented):
Addressed code review comments

kiagnose/internal/checkup/job/job.go
EdDev: I think it says you need to check the `conditions.status` value for `true` as well.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/ddd9397e459fcae68131837cbc7de147393b015a..2c5d72ed1160fc06d54ad2820f72cb4778a40eee): Check conditions.status when watching the checkup Job.

kiagnose/internal/checkup/job/job.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+	}
+}
+
+func finished(job *batchv1.Job) bool {
+	for _, condition := range job.Status.Conditions {
+		if condition.Type == batchv1.JobComplete || condition.Type == batchv1.JobFailed &&
+			condition.Status == corev1.ConditionTrue {
+			return true
+		}
```
EdDev, line 92: This seems incorrect to me: I think the `||` logical expression needs to be parenthesized.

Please check how you can cover this bug in the tests.

kiagnose/internal/checkup/checkup_test.go
EdDev: Lets proceed as is, this is good enough.

kiagnose/internal/checkup/checkup.go
ormergi: Sorry for the confusion this is correct ttlSecondsAfterFinished will trigger the job to delete and activeDeadlineSeconds trigger it to stop.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/013bd721262d165867fc2c756baf6be0b1a2ab61..65acadcccf133d5b606016f10f9056f006367250): Change the injected Job event condition definition to be explicit on tests, and cover the case that was mentioned at https://github.com/kiagnose/kiagnose/pull/27#discussion_r858336398

EdDev (approved):
Thank you!

PR #32, opened 2022-04-20
[WIP] Monitor job with informer

ormergi:
Currently we use bare watcher in order to monitor the checkup Job
and get an indication that it finished.
Using bare watcher does not tolerate disconnections with the cluster
API that may cause it to miss events and eventually report that
checkup timed out where it actually finished successfully.

This PR depends on #27 (first two commits)


EdDev:
This PR is no longer relevant.

PR #33, opened 2022-04-21
automation: Push kubevirt-vm-latency image

qinqon:
This check has it's own directory so the github actions use it as
workdir also the other actions has being moved to files with similar
names.

Signed-off-by: Enrique Llorente <ellorent@redhat.com>

### DIFF
--- a/.github/workflows/checkup-kubevirt-vm-latency.publish.yaml
+++ b/.github/workflows/checkup-kubevirt-vm-latency.publish.yaml
@@ -1,0 +1,34 @@
+name: checkup-kubevirt-vm-latency.publish
+
+on:
+  push:
+    branches: [ main ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    env:
+      CHECKUP_IMAGE_TAG: main
+      CRI: podman
+      CHECKUP: kubevirt-vm-latency
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+      - uses: arnested/go-version-action@v1
+        id: go-version
+      - uses: actions/setup-go@v2
+        with:
+          go-version: ${{ steps.go-version.outputs.minimal }}
+      - name: Logging to quay.io
+        run: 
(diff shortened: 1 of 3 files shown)


.github/workflows/checkup-kubevirt-vm-latency.publish.yaml
```diff
@@ -0,0 +1,43 @@
...
+    branches: [ main ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  go-versions:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.versions.outputs.matrix }}
```
EdDev, line 14: I think this is not needed.

.github/workflows/core-and-checkup-echo.check.yaml
```diff
@@ -1,4 +1,4 @@
-name: checks
+name: core-and-checkup-echo.checks
```
EdDev, line 1: Can we leave the parts that are not related to the latency checkup out of this PR?

I am not a fan of this two things name, but we can discuss it later.

.github/workflows/core-and-checkup-echo.publish.yaml
```diff
@@ -1,4 +1,4 @@
-name: publish
+name: core-and-checkup-echo.publish
```
EdDev, line 1: Same

EdDev (commented):
Thank you! Just a small request inline.

.github/workflows/checkup-kubevirt-vm-latency.publish.yaml
qinqon: Is the same mechanism we use to build at check.

EdDev: But there we need both (for unit tests), here we so not.
It is a nit.

qinqon: Ahh ok.

EdDev (approved):
Thank you!

PR #34, opened 2022-04-24
config: Refactor ConfigMap loading process

orelmisan:
This PR extracts loading of the ConfigMap's full name to a new file in the config package "environment.go".
This was done in order to consume the ConfigMap's full name in the future "reporter" entity.

Furthermore, This PR also removes the config.Loader struct, and replaces it with `config.NewFromConfigMap()`.
This was done because the first commit left the config.Loader struct only with a `client` member.


orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/8ed44fcbc5716cdbc89d7ac07237ca2bbc215b76..4687c5b9bb899dc21a86da19641152292d1fcf1c): Changed second commit's subject.

kiagnose/internal/config/config.go
```diff
@@ -33,3 +39,47 @@ type Config struct {
 	ClusterRoles []*rbacv1.ClusterRole
 	Roles        []*rbacv1.Role
 }
+
+func NewFromConfigMap(client kubernetes.Interface, configMapNamespace, configMapName string) (*Config, error) {
```
ormergi, line 43: Could please consider defining a client interface in order to declare explicitly which k8s API's are being used, similar to `checkup` package https://github.com/kiagnose/kiagnose/blob/main/kiagnose/internal/checkup/checkup.go#L41 ?

ormergi (commented):
Hi Orel, I like the new changes it make the more simple and cleaner

kiagnose/internal/config/config.go
orelmisan: The client definition was moved as-is from `NewLoader()` to `NewFromConfigMap()`, so there was no change.
If we want to make this change - IMO it should be in another PR.

TBH, at this point in time I don't see an advantage in limiting the clientset's interface.

EdDev (commented):
I got a bit lost here.
I do not understand why not just add an exposed function that extracts the config-map from the env in the `config` package and that is it. There are a lot of changes here with no clear benefit.

Please try to convince on why we need all these and not just introduce a new function.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/4687c5b9bb899dc21a86da19641152292d1fcf1c..45d6af9cfa473d339746315ce1b87d60831f019e): Removed the "environment" package.

kiagnose/internal/config/config.go
EdDev: I do not think this qualify as an instantiation object now, so using `New` is a bit odd.
It has no methods, so calling it `ReadFromConfigMap` sounds a bit better to me. WDYT?

orelmisan: Sure.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/45d6af9cfa473d339746315ce1b87d60831f019e..788797e0fc29d717cdcdeaf063e3f6069a952eee): Renamed NewFromConfigMap() => ReadFromConfigMap()

EdDev (approved):
Thank you!

ormergi (approved):
Looks good

PR #38, opened 2022-04-27
checkups/echo: Move the test automation scripts under echo

EdDev:
In order to keep the echo checkup as independent as possible,
place its automation scripts under its folder.

The calls from the root automation are kept in order to allow
users of it to migrate.


checkups/echo/README.md
```diff
@@ -26,9 +26,9 @@ status.failureReason: "MESSAGE environment variable is missing"
 - You have [podman](https://podman.io/) or other container engine capable of building images.
 ### Steps
 ```bash
-$ ./automation/make.sh --echo-build-image
+$ ./automation/make.sh --build-checkup-image
```
qinqon, line 29: We are already at a checkup no need to repeat ourselves
```suggestion
$ ./automation/make.sh --build-image
```

checkups/echo/README.md
```diff
@@ -26,9 +26,9 @@ status.failureReason: "MESSAGE environment variable is missing"
...
 # Using Docker to build the image:
-$ CRI=docker ./automation/make.sh --echo-build-image
+$ CRI=docker ./automation/make.sh --build-checkup-image
```
qinqon, line 31: ditto

qinqon (changes requested):
Little nit

checkups/echo/README.md
EdDev: I think that assumes we will have only one image.
Adding `checkup` assures we mean something specific and that we will not need to change it in the future.

(e.g. we could decide to distribute the manifests or other content using images)

checkups/echo/README.md
EdDev: Same

qinqon (approved):

automation/make.sh
```diff
@@ -111,22 +110,26 @@ if [ -n "${OPT_BUILD_CORE_IMAGE}" ]; then
...
-    ${CRI} build ${ECHO_CHECKUP_PATH} --file ${ECHO_CHECKUP_PATH}/Dockerfile --tag "${ECHO_IMAGE}"
+    ./automation/make.sh --build-checkup-image
+    cd -
 fi
 
 if [ -n "${OPT_ECHO_PUSH_IMAGE}" ]; then
-   echo "Pushing \"${ECHO_IMAGE}\"..."
-   ${CRI} push ${ECHO_IMAGE} 
+    cd ./checkups/echo/
+    echo "Pushing \"${ECHO_IMAGE}\"..."
```
orelmisan: nit: Why isn't this line removed in this commit?
It has a counterpart in the echo `make.sh` file.

.github/workflows/publish.yaml
```diff
@@ -39,7 +39,4 @@ jobs:
           ./automation/make.sh \
             --build-core \
             --build-core-image \
-            --push-core-image \
-            --echo-build \
```
orelmisan, line 43: I haven't seen the `echo-build` option in the past.
I guess it was a mistake right?

orelmisan (commented):
Thanks for the PR @EdDev.
There just a few nits in the inline comments.

automation/make.sh
EdDev: Done

automation/make.sh
```diff
@@ -111,22 +110,26 @@ if [ -n "${OPT_BUILD_CORE_IMAGE}" ]; then
...
-  ./entrypoint_test
-  cd -
+    cd ./checkups/echo/
+    ./automation/make.sh --unit-test
+    cd -
 fi
 
 if [ -n "${OPT_ECHO_BUILD_IMAGE}" ]; then
+    cd ./checkups/echo/
     echo "Trying to build image \"${ECHO_IMAGE}\"..."
```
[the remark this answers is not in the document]
EdDev: Done

PR #39, opened 2022-04-27
automation: Use defaults when no args are provided

EdDev:
The current default handling in the automation scripts are not scalable.
Each time a new argument is added, there is a need to also add it
to the default handling condition.

With this change, instead of checking if each specific argument is not
set, a global check is done against the number of arguments the script
is called with. If no arguments are provided, the default options are
triggered.

> *_Note_*: Depends on #38

### DIFF
--- a/automation/make.sh
+++ b/automation/make.sh
@@ -19,6 +19,8 @@
 
 set -e
 
+ARGCOUNT=$#
+
 CRI=${CRI:-podman}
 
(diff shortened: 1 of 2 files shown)


qinqon (changes requested):
Isn't the echo checkup missing here ?

qinqon (dismissed):

orelmisan (approved):
Thank you!

EdDev:
Change: Rebase

PR #40, opened 2022-04-28
kubevirt-vm-latency: Add testify dependencies

ormergi:
This PR adds Testify dependencies to Kubevirt VM latency checkup.


EdDev:
It will be to weird to take this in without usage.
I think we can let this in through the PR/s that use it.

ormergi:
Follow up at #41 

/close

PR #41, opened 2022-04-28
kubevirt-vm-latency: Intorduce skeleton

ormergi:
This PR introduces Kubevirt VM Latency checkup skeleton and entry point, including:
- Client: responsible for creating K8S API client.
- Config: responsible for loading and parsing checkup parameters from env.
- Launcher: responsible for top level logic flow.
- Checkup: responsible for spinning up VM's and measure network latency between them.
- Reporter: responsible for reporting execution results (e.g: network latency measurement results).
- Status: defines the interface between Launcher and Reporter.

All required boilerplate is placed up to interacting with the cluster API,
with this PR changes the checkup will run and exit with the following error: 
`Kubevirt VM latency checkup failed: run: not implemented`.

Interacting with the cluster API and updating the results `ConfigMap` is introduced at #49 

Actual latency check between VM's will be introduced at a follow up PR.


checkups/kubevirt-vm-latency/latencycheck/internal/reporter/reporter.go
```diff
@@ -0,0 +1,27 @@
...
+	"context"
+
+	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/latencycheck/internal/runner"
+)
+
+type reporter struct {
+	context   context.Context
+	client    interface{}
+	name      string
+	namespace string
```
orelmisan, line 13: Please consider adding `configMap` prefix to `name` and `namespace`.

checkups/kubevirt-vm-latency/latencycheck/internal/reporter/reporter.go
```diff
@@ -0,0 +1,27 @@
+package reporter
+
+import (
+	"context"
+
+	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/latencycheck/internal/runner"
+)
+
+type reporter struct {
```
EdDev, line 9: Having this named as the package name is problematic.
One of them needs to be changed.

checkups/kubevirt-vm-latency/latencycheck/internal/reporter/reporter.go
```diff
@@ -0,0 +1,27 @@
+package reporter
+
+import (
+	"context"
+
+	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/latencycheck/internal/runner"
```
EdDev, line 6: This is a dependency which you wanted to avoid by the interface, but you still need it. It is a smell.

EdDev (changes requested):

checkups/kubevirt-vm-latency/latencycheck/internal/runner/runner.go
```diff
@@ -0,0 +1,131 @@
...
+	succeeded      bool
+	failureReasons []string
+	results        map[string]string
+}
+
+func newStatus() *status {
+	return &status{succeeded: false, failureReasons: []string{}, results: map[string]string{}}
+}
+
+func (s *status) Succeed() bool {
```
orelmisan, line 37: Please consider renaming it to `Succeeded()`.

checkups/kubevirt-vm-latency/latencycheck/mainflow.go
```diff
@@ -0,0 +1,37 @@
...
+	"context"
+	"os"
+	"os/signal"
+
+	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/latencycheck/internal/checkup"
+	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/latencycheck/internal/reporter"
+	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/latencycheck/internal/runner"
+)
+
+func Run() error {
```
orelmisan, line 32: I think a configuration fetching logic should be called here (and added as a package).

orelmisan (commented):

ormergi (commented):
[Changes](https://github.com/kiagnose/kiagnose/compare/6ef35deebf7dca0aac7e1473d66993ee9d32bc85..0ec9fa457fc0ab2580077b03eafb8383928ce0ba)

checkups/kubevirt-vm-latency/latencycheck/mainflow.go
```diff
@@ -0,0 +1,37 @@
...
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt)
```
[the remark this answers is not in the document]
ormergi: After struggling Kubevirt client on another PR, it seems that the only client that includes console verb ("kubevirt.io/client-go") wont receive context as parameter and create new context.Background() on each operation.
I will remove it for now.

checkups/kubevirt-vm-latency/latencycheck/internal/runner/runner.go
```diff
@@ -0,0 +1,131 @@
...
+
+type checkup interface {
+	Preflights() error
+	Setup() error
+	Run() error
+	Teardown() error
+	Results() map[string]string
+}
+
+type Report interface {
```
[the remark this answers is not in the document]
ormergi: It seems that there will be dependency between `runner` and `reporter` as one is composited form the other,
Following our discussion offline I will remove this interface, and change Report() to receive `map[string]string` instead.

checkups/kubevirt-vm-latency/latencycheck/internal/runner/runner.go
```diff
@@ -0,0 +1,131 @@
...
+	Report(Report) error
+}
+
+type status struct {
+	succeeded      bool
+	failureReasons []string
+	results        map[string]string
+}
+
+func newStatus() *status {
```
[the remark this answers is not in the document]
ormergi: `status` related code is moved to new file, it also include some changes following other comments.

checkups/kubevirt-vm-latency/latencycheck/internal/runner/runner.go
```diff
@@ -0,0 +1,131 @@
+package runner
```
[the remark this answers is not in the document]
ormergi: To me it sounds better, I think launcher is better fit for "heavier" stuff, I dont mind changing it.

checkups/kubevirt-vm-latency/latencycheck/internal/runner/runner.go
```diff
@@ -0,0 +1,131 @@
...
+	checkup  checkup
+	reporter reporter
+	status   *status
+}
+
+func New(checkup checkup, reporter reporter) runner {
+	return runner{
+		checkup:  checkup,
+		reporter: reporter,
+		status:   newStatus(),
```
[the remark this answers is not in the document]
ormergi: Its not a must for now, I will change it to be local.

checkups/kubevirt-vm-latency/latencycheck/internal/runner/runner.go
```diff
@@ -0,0 +1,131 @@
...
+	FailureReason() error
+	Results() map[string]string
+}
+
+type reporter interface {
+	Report(Report) error
+}
+
+type status struct {
+	succeeded      bool
```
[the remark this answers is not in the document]
ormergi: I will change it accordingly, thanks for the heads up.

checkups/kubevirt-vm-latency/latencycheck/internal/reporter/reporter.go
ormergi: Done

checkups/kubevirt-vm-latency/latencycheck/internal/runner/runner.go
ormergi: Done

ormergi:
@EdDev @orelmisan please take another look

checkups/kubevirt-vm-latency/latencycheck/mainflow.go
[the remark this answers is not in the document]
EdDev: What does that mean? That we cannot gracefully terminate?

checkups/kubevirt-vm-latency/latencycheck/mainflow.go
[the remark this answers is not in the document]
EdDev: > Do you think its worth to add an empty implementation for them?

Yes, it will be easier to understand, reason and avoid reverting stuff.
But this is not a must, up to you.

checkups/kubevirt-vm-latency/latencycheck/internal/runner/runner.go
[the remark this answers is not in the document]
EdDev: Not sure why `map[string]string`, but I will try to see at the new implementation and comment there.

checkups/kubevirt-vm-latency/latencycheck/mainflow.go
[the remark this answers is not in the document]
ormergi: I think that If we use `kubevirt.io/client-go` to console the VM's, the console operation wont terminate gracefully,
other operations could be terminated gracefully when done with different client that receive context's (e.g: controller-runtime).

checkups/kubevirt-vm-latency/latencycheck/mainflow.go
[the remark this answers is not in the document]
ormergi: Done

ormergi:
Changes [1](https://github.com/kiagnose/kiagnose/compare/1dc3139cedd4c1ea5bc06f7fc200fd65aeb521c7..6429e88a1f69b215928e9301bb00d1c3d1bf60be) and [2](https://github.com/kiagnose/kiagnose/compare/6429e88a1f69b215928e9301bb00d1c3d1bf60be..881fe717410ae1792669a48444a487c6f7965d47): Added Client and Config packages with empty implementations

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/1dc3139cedd4c1ea5bc06f7fc200fd65aeb521c7..d856dc9020f56e8cc5f6b70c606840f2648011ef): Added Client and Config packages with empty implementations.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/d856dc9020f56e8cc5f6b70c606840f2648011ef..7fe903f06891ee2aba366fccdd34fe5f0f360456): Define an interface between Checkup, Launcher and Reporter

checkups/kubevirt-vm-latency/go.mod
```diff
@@ -1,3 +1,11 @@
 module github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency
 
 go 1.17
+
+require github.com/stretchr/testify v1.7.1
```
EdDev, line 5: Can you try and redirect it to the root vendor location? It does not make sense to keep this twice.

EdDev (commented):
Reviewed "kubevirt-vm-latency: Add testify dependencies ".

ormergi:
[Changes ](https://github.com/kiagnose/kiagnose/compare/7cb88a96fe25800a358544b4a1708e44c1e93492..cabccc95832e8994aa23dc985819fcfe13c66e38): Get rid of Runner naming

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/cabccc95832e8994aa23dc985819fcfe13c66e38..bedd19b90ec791ebf6341020a33efbec2079e9e1): Add missing files license

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/bedd19b90ec791ebf6341020a33efbec2079e9e1..145c135f722c657d6818187e376489538aa31caf): Fix Launcher tests function name

ormergi:
@EdDev @orelmisan I pushed new commits with those changes we will have the most basic 
checkup that includes loading env var and writing results to ConfigMap.
Please take a look, let me know what you think.
If this is too much I can split it to few PR's.

EdDev:
> @EdDev @orelmisan I pushed new commits with those changes we will have the most basic checkup that includes loading env var and writing results to ConfigMap. Please take a look, let me know what you think. If this is too much I can split it to few PR's.

The linter added notes inline of some problems.
I did start looking at this but you seem to push now frequently. Please notify again when you finished.

checkups/kubevirt-vm-latency/vmlatencycheck/internal/reporter/configmap/configmap.go
```diff
@@ -0,0 +1,56 @@
+/*
```
orelmisan, line 1: IMO this commit should be split to a PR of its own.

orelmisan (commented):
Thank you @ormergi for this PR.

I believe the last three commits should be split to two follow up PRs:
1. Reporting the results.
2. Introducing the Kubevirt client.

orelmisan:
@ormergi Please remove or strike-through `Depends on #40` in the PR description.

checkups/kubevirt-vm-latency/cmd/main.go
```diff
@@ -19,8 +19,14 @@
 
 package main
 
-import "log"
+import (
+	"log"
+
+	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/latencycheck"
```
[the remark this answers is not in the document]
orelmisan: @ormergi  Can you please change `check` to `checkup`?
They don't have the same meaning.

checkups/kubevirt-vm-latency/vmlatencycheck/internal/client/client.go
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright 2022 Red Hat, Inc.
+ *
+ */
+
+package client
+
+type Client struct{}
```
EdDev: This is suppose to be a placeholder for the `kubevirt` client?

checkups/kubevirt-vm-latency/vmlatencycheck/internal/checkup/checkup.go
```diff
@@ -0,0 +1,58 @@
...
+	client  *client.Client
+	config  *config.Config
+	results *status.Results
+}
+
+func New(clt *client.Client, cfg *config.Config) *checkup {
+	return &checkup{client: clt, config: cfg}
+}
+
+func (c *checkup) Preflights() error {
```
EdDev, line 40: I think it can say singular, like the others.

checkups/kubevirt-vm-latency/vmlatencycheck/internal/status/status.go
```diff
@@ -0,0 +1,38 @@
...
+	MeasurementDuration time.Duration
+}
+
+type Status struct {
+	FailureReasons []string
+	*Results
+}
+
+func New() *Status {
+	return &Status{Results: &Results{}}
```
EdDev, line 37: Why it returns a pointer?

checkups/kubevirt-vm-latency/vmlatencycheck/internal/checkup/checkup.go
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+package checkup
+
+import (
+	"errors"
+
+	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/vmlatencycheck/internal/client"
+	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/vmlatencycheck/internal/config"
+	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/vmlatencycheck/internal/status"
```
EdDev, line 27: I'm not sure if this dependency is required.
If it is, we should see it later on and not now.
(and if it can be avoided, it is better)

checkups/kubevirt-vm-latency/vmlatencycheck/internal/config/config.go
```diff
@@ -19,11 +19,107 @@
...
+	ErrInvalidResultsConfigMapNamespace = fmt.Errorf("%q environment variable is invalid", ResultsConfigMapNamespaceEnvVarName)
+	ErrNetworkNameMissing               = fmt.Errorf("%q environment variable is missing", NetworkNameEnvVarName)
+	ErrInvalidNetworkName               = fmt.Errorf("%q environment variable is invalid", NetworkNameEnvVarName)
+	ErrNetworkNamespaceMissing          = fmt.Errorf("%q environment variable is missing", NetworkNamespaceEnvVarName)
+	ErrInvalidNetworkNamespace          = fmt.Errorf("%q environment variable is invalid", NetworkNamespaceEnvVarName)
+)
+
+const (
+	DefaultSampleDurationSeconds         = 5
+	DefaultDesiredMaxLatencyMilliseconds = math.MaxInt
```
EdDev, line 65: To avoid this magic import, can we just use here a hard value like 1 hour?

checkups/kubevirt-vm-latency/vmlatencycheck/internal/config/config.go
```diff
@@ -19,11 +19,107 @@
...
+
+	networkNamespace, exists := env[NetworkNamespaceEnvVarName]
+	if !exists {
+		return nil, ErrNetworkNamespaceMissing
+	}
+	if networkNamespace == "" {
+		return nil, ErrInvalidNetworkNamespace
+	}
+
+	sampleDuration, err := strconv.Atoi(env[SampleDurationSecondsEnvVarName])
```
EdDev, line 105: Seems you ignore invalid values here and in the next one. Why?
If this is intentional, please elaborate in the commit message.

checkups/kubevirt-vm-latency/vmlatencycheck/internal/config/config_test.go
```diff
@@ -0,0 +1,157 @@
...
+	testEnv := map[string]string{
+		config.ResultsConfigMapNameEnvVarName:          "results",
+		config.ResultsConfigMapNamespaceEnvVarName:     "default",
+		config.NetworkNameEnvVarName:                   "blue-net",
+		config.NetworkNamespaceEnvVarName:              "default",
+		config.SampleDurationSecondsEnvVarName:         "",
+		config.DesiredMaxLatencyMillisecondsEnvVarName: "",
+	}
+
+	t.Run("succeed", func(t *testing.T) {
```
EdDev, line 40: There are many reasons it can succeed, so it is not clear what this scenario is trying to check.

checkups/kubevirt-vm-latency/vmlatencycheck/internal/config/config_test.go
```diff
@@ -0,0 +1,157 @@
...
+		testConfig, err := config.New(testEnv)
+		assert.NotNil(t, testConfig)
```
EdDev, line 42: I think you know what to expect, right? If so, please assert against the expectation.

checkups/kubevirt-vm-latency/vmlatencycheck/internal/config/config_test.go
```diff
@@ -0,0 +1,157 @@
...
+		assert.NoError(t, err)
```
EdDev, line 43: Usually the error is asserted first.

checkups/kubevirt-vm-latency/vmlatencycheck/internal/config/config_test.go
```diff
@@ -0,0 +1,157 @@
...
+				config.ResultsConfigMapNamespaceEnvVarName: "default",
+				config.NetworkNameEnvVarName:               "blue-net",
+				config.NetworkNamespaceEnvVarName:          "",
+			},
+		},
+	}
+	for _, testCase := range testCases {
+		t.Run(testCase.description, func(t *testing.T) {
+			testConfig, err := config.New(testCase.env)
+			assert.Nil(t, testConfig)
```
EdDev, line 153: I do not think this needs assertion. On error, the value is not defined and we should not say anything else to the reader.

EdDev (changes requested):
Reviewed until the `kubevirt-vm-latency, config: Load checkup parameters from env` commit (including).
Specifically with the config parsing commit, it would have been helpful you describe in the commit what it includes (which config params, who have defaults and who are mandatory, etc).

checkups/kubevirt-vm-latency/vmlatencycheck/internal/reporter/configmap/configmap.go
```diff
@@ -0,0 +1,56 @@
...
+
+func (e ErrUpdateConfigMap) Error() string {
+	return fmt.Sprintf("failed to update ConfigMap \"%s/%s\": %s", e.namespace, e.name, e.message)
+}
+
+func NewConfigMapUpdateError(namespace, name, message string) ErrUpdateConfigMap {
+	return ErrUpdateConfigMap{namespace: namespace, name: name, message: message}
+}
+
+func UpdateData(ctx context.Context, clt corv1client.CoreV1Interface, namespace, name string, data map[string]string) error {
```
EdDev, line 45: It looks like this function is updating the whole ConfigMap, not only the data. In practice it overrides it.
Therefore, calling it just `Update` is a bit more accurate.

checkups/kubevirt-vm-latency/vmlatencycheck/internal/reporter/configmap/configmap.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corv1client "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+type ErrUpdateConfigMap struct {
```
EdDev, line 31: The reasoning for this error type in the commit message and in the tests usage is not that strong.
There is a single error returned, which means there is no need to differentiate between returned errors.
Perhaps you can explain why asserting for the existence of the error is not enough.

checkups/kubevirt-vm-latency/vmlatencycheck/internal/reporter/reporter_test.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+type testsClient struct{ *fake.Clientset }
+
+func (c *testsClient) failConfigMapUpdate(err error) {
+	reactionFn := func(action clienttesting.Action) (bool, runtime.Object, error) {
+		return true, nil, err
+	}
+	c.PrependReactor("update", "configmaps", reactionFn)
+}
```
EdDev, line 90: This is used once and seems an overkill at this stage.
How about:
```go
func failCondifMapUpdate(c *fake.Clientset, err error) {
    c.PrependReactor("update", "configmaps", func(_ clienttesting.Action) (bool, runtime.Object, error) {
		return true, nil, err
    })
}
```

checkups/kubevirt-vm-latency/vmlatencycheck/internal/reporter/reporter.go
```diff
@@ -20,24 +20,77 @@
...
+
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+
+	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/vmlatencycheck/internal/reporter/configmap"
 	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/vmlatencycheck/internal/status"
 )
 
+type client interface {
+	CoreV1() corev1.CoreV1Interface
+}
```
EdDev, line 38: Why is this needed now? Is it serving something significant?

EdDev (changes requested):
Finished passing over all the commits.

The last two commits can be extracted away to a different PR, reducing the size of this one.

I would also like to see redirections to the root vendor folder and reduce duplication.
We are currently maintaining all in the same monorepo, therefore we better take advantage of it and be consistent with versioning and reduce storage load from the repo.

Lastly, please try to review again the commit messages, some require some fixes to make it simpler and clearer. I will try to be more specific in the next review round.

checkups/kubevirt-vm-latency/go.mod
ormergi: In general I agree they should be separated, at this point in time I this its fine because it should be easy to 
move it to another repo.
Thus I did tried to reuse `testify` from the root vendor directory, but I didn't managed to make it work.
I added replace rule to go.mod: `replace github.com/stretchr/testify => ../../vendor/ ../../vendor/github.com/stretchr/testify`
but got the following error from `go mod tidy`:
```
go: github.com/stretchr/testify@v1.7.1 (replaced by ../../vendor/github.com/stretchr/testify): reading ../../vendor/github.com/stretchr/testify/go.mod: open /home/omergi/go/src/github.com/kiagnose/kiagnose/vendor/github.com/stretchr/testify/go.mod: no such file or directory
```
Note that this would work for non vendored packages.

ormergi (commented):
Addressing "kubevirt-vm-latency: Intorduce skeleton" comments

ormergi:
Changes: [[1]](https://github.com/kiagnose/kiagnose/compare/734bfdcbf3d5e21fb14e13c74bde58fc9b273d75..f765df3d2944d0f2571de641793f0b02c72a2ab4) renaming 
[[2]](https://github.com/kiagnose/kiagnose/compare/f765df3d2944d0f2571de641793f0b02c72a2ab4..bbf6307354b9972fc9bb95f5013d543fd1ab49c3) rename vmlatncycheckup package

checkups/kubevirt-vm-latency/vmlatencycheck/internal/reporter/reporter.go
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package reporter
+
+import (
+	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/vmlatencycheck/internal/client"
+	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/vmlatencycheck/internal/status"
+)
+
+type configMapReporter struct {
```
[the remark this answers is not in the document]
orelmisan: How is this passing in the framework?

checkups/kubevirt-vm-latency/vmlatencycheck/mainflow.go
```diff
@@ -0,0 +1,43 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright 2022 Red Hat, Inc.
+ *
+ */
+
+package vmlatencycheck
```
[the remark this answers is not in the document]
orelmisan: I favor `vmlatency`.

checkups/kubevirt-vm-latency/vmlatencycheck/internal/config/config.go
ormergi: Using `time.Hour` requires some ugly conversions, I will change it to hardcoded hour in milliseconds.

checkups/kubevirt-vm-latency/vmlatencycheck/internal/config/config_test.go
ormergi: The scenario is when both sample duration and desired max latency parameters are missing, 
it is expected that they will set with the default values.
Looking at this test again I see no value in having it, the other two covers it.
I will remove it.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/bbf6307354b9972fc9bb95f5013d543fd1ab49c3..bcdb47b8d954d6e41824e1def37fcc5088b1ca3d): update imports following `vmlatencycheck` rename.

checkups/kubevirt-vm-latency/vmlatencycheck/internal/config/config.go
ormergi: Following our discussion, I change the code to fail if its not valid and added tests.
If the a parameter env var is missing it will still be set with the default.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/bcdb47b8d954d6e41824e1def37fcc5088b1ca3d..ef8e54a2f2e68d7c8c00035c7a3b511e615e7428): Change Config to fail in case optional checkup parameters env vars are invalid (following https://github.com/kiagnose/kiagnose/pull/41#discussion_r867492974)

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/bcdb47b8d954d6e41824e1def37fcc5088b1ca3d..b3883d2a1462362b45fc0926be302cfc022370fc): Change Config to fail in case optional checkup parameters env vars are invalid (following https://github.com/kiagnose/kiagnose/pull/41#discussion_r867492974)

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/b3883d2a1462362b45fc0926be302cfc022370fc..6be8289f9dbb6ad1525b98018fc9a9e784cad08b): Change Config successful create test to assert the Config{} object instead specific attributes, also remove redundant test.

checkups/kubevirt-vm-latency/vmlatencycheck/internal/config/config_test.go
ormergi: Done

checkups/kubevirt-vm-latency/vmlatencycheck/internal/config/config_test.go
ormergi: Done

ormergi:
@EdDev @orelmisan regarding splitting this PR,
The commit "kubevirt-vm-latency, reporter: Format results and update configmap"[ [1]](https://github.com/kiagnose/kiagnose/pull/41/commits/0c3188010e2ff5762d8b9a2912d32d67c76d2881) introduces
tests that requires some k8s dependencies, meaning that moving last two commits out of this PR
will make this PR to not pass CI.
In order to move last two commits to different PR, it is necessary to move out the tests [1] introduces, WDYT?

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/6be8289f9dbb6ad1525b98018fc9a9e784cad08b..035525eb4d1c17c76b9fa15f960eb17024df17a9): 
- The last two commits (introduces Kubevirt client dependencies and integration) are removed and will be moved to follow up PR.
- `Reporter` tests are changed to use uses stub client instead of K8S fake client

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/035525eb4d1c17c76b9fa15f960eb17024df17a9..dbd21bbf439af569b27d336f5edd5af8889c7512): Add checkup test for visibility of its implementation state.

checkups/kubevirt-vm-latency/vmlatency/internal/launcher/launcher_test.go
```diff
@@ -0,0 +1,156 @@
...
+	"errors"
+	"testing"
+
+	assert "github.com/stretchr/testify/require"
+
+	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/vmlatency/internal/launcher"
+	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/vmlatency/internal/status"
+)
+
+func TestLauncherShould(t *testing.T) {
```
orelmisan, line 32: Please consider splitting this function in to three groups:
1. Success scenario.
2. Failures that produces a report.
3. Failures that don't produce a report.

orelmisan (commented):
Second commit.

checkups/kubevirt-vm-latency/vmlatencycheck/internal/config/config.go
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Copyright 2022 Red Hat, Inc.
+ *
+ */
+
+package config
+
+type Config struct {
+	ResultsConfigMapName      string
+	ResultsConfigMapNamespace string
```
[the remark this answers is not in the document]
ormergi: Is there a problem if  `checkup` "knows" these fields?

checkups/kubevirt-vm-latency/vmlatencycheck/internal/reporter/reporter.go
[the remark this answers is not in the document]
orelmisan: Let me rephrase, I've asked how do we have in the framework a struct `reporter.Reporter`?, and here the linter fails it.

checkups/kubevirt-vm-latency/vmlatencycheck/internal/config/config.go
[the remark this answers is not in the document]
orelmisan: These are different configs for different entities.
We can break it in a future PR if its too much work.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/dbd21bbf439af569b27d336f5edd5af8889c7512..a8871f3fd0461fa55fa933f64312a923fe92ed2e): rename `reporter` struct.

checkups/kubevirt-vm-latency/vmlatencycheck/internal/reporter/reporter.go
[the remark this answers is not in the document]
ormergi: Done

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/a8871f3fd0461fa55fa933f64312a923fe92ed2e..433fbdc92567db0aa8fe2bb7fbbe89dc5a312e32): Pass only the necessary parameters to checkup from `Config` ,following https://github.com/kiagnose/kiagnose/pull/41#discussion_r864590878

checkups/kubevirt-vm-latency/vmlatencycheck/internal/config/config.go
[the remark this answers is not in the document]
ormergi: Done

checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
```diff
@@ -0,0 +1,32 @@
...
+
+type CheckupParameters struct{}
+
+type Config struct {
+	ResultsConfigMapName      string
+	ResultsConfigMapNamespace string
+	CheckupParameters
+}
+
+func New(_ map[string]string) (*Config, error) {
```
EdDev: I have seen no evidence or reasoning that it merits the use of a pointer.
If the is no strong reason to use it, please avoid and use the minimum needed.
(an example of a possible reasoning can be: It is passed to other entities as a pointer)

checkups/kubevirt-vm-latency/vmlatency/internal/status/status.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+type Status struct {
+	FailureReasons []string
+	*Results
+}
+
+func New() *Status {
+	return &Status{Results: &Results{}}
+}
```
EdDev, line 38: I do not see a specific reason for this data structure to have a factory. Nor why it needs to be a pointer.
If it is due to some futuristic usage, in a following PR, please reason for it there.

checkups/kubevirt-vm-latency/vmlatency/internal/status/status.go
```diff
@@ -0,0 +1,38 @@
...
+
+type Results struct {
+	MinLatency          time.Duration
+	AvgLatency          time.Duration
+	MaxLatency          time.Duration
+	MeasurementDuration time.Duration
+}
+
+type Status struct {
+	FailureReasons []string
```
EdDev, line 32: I would prefer to see singular names where possible.
E.g. `FailureCondition`.
But this is a nit.

checkups/kubevirt-vm-latency/vmlatency/internal/launcher/launcher.go
```diff
@@ -0,0 +1,93 @@
...
+ */
+
+package launcher
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/vmlatency/internal/status"
```
EdDev, line 27: Seems like too many components (launcher, reporter, checkup) depend on this data, which means all are tightly coupled.
Although it is not a blocker at the moment, the approach taken in `kiagnose` core is nicer IMO.

checkups/kubevirt-vm-latency/vmlatency/internal/launcher/launcher.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func New(checkup checkup, reporter reporter) launcher {
+	return launcher{
+		checkup:  checkup,
+		reporter: reporter,
+	}
+}
+
+func (l *launcher) Run() (runErr error) {
```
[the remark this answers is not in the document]
EdDev: If the method is not going to mutate its members directly, I do not think it should have a pointer receiver.

checkups/kubevirt-vm-latency/vmlatency/internal/launcher/launcher.go
```diff
@@ -0,0 +1,93 @@
...
+	runStatus := status.New()
+
+	defer func() {
+		runStatus.Results = l.checkup.Results()
+		if err := l.reporter.Report(runStatus); err != nil {
+			runStatus.FailureReasons = append(runStatus.FailureReasons, fmt.Sprintf("report: %v", err))
```
EdDev, line 60: IMO the called method should add this `report:` prefix, declaring itself.
I do not think that each potential caller needs to specify it, it is not adding information that is relevant to this abstraction layer.

Same for all the rest.

checkups/kubevirt-vm-latency/vmlatency/internal/launcher/launcher.go
```diff
@@ -0,0 +1,93 @@
...
+	if err := l.checkup.Preflight(); err != nil {
+		runStatus.FailureReasons = append(runStatus.FailureReasons, fmt.Sprintf("preflight: %v", err))
+		return err
+	}
+
+	if err := l.checkup.Setup(); err != nil {
+		runStatus.FailureReasons = append(runStatus.FailureReasons, fmt.Sprintf("setup: %v", err))
+		return err
+	}
+	defer func() {
```
EdDev, line 74: Please separate with a newline above.

checkups/kubevirt-vm-latency/vmlatency/internal/launcher/launcher_test.go
EdDev: Prettifying it can be done later (although, arranging it well and in tables, may help with adding new tests easier).

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
```diff
@@ -0,0 +1,22 @@
...
+	"testing"
+
+	assert "github.com/stretchr/testify/require"
+
+	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/vmlatency/internal/checkup"
+	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/vmlatency/internal/client"
+	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/vmlatency/internal/config"
+)
+
+func TestCheckupShouldFailDueToRunNotImplemented(t *testing.T) {
```
EdDev, line 14: Please drop the `DueToRunNotImplemented`.

Generally speaking, I think this should have been covered through the `launcher` using the production checkup.
Leaving here only scenarios that cannot be reached from the `launcher` in the future.

I am not going to insist on this at the moment, we can improve the tests later.

EdDev (commented):
Reviewed the "Introduce skeleton" commit.

EdDev (commented):
>  kubevirt-vm-latency, config: Load checkup parameters from env
>
> CheckupParameters struct is definded as pointer as checkup.Checkup
> attribute due to linter error regarding its size.

I prefer this is avoided. I disagree with the linter.
See #51 

But even if this will stay in, it makes it hard for me to review two changes, please split into two, one answering the linter and the other adding the needed.

checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
```diff
@@ -19,14 +19,136 @@
...
+func NewInvalidEnvVarError(envVarName string, err error) ErrInvalidEnvVarValue {
+	return ErrInvalidEnvVarValue{
+		message:      fmt.Sprintf("%q environment variable is invalid: %v", envVarName, err),
+		wrappedError: err,
+	}
+}
+
+func (e ErrInvalidEnvVarValue) Error() string { return e.message }
+
+func (e ErrInvalidEnvVarValue) Unwrap() error { return e.wrappedError }
```
EdDev, line 81: Sorry, I (still) do not understand why the custom error type is needed.

checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
```diff
@@ -19,14 +19,136 @@
...
+
+const (
+	DefaultSampleDurationSeconds         = 5
+	DefaultDesiredMaxLatencyMilliseconds = math.MaxInt
+)
+
+func New(env map[string]string) (*Config, error) {
+	if env == nil {
```
EdDev, line 89: I am not sure we need to be that defensive here.
We do not expect `vmlatency` to be used as a library by others, and even if someone will use it like that, we could be more generic and say that if the input is "empty", we return a proper error (and that fits both `nil` and an empty map).

Will that work?

checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
```diff
@@ -19,14 +19,136 @@
...
+		return nil, ErrNetworkNamespaceMissing
+	}
+	if networkNamespace == "" {
+		return nil, ErrInvalidNetworkNamespace
+	}
+
+	var err error
+	sampleDuration := DefaultSampleDurationSeconds
+	if value, exists := env[SampleDurationSecondsEnvVarName]; exists {
+		sampleDuration, err = strconv.Atoi(value)
```
EdDev, line 128: nit: I think it can be inline in the error below.

checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
```diff
@@ -19,14 +19,136 @@
...
+		if err != nil {
+			return nil, NewInvalidEnvVarError(SampleDurationSecondsEnvVarName, err)
+		}
+	}
+
+	desiredMaxLatency := DefaultDesiredMaxLatencyMilliseconds
+	if value, exists := env[DesiredMaxLatencyMillisecondsEnvVarName]; exists {
+		desiredMaxLatency, err = strconv.Atoi(value)
```
EdDev, line 136: same, can be inline.

EdDev (commented):
Reviewed up to the "Load checkup parameters from env" commit.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter.go
```diff
@@ -20,24 +20,73 @@
...
+
 type reporter struct {
-	client             *client.Client
+	client             configMapUpdater
 	configMapName      string
 	configMapNamespace string
 }
 
-func New(clt *client.Client, configMapNamespace, configMapName string) *reporter {
+func New(clt configMapUpdater, configMapNamespace, configMapName string) *reporter {
```
EdDev, line 42: I can't find where I commented about prefering `c` over `clt`.
Could you change it and check if it exists anywhere else?

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter.go
```diff
@@ -20,24 +20,73 @@
...
+var ErrInvalidStatus = errors.New("status is invalid")
+
+func (r *reporter) Report(s *status.Status) error {
+	data, err := formatStatus(s)
+	if err != nil {
+		return err
+	}
+
+	if raw, err := json.MarshalIndent(data, "", " "); err == nil {
+		fmt.Printf("reporting status:\n%s\n", string(raw))
```
EdDev, line 59: We usually do not print directly to `stdout`.
What does it mean if it fails to parse it?

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter.go
```diff
@@ -20,24 +20,73 @@
...
+	}
+
+	return r.client.UpdateConfigMap(r.configMapNamespace, r.configMapName, data)
+}
+
+func formatStatus(s *status.Status) (map[string]string, error) {
+	if s == nil {
+		return nil, ErrInvalidStatus
+	}
```
EdDev, line 68: This is an example of why a pointer is not always good.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter.go
```diff
@@ -20,24 +20,73 @@
...
+		resultAvgLatencyKey          = "status.result.avgLatencyNanoSec"
+		resultMaxLatencyKey          = "status.result.maxLatencyNanoSec"
+		resultMeasurementDurationKey = "status.result.measurementDurationSec"
+	)
+	data := map[string]string{}
+
+	data[succeededKey] = strconv.FormatBool(len(s.FailureReasons) == 0)
+	data[failureReasonKey] = strings.Join(s.FailureReasons, ", ")
+
+	if s.Results != nil {
```
EdDev, line 83: In what scenario this can happen?

EdDev (commented):
Finished review.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/433fbdc92567db0aa8fe2bb7fbbe89dc5a312e32..33b54faaa6dfe8746f6e2f2dba19cfcf38dcdb73): 
- Remove error prefix's from launcher.Run()
- Change Config and Launcher to be used as value instead of pointer
- Move `checkup` failing test to be tested by `launcher` test suite with production `checkup.Checkup{}`.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/33b54faaa6dfe8746f6e2f2dba19cfcf38dcdb73..f0b67712dc94eb6018739a946203178391ad851d): Change `config.Config` pointer usage to value in tests, and fix small issue with launcher test.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/f0b67712dc94eb6018739a946203178391ad851d..c68a6561adc7c6ce957d2106bdea7dbeb535bdff): Rename `FaliureReasons` to singular form.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/db2463857c408c50fa08183e4bf086ad4d629297..eb692a05c3d828906081d0ddcaa68fe5ac4d8186): Change` status.Status{} `pointer usage to value.

checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
ormergi: Thanks for explaining, the reason for this is to return meaningful error if `env` is `nil` instead of `ErrResultsConfigMapNameMissing` which will be the first thing to fail.
Looking at this again if `env` is empty it will return `ErrResultsConfigMapNameMissing,
I can fix so it will return `ErrInvalidEnv` is `env` is nil or empty.

Make sense?

checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
ormergi: Done

checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
ormergi: Done

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter.go
ormergi: > We usually do not print directly to stdout.

Sorry this is a mistake, I meant to to use `log.Printf()`, I changed it.

>What does it mean if it fails to parse it?

It means that `data` is not supported cannot be serialized to json,
it is not likely to happen since `data` is of type `map[string]string` which is supported.
Do you think its work to define `data` explicitly?

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/eb692a05c3d828906081d0ddcaa68fe5ac4d8186..be5c681a8311ec220dd29782af46b1f2d83b8901): Change `config.New()` to return an error in case of nil or empty map, and inline `Atoi` `if` closures at config.New().

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/be5c681a8311ec220dd29782af46b1f2d83b8901..6de22aacbb7e5aca799146977f93e039544b3621): Rename client argument `clt` to `c`, all over

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter.go
ormergi: Done all over.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/6de22aacbb7e5aca799146977f93e039544b3621..872133860dd09c084fdf83aa1dc8a1dfbe760424): Log the checkup results instead of sending it to stdout directly.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter.go
ormergi: In case status.Results is not initialized, status.Results is expected to be initialized by checup.Run().
If checkup.Run() did not executed (for example: due to preflight or setup error) it should be nil,
so that the end results is that only `Succedded` and `FailureReason` appears at the results `ConfigMap`.

ormergi (commented):
@EdDev Thanks for the review.
I addressed most of comments, please take another look.

checkups/kubevirt-vm-latency/vmlatency/internal/status/status.go
EdDev: > As for `Results{}`, it represent the checkup results
> which may be absent if the checkup did not ran, for example due to setup or preflight failure.

I fail to understand why does it matter. If it is `nil` or with defaults (zero values) it should behave the same.
It is also a cognitive burden for me to understand what an embedded pointer means in the struct.

By using pointers all over the complexity of the code increases.
it is hard for me to understand if this is a must or not.

checkups/kubevirt-vm-latency/vmlatencycheck/internal/reporter/reporter_test.go
orelmisan: This was removed right? I can't find it.

checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
EdDev: I think what you can do is check its size (`len`) and but that you solve the problem.

checkups/kubevirt-vm-latency/vmlatency/internal/launcher/launcher.go
```diff
@@ -0,0 +1,93 @@
...
+	Run() error
+	Teardown() error
+	Results() *status.Results
+}
+
+type reporter interface {
+	Report(*status.Status) error
+}
+
+type launcher struct {
```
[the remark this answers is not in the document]
orelmisan: I find it confusing that we are using unexported types from other packages.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter.go
EdDev: Based on the code, you do not do anything with it, even if it fails.
So it seems redundant to me (except maybe for debugging, but I am unsure if that was the intention).

checkups/kubevirt-vm-latency/vmlatency/internal/launcher/launcher_test.go
orelmisan: Fair enough.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter.go
EdDev: I do not see a problem with showing other parameters with a zero value.
But even then you can check for the zero value and not add the burden of a `nil` on the code.

checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
orelmisan: If the map is empty, I would expect to see the first mandatory field failing to be parsed.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter.go
orelmisan: I don't see this code anymore, was it removed?

checkups/kubevirt-vm-latency/vmlatency/mainflow.go
```diff
@@ -0,0 +1,43 @@
...
+	if err != nil {
+		return err
+	}
+
+	cfg, err := config.New(env)
+	if err != nil {
+		return err
+	}
+
+	r := launcher.New(checkup.New(c, cfg.CheckupParameters), reporter.New(c, cfg.ResultsConfigMapNamespace, cfg.ResultsConfigMapName))
```
EdDev: nit: this is a `l`, not a `r`.

checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
ormergi: Alright fair enough, I will do it on a follow up PR.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/872133860dd09c084fdf83aa1dc8a1dfbe760424..ebd09da28917af34caa579982b5d8d33986b6c60): Remove custom error type from `config`.

ormergi:
[Chnages](https://github.com/kiagnose/kiagnose/compare/ebd09da28917af34caa579982b5d8d33986b6c60..a65fb6b7d5a201c69fb1cc98992ab8cc9385a16e): Get rid of using statis.Results as pointer.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter.go
ormergi: I changed Results{} pointer usage to value.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/a65fb6b7d5a201c69fb1cc98992ab8cc9385a16e..e42420990ffa6200b446541a164130d9653d916d): Rename launcher local variable name at mainfliow.go

checkups/kubevirt-vm-latency/vmlatency/mainflow.go
ormergi: Done

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/e42420990ffa6200b446541a164130d9653d916d..dec41cab24e7ba32b2a5a3f31ecbd7c351867b0c): Split converting `CheckupParameters{}` to pointer due to linter error to its own commit, following 
 https://github.com/kiagnose/kiagnose/pull/41#pullrequestreview-968763415

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/8cb23c675523b3970e03c486dde945320556b85d..d4a0b450e0ac724454f1395a3aaf2e3fbc199527): Get rid of pointer receiver at `reporter` tests

EdDev (commented):
Thank you!

> kubevirt-vm-latency, checkup: Change CheckupParameters to pointer 

This commit can be dropped, the linter has been adjusted not to fail now.
If you on the other hand want me to send a revert for this after, I am fine with it.

EdDev (commented):
Thank you!

> kubevirt-vm-latency, checkup: Change CheckupParameters to pointer 

This commit can be dropped, the linter has been adjusted not to fail now.
If you on the other hand want me to send a revert for this after, I am fine with it.

ormergi:
Reabsed on latest main

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/33e818613e253f12309a4ac6cf716ee2240c8bdd..69348970bde2114c5af8b43973e215170f5ebdfd): Get rid of `CheckupParameters` pointer usage.

checkups/kubevirt-vm-latency/vmlatency/internal/launcher/launcher.go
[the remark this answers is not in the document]
ormergi: @orelmisan in general this is a way to restrict how the object will be created, that way there is no need to expose it.

PR #44, opened 2022-04-28
automation/CI: Exclude linting checkups, bump lint version, add verbosity and increase timeout

EdDev:
Optimize and fix lint issues:
- Run root lint on the `kiagnose` and `cmd` folders (excluse `checkups`).
- Bump lint version.
- Add verbosity to the run (to see where is time consumed).
- Increase timeout (as the current one started failing).


orelmisan (approved):
Thank you!

PR #45, opened 2022-04-28
reporter: Add basic reporting capability

orelmisan:
This PR adds basic reporting capability:
1. The `startTimestamp` and `completionTimestamp` are recorded and reported.
2. `succeeded` field is always reported as `false`.
3. `failureReason` field is always reported as `Failure: implement me`.

### DIFF
--- a/kiagnose/internal/launcher/launcher.go
+++ b/kiagnose/internal/launcher/launcher.go
@@ -57,6 +57,7 @@ func (l Launcher) Run() (runErr error) {
 
 	defer func() {
 		statusData.Succeeded = false
+		statusData.FailureReason = "Failure: implement me"
 		statusData.CompletionTimestamp = time.Now()
 
 		if reportErr := l.reporter.Report(statusData); reportErr != nil {
--- a/kiagnose/internal/launcher/launcher.go
+++ b/kiagnose/internal/launcher/launcher.go
@@ -21,6 +21,9 @@ package launcher
 
 import (
 	"fmt"
+	"time"
+
+	"github.com/kiagnose/kiagnose/kiagnose/internal/status"
 )
 
 type workload interface {
@@ -30,7 +33,7 @@ type workload interface {
 }
 
 type reporter interface {
-	Report() error
+	Report(status.Status) error
 }
 
 type Launcher struct {
@@ -46,8 +49,14 @@ func New(checkup workload, reporter reporter) Launcher {
 }
 
 func (l Launcher) Run() (runErr error) {
+	statusData := status.Status{StartTimestamp: time.Now()}
+
+	if err := l.reporter.Report(statusData); err != nil {
+		return err
+	}
+
 	defer func() {
-		if reportErr := l.reporter.Report(); reportErr != nil {
+		if reportErr := l.reporter.Report(statusData); reportErr != nil {
 			if runErr != nil {
 				runErr = fmt.Errorf("%v, %v", runErr, reportErr)
 			} else {
--- a/kiagnose/internal/launcher/launcher.go
+++ b/kiagnose/internal/launcher/launcher.go
@@ -56,6 +56,7 @@ func (l Launcher) Run() (runErr error) {
 	}
 
 	defer func() {
+		statusData.Succeeded = false
 		statusData.CompletionTimestamp = time.Now()
 
 		if reportErr := l.reporter.Report(statusData); reportErr != nil {
--- a/kiagnose/internal/launcher/launcher.go
+++ b/kiagnose/internal/launcher/launcher.go
@@ -56,6 +56,8 @@ func (l Launcher) Run() (runErr error) {
 	}
 
 	defer func() {
+		statusData.CompletionTimestamp = time.Now()
+
 		if reportErr := l.reporter.Report(statusData); reportErr != nil {
 			if runErr != nil {
 				runErr = fmt.Errorf("%v, %v", runErr, reportErr)
--- a/kiagnose/internal/launcher/launcher_test.go
+++ b/kiagnose/internal/launcher/launcher_test.go
@@ -26,81 +26,116 @@ import (
 	assert "github.com/stretchr/testify/require"
 
 	"github.com/kiagnose/kiagnose/kiagnose/internal/launcher"
+	"github.com/kiagnose/kiagnose/kiagnose/internal/status"
 )
 
 func TestLauncherShould(t *testing.T) {
 	t.Run("run successfully", func(t *testing.T) {
-		testLauncher := launcher.New(checkupStub{}, reporterStub{})
+		testLauncher := launcher.New(
+			checkupStub{},
+			&reporterStub{},
+		)
+
 		assert.NoError(t, testLauncher.Run())
 	})
 
-	t.Run("fail when report is failing", func(t *testing.T) {
-		testLauncher := launcher.New(checkupStub{}, reporterStub{failReport: errorReport})
-		assert.ErrorContains(t, testLauncher.Run(), errorReport.Error())
+	t.Run("fail when report on checkup start is failing", func(t *testing.T) {
+		testLauncher := launcher.New(
+			checkupStub{},
+			&reporterStub{reportErr: errorFailOnInitialReport},
+		)
+
+		assert.ErrorContains(t, testLauncher.Run(), errorFailOnInitialReport.Error())
+	})
+
+	t.Run("fail when report on checkup completion is failing", func(t *testing.T) {
+		testLauncher := launcher.New(
+			checkupStub{},
+			&reporterStub{reportErr: errorFailOnFinalReport},
+		)
+
+		assert.ErrorContains(t, testLauncher.Run(), errorFailOnFinalReport.Error())
 	})
 
 	t.Run("fail when setup is failing", func(t *testing.T) {
-		testLauncher := launcher.New(checkupStub{failSetup: errorSetup}, reporterStub{})
+		testLauncher := launcher.New(
+			checkupStub{failSetup: errorSetup},
+			&reporterStub{},
+		)
+
 		assert.ErrorContains(t, testLauncher.Run(), errorSetup.Error())
 	})
 
-	t.Run("fail when setup and report are failing", func(t *testing.T) {
+	t.Run("fail when setup and report on checkup completion are failing", func(t *testing.T) {
 		testLauncher := launcher.New(
 			checkupStub{failSetup: errorSetup},
-			reporterStub{failReport: errorReport},
+			&reporterStub{reportErr: errorFailOnFinalReport},
 		)
+
 		err := testLauncher.Run()
 		assert.ErrorContains(t, err, errorSetup.Error())
-		assert.ErrorContains(t, err, errorReport.Error())
+		assert.ErrorContains(t, err, errorFailOnFinalReport.Error())
 	})
 
 	t.Run("fail when run is failing", func(t *testing.T) {
-		testLauncher := launcher.New(checkupStub{failRun: errorRun}, reporterStub{})
+		testLauncher := launcher.New(
+			checkupStub{failRun: errorRun},
+			&reporterStub{},
+		)
+
 		assert.ErrorContains(t, testLauncher.Run(), errorRun.Error())
 	})
 
 	t.Run("fail when teardown is failing", func(t *testing.T) {
-		testLauncher := launcher.New(checkupStub{failTeardown: errorTeardown}, reporterStub{})
+		testLauncher := launcher.New(
+			checkupStub{failTeardown: errorTeardown},
+			&reporterStub{},
+		)
+
 		assert.ErrorContains(t, testLauncher.Run(), errorTeardown.Error())
 	})
 
-	t.Run("fail when run and report are failing", func(t *testing.T) {
+	t.Run("fail when run and report on checkup completion are failing", func(t *testing.T) {
 		testLauncher := launcher.New(
 			checkupStub{failRun: errorRun},
-			reporterStub{failReport: errorReport},
+			&reporterStub{reportErr: errorFailOnFinalReport},
 		)
+
 		err := testLauncher.Run()
 		assert.ErrorContains(t, err, errorRun.Error())
-		assert.ErrorContains(t, err, errorReport.Error())
+		assert.ErrorContains(t, err, errorFailOnFinalReport.Error())
 	})
 
-	t.Run("fail when teardown and report are failing", func(t *testing.T) {
+	t.Run("fail when teardown and report on checkup completion are failing", func(t *testing.T) {
 		testLauncher := launcher.New(
 			checkupStub{failTeardown: errorTeardown},
-			reporterStub{failReport: errorReport},
+			&reporterStub{reportErr: errorFailOnFinalReport},
 		)
+
 		err := testLauncher.Run()
 		assert.ErrorContains(t, err, errorTeardown.Error())
-		assert.ErrorContains(t, err, errorReport.Error())
+		assert.ErrorContains(t, err, errorFailOnFinalReport.Error())
 	})
 
-	t.Run("fail when run, teardown and report are failing", func(t *testing.T) {
+	t.Run("fail when run, teardown and report on checkup completion are failing", func(t *testing.T) {
 		testLauncher := launcher.New(
 			checkupStub{failRun: errorRun, failTeardown: errorTeardown},
-			reporterStub{failReport: errorReport},
+			&reporterStub{reportErr: errorFailOnFinalReport},
 		)
+
 		err := testLauncher.Run()
 		assert.ErrorContains(t, err, errorRun.Error())
 		assert.ErrorContains(t, err, errorTeardown.Error())
-		assert.ErrorContains(t, err, errorReport.Error())
+		assert.ErrorContains(t, err, errorFailOnFinalReport.Error())
 	})
 }
 
 var (
-	errorSetup    = errors.New("setup error")
-	errorRun      = errors.New("run error")
-	errorTeardown = errors.New("teardown error")
-	errorReport   = errors.New("report error")
+	errorSetup               = errors.New("setup error")
+	errorRun                 = errors.New("run error")
+	errorTeardown            = errors.New("teardown error")
+	errorFailOnInitialReport = errors.New("initial report error")
+	errorFailOnFinalReport   = errors.New("final report error")
 )
 
 type checkupStub struct {
@@ -122,9 +157,20 @@ func (s checkupStub) Teardown() error {
 }
 
 type reporterStub struct {
-	failReport error
+	reportErr   error
+	reportCount int
 }
 
-func (r reporterStub) Report() error {
-	return r.failReport
+func (r *reporterStub) Report(_ status.Status) error {
+	r.reportCount++
+	if r.reportCount > 2 {
+		panic("Report was called more than twice")
+	}
+
+	if r.reportCount == 1 && r.reportErr == errorFailOnInitialReport ||
+		r.reportCount == 2 && r.reportErr == errorFailOnFinalReport {
+		return r.reportErr
+	}
+
+	return nil
 }
--- a/kiagnose/internal/reporter/reporter.go
+++ b/kiagnose/internal/reporter/reporter.go
@@ -19,13 +19,63 @@
 
 package reporter
 
+import (
+	"errors"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/kiagnose/kiagnose/kiagnose/internal/configmap"
+	"github.com/kiagnose/kiagnose/kiagnose/internal/status"
+)
+
+const StartTimestampKey = "status.startTimestamp"
+
+var ErrConfigMapDataIsNil = errors.New("configMap Data is nil")
+
 type Reporter struct {
+	client    kubernetes.Interface
+	configMap *corev1.ConfigMap
 }
 
-func New() Reporter {
-	return Reporter{}
+func New(client kubernetes.Interface, configMapNamespace, configMapName string) *Reporter {
+	return &Reporter{
+		client: client,
+		configMap: &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      configMapName,
+				Namespace: configMapNamespace,
+			},
+		},
+	}
 }
 
-func (r Reporter) Report() error {
+func (r *Reporter) Report(statusData status.Status) error {
+	if r.configMap.Data == nil {
+		configMap, err := configmap.Get(r.client.CoreV1(), r.configMap.Namespace, r.configMap.Name)
+		if err != nil {
+			return err
+		}
+
+		r.configMap = configMap
+	}
+
+	if r.configMap.Data == nil {
+		return ErrConfigMapDataIsNil
+	}
+
+	if !statusData.StartTimestamp.IsZero() {
+		r.configMap.Data[StartTimestampKey] = statusData.StartTimestamp.Format(time.RFC3339)
+	}
+
+	updatedConfigMap, err := configmap.Update(r.client.CoreV1(), r.configMap)
+	if err != nil {
+		return err
+	}
+
+	r.configMap = updatedConfigMap
+
 	return nil
 }
--- a/kiagnose/internal/reporter/reporter.go
+++ b/kiagnose/internal/reporter/reporter.go
@@ -34,6 +34,7 @@ import (
 
 const (
 	SucceededKey           = "status.succeeded"
+	FailureReasonKey       = "status.failureReason"
 	StartTimestampKey      = "status.startTimestamp"
 	CompletionTimestampKey = "status.completionTimestamp"
 )
@@ -78,6 +79,7 @@ func (r *Reporter) Report(statusData status.Status) error {
 	if !statusData.CompletionTimestamp.IsZero() {
 		r.configMap.Data[CompletionTimestampKey] = statusData.CompletionTimestamp.Format(time.RFC3339)
 		r.configMap.Data[SucceededKey] = strconv.FormatBool(statusData.Succeeded)
+		r.configMap.Data[FailureReasonKey] = statusData.FailureReason
 	}
 
 	updatedConfigMap, err := configmap.Update(r.client.CoreV1(), r.configMap)
--- a/kiagnose/internal/reporter/reporter.go
+++ b/kiagnose/internal/reporter/reporter.go
@@ -31,7 +31,10 @@ import (
 	"github.com/kiagnose/kiagnose/kiagnose/internal/status"
 )
 
-const StartTimestampKey = "status.startTimestamp"
+const (
+	StartTimestampKey      = "status.startTimestamp"
+	CompletionTimestampKey = "status.completionTimestamp"
+)
 
 var ErrConfigMapDataIsNil = errors.New("configMap Data is nil")
 
@@ -68,6 +71,10 @@ func (r *Reporter) Report(statusData status.Status) error {
 
 	if !statusData.StartTimestamp.IsZero() {
 		r.configMap.Data[StartTimestampKey] = statusData.StartTimestamp.Format(time.RFC3339)
+	}
+
+	if !statusData.CompletionTimestamp.IsZero() {
+		r.configMap.Data[CompletionTimestampKey] = statusData.CompletionTimestamp.Format(time.RFC3339)
 	}
 
 	updatedConfigMap, err := configmap.Update(r.client.CoreV1(), r.configMap)
--- a/kiagnose/internal/reporter/reporter.go
+++ b/kiagnose/internal/reporter/reporter.go
@@ -21,6 +21,7 @@ package reporter
 
 import (
 	"errors"
+	"strconv"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
@@ -32,6 +33,7 @@ import (
 )
 
 const (
+	SucceededKey           = "status.succeeded"
 	StartTimestampKey      = "status.startTimestamp"
 	CompletionTimestampKey = "status.completionTimestamp"
 )
@@ -75,6 +77,7 @@ func (r *Reporter) Report(statusData status.Status) error {
 
 	if !statusData.CompletionTimestamp.IsZero() {
 		r.configMap.Data[CompletionTimestampKey] = statusData.CompletionTimestamp.Format(time.RFC3339)
+		r.configMap.Data[SucceededKey] = strconv.FormatBool(statusData.Succeeded)
 	}
 
 	updatedConfigMap, err := configmap.Update(r.client.CoreV1(), r.configMap)
--- a/kiagnose/internal/reporter/reporter_test.go
+++ b/kiagnose/internal/reporter/reporter_test.go
@@ -75,10 +75,14 @@ func TestReportShouldSucceed(t *testing.T) {
 		setup()
 
 		assert.NoError(t, reporterUnderTest.Report(checkupStatus))
+
+		checkupStatus.CompletionTimestamp = checkupStatus.StartTimestamp.Add(time.Minute)
+
 		assert.NoError(t, reporterUnderTest.Report(checkupStatus))
 
 		expectedReportData := map[string]string{
-			reporter.StartTimestampKey: timestamp(checkupStatus.StartTimestamp),
+			reporter.StartTimestampKey:      timestamp(checkupStatus.StartTimestamp),
+			reporter.CompletionTimestampKey: timestamp(checkupStatus.CompletionTimestamp),
 		}
 
 		assert.Equal(t,
--- a/kiagnose/internal/reporter/reporter_test.go
+++ b/kiagnose/internal/reporter/reporter_test.go
@@ -45,8 +45,9 @@ const (
 
 func TestReportShouldSucceed(t *testing.T) {
 	type successTestCase struct {
-		description string
-		succeeded   bool
+		description   string
+		succeeded     bool
+		failureReason string
 	}
 
 	var (
@@ -79,12 +80,14 @@ func TestReportShouldSucceed(t *testing.T) {
 
 	testCases := []successTestCase{
 		{
-			description: "on checkup successful completion",
-			succeeded:   true,
+			description:   "on checkup successful completion",
+			succeeded:     true,
+			failureReason: "",
 		},
 		{
-			description: "on checkup failed completion",
-			succeeded:   false,
+			description:   "on checkup failed completion",
+			succeeded:     false,
+			failureReason: "some reason",
 		},
 	}
 
@@ -95,6 +98,7 @@ func TestReportShouldSucceed(t *testing.T) {
 			assert.NoError(t, reporterUnderTest.Report(checkupStatus))
 
 			checkupStatus.Succeeded = testCase.succeeded
+			checkupStatus.FailureReason = testCase.failureReason
 			checkupStatus.CompletionTimestamp = checkupStatus.StartTimestamp.Add(time.Minute)
 
 			assert.NoError(t, reporterUnderTest.Report(checkupStatus))
@@ -102,6 +106,7 @@ func TestReportShouldSucceed(t *testing.T) {
 			expectedReportData := map[string]string{
 				reporter.StartTimestampKey:      timestamp(checkupStatus.StartTimestamp),
 				reporter.SucceededKey:           strconv.FormatBool(checkupStatus.Succeeded),
+				reporter.FailureReasonKey:       checkupStatus.FailureReason,
 				reporter.CompletionTimestampKey: timestamp(checkupStatus.CompletionTimestamp),
 			}
 
--- a/kiagnose/internal/reporter/reporter_test.go
+++ b/kiagnose/internal/reporter/reporter_test.go
@@ -21,6 +21,7 @@ package reporter_test
 
 import (
 	"context"
+	"strconv"
 	"testing"
 	"time"
 
@@ -43,6 +44,11 @@ const (
 )
 
 func TestReportShouldSucceed(t *testing.T) {
+	type successTestCase struct {
+		description string
+		succeeded   bool
+	}
+
 	var (
 		fakeClient        *fake.Clientset
 		reporterUnderTest *reporter.Reporter
@@ -71,25 +77,40 @@ func TestReportShouldSucceed(t *testing.T) {
 		)
 	})
 
-	t.Run("on final report", func(t *testing.T) {
-		setup()
+	testCases := []successTestCase{
+		{
+			description: "on checkup successful completion",
+			succeeded:   true,
+		},
+		{
+			description: "on checkup failed completion",
+			succeeded:   false,
+		},
+	}
 
-		assert.NoError(t, reporterUnderTest.Report(checkupStatus))
+	for _, testCase := range testCases {
+		t.Run(testCase.description, func(t *testing.T) {
+			setup()
 
-		checkupStatus.CompletionTimestamp = checkupStatus.StartTimestamp.Add(time.Minute)
+			assert.NoError(t, reporterUnderTest.Report(checkupStatus))
 
-		assert.NoError(t, reporterUnderTest.Report(checkupStatus))
+			checkupStatus.Succeeded = testCase.succeeded
+			checkupStatus.CompletionTimestamp = checkupStatus.StartTimestamp.Add(time.Minute)
 
-		expectedReportData := map[string]string{
-			reporter.StartTimestampKey:      timestamp(checkupStatus.StartTimestamp),
-			reporter.CompletionTimestampKey: timestamp(checkupStatus.CompletionTimestamp),
-		}
+			assert.NoError(t, reporterUnderTest.Report(checkupStatus))
 
-		assert.Equal(t,
-			mergeMaps(checkupSpecData(), expectedReportData),
-			getCheckupData(t, fakeClient, configMapNamespace, configMapName),
-		)
-	})
+			expectedReportData := map[string]string{
+				reporter.StartTimestampKey:      timestamp(checkupStatus.StartTimestamp),
+				reporter.SucceededKey:           strconv.FormatBool(checkupStatus.Succeeded),
+				reporter.CompletionTimestampKey: timestamp(checkupStatus.CompletionTimestamp),
+			}
+
+			assert.Equal(t,
+				mergeMaps(checkupSpecData(), expectedReportData),
+				getCheckupData(t, fakeClient, configMapNamespace, configMapName),
+			)
+		})
+	}
 }
 
 func TestReportShouldFail(t *testing.T) {
--- a/kiagnose/internal/reporter/reporter_test.go
+++ b/kiagnose/internal/reporter/reporter_test.go
@@ -1,0 +1,170 @@
+/*
+ * This file is part of the kiagnose project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright 2022 Red Hat, Inc.
+ *
+ */
+
+package reporter_test
+
+import (
+	"context"
+	"testing"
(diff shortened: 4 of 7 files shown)


orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/35b495cb153df6d4012c0acabd1477720958fa13..ab4b72460310f33290fadf7372b6f38090785e1d): Moved handling and testing of failure cases to first commit.

ormergi (commented):
Hi Orel,

It looks better then first implementation!
I couldn't find the part where `succeeded` and `failureReason` are being set at Launcher.Run(), did I miss something?

It seems that `startTimeStamp`, `completionTimeStamp`, `succeed` and `failureReason` describing the checkup execution, I think it will make more sense if `Launcher{}` would have the `status` instead of `Reporter{}`, its closer to reality which makes it it easier to understated IMO.
It will make the reporter very simple and easier to replace (switching to CRD)
and its seems it will be easier to integrate with the checkup resutls ConfigMap data: 
```
type reporter interface {
	Report(map[string]string) error
}

type status struct {
        succeeded bool
        failureReason
        startTs time.Time
        completionTs time.Time
        checkupResults map[string]string
}

type Launcher struct {
    ...
    status status
}

func (l Launcher) Run() (runErr error) {
	defer func() {
		**l.status.SetCompletionTimestamp(time.Now())**
                **l.status.SetResults(checkup.Results())**
		if reportErr := l.reporter.Report(**l.status.StringsMap()**); reportErr != nil {
        ...
        }
   	**l.status.SetStartTimestamp(time.Now())**
        ...
        l.status.SetSucceeded()
        return nil
 ```

File: kiagnose/internal/reporter/reporter.go [hunk 1]
```diff
@@ -19,13 +19,69 @@
...
+	configMap      *corev1.ConfigMap
+	startTimestamp time.Time
 }
 
-func New() Reporter {
-	return Reporter{}
+func New(client kubernetes.Interface, configMapNamespace, configMapName string) *Reporter {
+	return &Reporter{
+		client: client,
+		configMap: &corev1.ConfigMap{
```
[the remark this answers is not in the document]
orelmisan: In the first time we try to report, the user-supplied ConfigMap object is fetched, so relevant fields are filled.

kiagnose/internal/reporter/reporter_test.go
```diff
@@ -0,0 +1,121 @@
...
+	t.Run("when user-supplied ConfigMap doesn't exists", func(t *testing.T) {
+		fakeClient := fake.NewSimpleClientset()
+		reporterUnderTest := reporter.New(fakeClient, configMapNamespace, configMapName)
+		assert.ErrorContains(t, reporterUnderTest.Report(), "not found")
+	})
+
+	t.Run("when user-supplied ConfigMap is deleted after first fetch", func(t *testing.T) {
+		fakeClient := fake.NewSimpleClientset(newConfigMap(checkupRawSpec))
+		reporterUnderTest := reporter.New(fakeClient, configMapNamespace, configMapName)
+		reporterUnderTest.SetStartTimestamp(time.Now())
```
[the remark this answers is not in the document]
orelmisan: We can remove it, it doesn't change the test.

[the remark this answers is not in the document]
orelmisan: Done.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/ab4b72460310f33290fadf7372b6f38090785e1d..ba9e8f15fa0ea811ddfb4daef8360fc96c4d2683): Answered code review comment.

kiagnose/internal/reporter/reporter.go
```diff
@@ -19,13 +19,69 @@
...
+		if err != nil {
+			return err
+		}
+
+		r.configMap = configMap
+	}
+
+	if r.configMap.Data == nil {
+		return ErrConfigMapDataIsNil
+	}
```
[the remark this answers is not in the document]
orelmisan: Imaging this scenario:
1. The user creates a ConfigMap with all required checkup configuration.
1. The user starts the framework with a reference to the ConfigMap from the previous step.

orelmisan:
> I couldn't find the part where succeeded and failureReason are being set at Launcher.Run(), did I miss something?

As the PR's description states: 
1. `succeeded` field is always reported as false.
2. `failureReason` field is always reported as Failure: implement me.

`succeeded`, `failureReason` and `results` will be integrated in later PR(s).

orelmisan:
> It seems that startTimeStamp, completionTimeStamp, succeed and failureReason describing the checkup execution, I think it will make more sense if Launcher{} would have the status instead of Reporter{}, its closer to reality which makes it it easier to understated IMO.
It will make the reporter very simple and easier to replace (switching to CRD)
and its seems it will be easier to integrate with the checkup resutls ConfigMap data:

Please note the the `status` type introduced by this PR is an interface - there is no integration of it with the Launcher.

ormergi:
> As the PR's description states:
> 
> 1. `succeeded` field is always reported as "false".
> 2. `failureReason` field is always reported as "Failure: implement me".
> 
> `succeeded`, `failureReason` and `results` will be integrated in later PR(s).

@orelmisan It will make more sense if you introduce  `succeeded`, `failureReason` on the same PR with the integration changes.

orelmisan:
@ormergi Please see @EdDev's suggestion [here](https://github.com/kiagnose/kiagnose/pull/35#issuecomment-1110950878).
This PR covers the second bullet.

kiagnose/internal/reporter/reporter_test.go
```diff
@@ -0,0 +1,120 @@
...
+	})
+
+	t.Run("when user-supplied ConfigMap is deleted after first fetch", func(t *testing.T) {
+		fakeClient := fake.NewSimpleClientset(newConfigMap(checkupRawSpec))
+		reporterUnderTest := reporter.New(fakeClient, configMapNamespace, configMapName)
+		assert.NoError(t, reporterUnderTest.Report())
+
+		assert.NoError(t, fakeClient.CoreV1().ConfigMaps(configMapNamespace).Delete(context.Background(), configMapName, metav1.DeleteOptions{}))
+		assert.ErrorContains(t, reporterUnderTest.Report(), "not found")
+	})
```
EdDev, line 137: What about when the `Update` fails?

kiagnose/internal/reporter/reporter_test.go
```diff
@@ -0,0 +1,120 @@
...
+	configMapNamespace = "kiagnose"
+	configMapName      = "checkup1"
+
+	testImageValue   = "mycheckup:v0.1.0"
+	testTimeoutValue = "1m"
+
+	testStartTimestamp = 1650882937
+)
+
+var checkupRawSpec = map[string]string{config.ImageKey: testImageValue, config.TimeoutKey: testTimeoutValue}
```
EdDev, line 51: It is not clear why the data is useful in the tests below. I have not seen them in use.

EdDev (changes requested):
First & second commits:

I think the two belong to the same commit because the second show how the first is used (and why it should be like it is).

I have not checked yet the rest, but maybe reporting and setting should always come toghether due to the same reasoning.

I would also check if you can cover the integration between all the components at the `launcher`. Current tests use a `stub, but I think you can use the real thing now and even add the expected assertions.
Generally speaking, I would prefer to see everything exercised through the `launcher` and only scenarios which cannot, see directly on the reporter.

kiagnose/internal/reporter/reporter.go
```diff
@@ -78,6 +82,14 @@ func (r *Reporter) Report() error {
 		r.configMap.Data[CompletionTimestampKey] = r.completionTimestamp.Format(time.RFC3339)
 	}
 
+	if r.status != nil {
+		if succeeded := r.status.Succeeded(); succeeded != "" {
```
EdDev, line 87: I would expect it to be a boolean. Why this is not the case?

EdDev (commented):
The last two commits gives no hint about what are the concretes you intend to pass in.

There are several options, one is to pass an object (that implements the methods), another is to pass a DTO (struct, with no behavior) that was composed in the `launcher`.
I am unclear which is better, but perhaps you can raise reason why one is better than the other.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/ba9e8f15fa0ea811ddfb4daef8360fc96c4d2683..854d2d8b80ee6d8a47731734d0abea0891a4dce9): Rebase
Answered review comments.

kiagnose/internal/reporter/reporter.go
orelmisan: It is now a boolean.
This field is reported only if the checkup is completed (completion timestamp is non zero).

kiagnose/internal/reporter/reporter_test.go
orelmisan: This data is simulating the checkup spec provided by the user.
It is used to check that the reporter appends the status data to it, without modifying the spec.

kiagnose/internal/reporter/reporter_test.go
orelmisan: It is checked in the test where the ConfigMap is deleted after the first fetch:
1. The ConfigMap is fetched and stored.
2. There is a successful update.
3. The test deletes the ConfigMap without the reporter knowing about it.
4. The reporter tries to update the ConfigMap another time but fails because it doesn't exists.

kiagnose/internal/reporter/reporter_test.go
```diff
@@ -0,0 +1,120 @@
...
+			reporter.StartTimestampKey: startTimestamp.Format(time.RFC3339),
+		}
+
+		fakeClient := fake.NewSimpleClientset(newConfigMap(checkupRawSpec))
+		reporterUnderTest := reporter.New(fakeClient, configMapNamespace, configMapName)
+
+		reporterUnderTest.SetStartTimestamp(startTimestamp)
+		assert.NoError(t, reporterUnderTest.Report())
+		assertReport(t, fakeClient, checkupRawSpec, expectedReport)
+	})
```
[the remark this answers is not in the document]
orelmisan: This was added in the first commit.
From the second commit onward - it is checked twice:
1. After the first report (before Setup()).
2. After the second report (After Teardown() or a failure).

kiagnose/internal/reporter/reporter_test.go
```diff
@@ -0,0 +1,120 @@
...
+		assert.ErrorIs(t, reporterUnderTest.Report(), reporter.ErrConfigMapDataIsNil)
+	})
+
+	t.Run("when user-supplied ConfigMap doesn't exists", func(t *testing.T) {
+		fakeClient := fake.NewSimpleClientset()
+		reporterUnderTest := reporter.New(fakeClient, configMapNamespace, configMapName)
+		assert.ErrorContains(t, reporterUnderTest.Report(), "not found")
+	})
+
+	t.Run("when user-supplied ConfigMap is deleted after first fetch", func(t *testing.T) {
```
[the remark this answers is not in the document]
orelmisan: There are two different code paths, checked by two tests:
1. When the initial ConfigMap fetch fails.
2. When an update to the stored ConfigMap fails.

kiagnose/internal/reporter/reporter_test.go
```diff
@@ -0,0 +1,120 @@
...
+		startTimestamp := time.Unix(testStartTimestamp, 0)
+		expectedReport := map[string]string{
+			reporter.StartTimestampKey: startTimestamp.Format(time.RFC3339),
+		}
+
+		fakeClient := fake.NewSimpleClientset(newConfigMap(checkupRawSpec))
+		reporterUnderTest := reporter.New(fakeClient, configMapNamespace, configMapName)
+
+		reporterUnderTest.SetStartTimestamp(startTimestamp)
+		assert.NoError(t, reporterUnderTest.Report())
```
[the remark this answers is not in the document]
orelmisan: Done.

kiagnose/internal/reporter/reporter_test.go
```diff
@@ -0,0 +1,120 @@
...
+	testStartTimestamp = 1650882937
+)
+
+var checkupRawSpec = map[string]string{config.ImageKey: testImageValue, config.TimeoutKey: testTimeoutValue}
+
+func TestReportShouldSucceed(t *testing.T) {
+	t.Run("when reporting startTimestamp", func(t *testing.T) {
+		startTimestamp := time.Unix(testStartTimestamp, 0)
+		expectedReport := map[string]string{
+			reporter.StartTimestampKey: startTimestamp.Format(time.RFC3339),
```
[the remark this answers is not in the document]
orelmisan: Done.

kiagnose/internal/reporter/reporter_test.go
```diff
@@ -0,0 +1,120 @@
...
+	testImageValue   = "mycheckup:v0.1.0"
+	testTimeoutValue = "1m"
+
+	testStartTimestamp = 1650882937
+)
+
+var checkupRawSpec = map[string]string{config.ImageKey: testImageValue, config.TimeoutKey: testTimeoutValue}
+
+func TestReportShouldSucceed(t *testing.T) {
+	t.Run("when reporting startTimestamp", func(t *testing.T) {
```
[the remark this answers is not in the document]
orelmisan: Done.

kiagnose/internal/reporter/reporter_test.go
```diff
@@ -0,0 +1,120 @@
...
+ */
+
+package reporter_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
```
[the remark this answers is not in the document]
orelmisan: Done.

kiagnose/internal/reporter/reporter.go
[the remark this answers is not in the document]
orelmisan: @EdDev please see [your comment](https://github.com/kiagnose/kiagnose/pull/35/files#r858751693) from PR #35.

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -122,9 +128,20 @@ func (s checkupStub) Teardown() error {
 }
 
 type reporterStub struct {
-	failReport error
+	failReport      error
+	failOnReportNum int
```
EdDev, line 132: This seems to assume the inner behavior or the reporting action.
It makes the test fragile as it is strongly bound to the implementation details.

Maybe a software dependency on the behavior will be this: We expect the report to be called twice, once at pre-setup and the second time on teardown.
If we define it like that, you can have two flags: One for the `failReportAtStart` and the second for `failReportAtTeardown`. Seeing that at the test will make things clearer and how you do it it is up to the stub logic.

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -26,68 +26,74 @@ import (
...
 func TestLauncherShould(t *testing.T) {
 	t.Run("run successfully", func(t *testing.T) {
-		testLauncher := launcher.New(checkupStub{}, reporterStub{})
+		testLauncher := launcher.New(checkupStub{}, newReporterStub(nil, 0))
 		assert.NoError(t, testLauncher.Run())
 	})
 
-	t.Run("fail when report is failing", func(t *testing.T) {
-		testLauncher := launcher.New(checkupStub{}, reporterStub{failReport: errorReport})
+	t.Run("fail when first report is failing", func(t *testing.T) {
```
EdDev, line 38: Please do not use `first` (or `second`) as mentioned in another comment.

EdDev (changes requested):

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/854d2d8b80ee6d8a47731734d0abea0891a4dce9..4bea2f0e7b54a8d51698b41dee0d3544aabb978f): Answered review comments.

kiagnose/internal/launcher/launcher_test.go
orelmisan: Changed "first" to "pre Setup()".
Changed "final" to "post Teardown()".

kiagnose/internal/launcher/launcher_test.go
orelmisan: The `reporterStub` has tree capabilities:
1. Never fail.
2. Always fail (to cover failure on the pre Setup() report).
3. Fail after teardown (fail on the second reporting).

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -122,9 +128,20 @@ func (s checkupStub) Teardown() error {
...
 type reporterStub struct {
-	failReport error
+	failReport      error
+	failOnReportNum int
+	reportCount     int
 }
 
-func (r reporterStub) Report() error {
-	return r.failReport
+func newReporterStub(failReport error, failOnReportNum int) *reporterStub {
```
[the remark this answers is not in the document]
orelmisan: Done.

kiagnose/internal/launcher/launcher.go
```diff
@@ -46,8 +49,10 @@ func New(checkup workload, reporter reporter) Launcher {
 }
 
 func (l Launcher) Run() (runErr error) {
+	currentStatus := status.Status{StartTimestamp: time.Now()}
```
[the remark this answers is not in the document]
orelmisan: Changed `currentStatus` to `statusData`.

kiagnose/internal/launcher/launcher.go
```diff
@@ -30,7 +33,7 @@ type workload interface {
 }
 
 type reporter interface {
-	Report() error
+	Report(*status.Status) error
```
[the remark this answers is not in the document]
orelmisan: Changed to regular struct.

kiagnose/internal/reporter/reporter_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+	t.Run("when user-supplied ConfigMap is deleted after first fetch", func(t *testing.T) {
+		fakeClient := fake.NewSimpleClientset(newConfigMap(checkupRawSpec))
+		reporterUnderTest := reporter.New(fakeClient, configMapNamespace, configMapName)
+
+		statusToReport := status.Status{}
+
+		assert.NoError(t, reporterUnderTest.Report(&statusToReport))
+
+		assert.NoError(t, fakeClient.CoreV1().ConfigMaps(configMapNamespace).Delete(context.Background(), configMapName, metav1.DeleteOptions{}))
```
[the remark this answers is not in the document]
orelmisan: Added a function `injectFailureToAccessCheckupData()`.

kiagnose/internal/reporter/reporter_test.go
```diff
@@ -0,0 +1,134 @@
...
+func TestReportShouldSucceed(t *testing.T) {
+	t.Run("when checkup succeeds", func(t *testing.T) {
+		fakeClient := fake.NewSimpleClientset(newConfigMap(checkupRawSpec))
+		reporterUnderTest := reporter.New(fakeClient, configMapNamespace, configMapName)
+
+		statusToReport := status.Status{StartTimestamp: time.Now()}
+
+		expectedReport := map[string]string{reporter.StartTimestampKey: timestamp(statusToReport.StartTimestamp)}
+
+		for i := 0; i < 2; i++ {
```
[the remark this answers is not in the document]
orelmisan: Done.

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -26,69 +26,70 @@ import (
...
 func TestLauncherShould(t *testing.T) {
 	t.Run("run successfully", func(t *testing.T) {
-		testLauncher := launcher.New(checkupStub{}, reporterStub{})
+		testLauncher := launcher.New(checkupStub{}, &reporterStub{})
 		assert.NoError(t, testLauncher.Run())
 	})
 
-	t.Run("fail when report is failing", func(t *testing.T) {
-		testLauncher := launcher.New(checkupStub{}, reporterStub{failReport: errorReport})
+	t.Run("fail when pre Setup() report is failing", func(t *testing.T) {
```
EdDev, line 38: Please try to avoid using function/method names.
We want to test scenarios, not individual functions.

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -26,69 +26,70 @@ import (
...
+		testLauncher := launcher.New(checkupStub{}, &reporterStub{reportErr: errorReport})
+
+		assert.ErrorContains(t, testLauncher.Run(), errorReport.Error())
+	})
+
+	t.Run("fail when post Teardown() report is failing", func(t *testing.T) {
```
EdDev, line 44: Same

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -26,69 +26,70 @@ import (
...
+		testLauncher := launcher.New(checkupStub{failSetup: errorSetup}, &reporterStub{})
 		assert.ErrorContains(t, testLauncher.Run(), errorSetup.Error())
 	})
 
-	t.Run("fail when setup and report are failing", func(t *testing.T) {
-		testLauncher := launcher.New(
-			checkupStub{failSetup: errorSetup},
-			reporterStub{failReport: errorReport},
-		)
+	t.Run("fail when setup and post Teardown() report are failing", func(t *testing.T) {
```
EdDev, line 54: same

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -26,69 +26,70 @@ import (
...
+		testLauncher := launcher.New(checkupStub{failTeardown: errorTeardown}, &reporterStub{})
 		assert.ErrorContains(t, testLauncher.Run(), errorTeardown.Error())
 	})
 
-	t.Run("fail when run and report are failing", func(t *testing.T) {
-		testLauncher := launcher.New(
-			checkupStub{failRun: errorRun},
-			reporterStub{failReport: errorReport},
-		)
+	t.Run("fail when run and post Teardown() report are failing", func(t *testing.T) {
```
EdDev, line 71: same

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -26,69 +26,70 @@ import (
...
 		err := testLauncher.Run()
 		assert.ErrorContains(t, err, errorRun.Error())
 		assert.ErrorContains(t, err, errorReport.Error())
 	})
 
-	t.Run("fail when teardown and report are failing", func(t *testing.T) {
-		testLauncher := launcher.New(
-			checkupStub{failTeardown: errorTeardown},
-			reporterStub{failReport: errorReport},
+	t.Run("fail when teardown and post Teardown() report are failing", func(t *testing.T) {
```
EdDev, line 78: same

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -26,69 +26,70 @@ import (
...
+		testLauncher := launcher.New(checkupStub{failTeardown: errorTeardown},
+			&reporterStub{reportErr: errorReport, failReportAfterTeardown: true},
```
EdDev, line 80: I wish we had a formatter that does it nicely.
I do not understand why gofmt is not consistent in this :/

It is hard to read if there is no consistency in either all the args are on the same line or not.
`foo(a, b, c)`
OR
```
foo(
  a,
  b,
  c,
)
```

This is even more important when the args themselves have args...

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -26,69 +26,70 @@ import (
...
 		err := testLauncher.Run()
 		assert.ErrorContains(t, err, errorTeardown.Error())
 		assert.ErrorContains(t, err, errorReport.Error())
 	})
 
-	t.Run("fail when run, teardown and report are failing", func(t *testing.T) {
-		testLauncher := launcher.New(
-			checkupStub{failRun: errorRun, failTeardown: errorTeardown},
-			reporterStub{failReport: errorReport},
+	t.Run("fail when run, teardown and post Teardown() report are failing", func(t *testing.T) {
```
EdDev, line 88: same

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -122,9 +123,21 @@ func (s checkupStub) Teardown() error {
...
-func (r reporterStub) Report() error {
-	return r.failReport
+func (r *reporterStub) Report(_ status.Status) error {
+	r.reportCount++
+
+	if r.reportErr != nil && !r.failReportAfterTeardown {
+		return r.reportErr
+	}
+
+	if r.failReportAfterTeardown && r.reportCount == 2 {
```
EdDev, line 138: Maybe you should panic if it is more than 2? You now assume this is what will happen, so if it changes, you better know about it.

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -122,9 +123,21 @@ func (s checkupStub) Teardown() error {
...
+	failReportAfterTeardown bool
+	reportCount             int
 }
 
-func (r reporterStub) Report() error {
-	return r.failReport
+func (r *reporterStub) Report(_ status.Status) error {
+	r.reportCount++
+
+	if r.reportErr != nil && !r.failReportAfterTeardown {
```
EdDev, line 134: And another option could have been to input two errors, one for the pre and the other for the post/after.
It would have looked nicer IMO with simpler conditions.

EdDev (commented):
First pass.

kiagnose/internal/reporter/reporter_test.go
orelmisan: Done.

kiagnose/internal/reporter/reporter_test.go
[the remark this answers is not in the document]
orelmisan: Done.

EdDev (commented):
Finished reviewing the 1st commit.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/4bea2f0e7b54a8d51698b41dee0d3544aabb978f..a50fc36a7c97eda5e7c326da5f516101a20f6651): Answered review comments.

kiagnose/internal/launcher/launcher.go
```diff
@@ -52,6 +52,7 @@ func (l Launcher) Run() (runErr error) {
 	statusData := status.Status{StartTimestamp: time.Now()}
 
 	defer func() {
+		statusData.Succeeded = false
```
EdDev, line 55: I do not really mind much, but I think it is redundant.
By default the boolean is false and no where else you set it to true.

EdDev (commented):
Finished reviewing the rest.

kiagnose/internal/reporter/reporter_test.go
```diff
@@ -0,0 +1,151 @@
...
+	configMapNamespace = "kiagnose"
+	configMapName      = "checkup1"
+)
+
+func TestReportShouldSucceed(t *testing.T) {
+	t.Run("when checkup succeeds", func(t *testing.T) {
+		fakeClient := fake.NewSimpleClientset(newConfigMap(checkupSpecData()))
+		reporterUnderTest := reporter.New(fakeClient, configMapNamespace, configMapName)
+
+		// Pre Setup() report
```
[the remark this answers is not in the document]
orelmisan: Done.

kiagnose/internal/launcher/launcher.go
orelmisan: IMO it should be explicit to emphasize its not integrated to the checkup.

kiagnose/internal/launcher/launcher_test.go
orelmisan: Done.

kiagnose/internal/launcher/launcher_test.go
orelmisan: Done.

kiagnose/internal/launcher/launcher_test.go
orelmisan: All calls to launcher.New() are now consistent.

kiagnose/internal/launcher/launcher_test.go
orelmisan: Done.

kiagnose/internal/launcher/launcher_test.go
orelmisan: Done.

kiagnose/internal/launcher/launcher_test.go
orelmisan: Done.

kiagnose/internal/launcher/launcher_test.go
orelmisan: Done.

kiagnose/internal/launcher/launcher_test.go
orelmisan: Done.

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -122,9 +156,26 @@ func (s checkupStub) Teardown() error {
 }
 
 type reporterStub struct {
-	failReport error
+	reportErr           error
+	failAtInitialReport bool
+	failAtFinalReport   bool
```
EdDev, line 161: From the test perspective, I find this messy.
Can you just pass in the error you expect to get and let the `Report` stub calculate it based on the specific error?

If one passes the "start report" error, it will know to raise it one the first attempt.
If one passes the "completion report" error, it will know to raise it only at the second attempt.

This way, you do not need 3 members, but only one.

kiagnose/internal/launcher/launcher_test.go
EdDev: I continued on a new [thread](https://github.com/kiagnose/kiagnose/pull/45/commits/393e0778847f8573082216528df011622d4cb016#r867445479).
I did not originally meant to pass booleans, that is always a smell.

kiagnose/internal/launcher/launcher.go
EdDev: Ok

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/012a0a1cbc7d4394866a103170dce0bfe0b211f4..d921a5a3801bce037fb14614b6c0b65cab493dec): Removed redundant members of struct `checkupStub`.

kiagnose/internal/launcher/launcher_test.go
orelmisan: Done.

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -122,9 +157,24 @@ func (s checkupStub) Teardown() error {
...
+	}
+
+	if r.reportErr == nil {
+		return nil
+	}
+
+	if r.reportCount == 1 && r.reportErr == errorFailOnInitialReport ||
+		r.reportCount == 2 && r.reportErr == errorFailOnFinalReport {
+		return r.reportErr
+	}
```
EdDev, line 177: Can this be simplified to: `return r.reportErr` ?
Maybe you do not even need to counter now.

If it returns the wrong error, you will catch it.

orelmisan: I think we need to return an error only when it is required.
For example, is the error should be on the second report, IMO we should return `nil`, and not an error.

EdDev: Right.

EdDev: Hmm.. but then, why do you need the `	if r.reportErr == nil` condition? Isn't it redundant?

orelmisan: It was a left over and was removed.
Thanks.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/d921a5a3801bce037fb14614b6c0b65cab493dec..862f725325b0b02295d19bfc825b291bad80d3e4): Answered a review comment.

EdDev (dismissed):
Thank you!

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -26,81 +26,116 @@ import (
...
 var (
-	errorSetup    = errors.New("setup error")
-	errorRun      = errors.New("run error")
-	errorTeardown = errors.New("teardown error")
-	errorReport   = errors.New("report error")
+	errorSetup               = errors.New("setup error")
+	errorRun                 = errors.New("run error")
+	errorTeardown            = errors.New("teardown error")
+	errorFailOnInitialReport = errors.New("initial report error")
+	errorFailOnFinalReport   = errors.New("final report error")
```
ormergi, line 138: nit: Please consider renaming `errorFailOnFinalReport` to something like `errorCompletionReport`
it will be easier to understand the test flow as it will be consistent with the test description.

kiagnose/internal/reporter/reporter.go
```diff
@@ -19,13 +19,64 @@
...
+				Name:      configMapName,
+				Namespace: configMapNamespace,
+			},
+			Data: nil,
+		},
+	}
 }
 
-func (r Reporter) Report() error {
+func (r *Reporter) Report(statusData status.Status) error {
```
ormergi, line 62: Looking at this again, I think there is no need for the Reporter{} to keep an instance of the `ConfigMap`
it can be used locally instead, I think it will simplify this function and also `reporter.New().`

kiagnose/internal/reporter/reporter.go
```diff
@@ -19,13 +19,64 @@
...
+	if r.configMap.Data == nil {
+		configMap, err := configmap.Get(r.client.CoreV1(), r.configMap.Namespace, r.configMap.Name)
+		if err != nil {
+			return err
+		}
+
+		r.configMap = configMap
+	}
```
ormergi, line 70: I think fetching the `ConfigMap` should not be conditional, its better to always fetch the latest state of the `ConfigMap`
it may prevent unexpected behavior.

kiagnose/internal/reporter/reporter_test.go
```diff
@@ -0,0 +1,170 @@
...
+		reporterUnderTest *reporter.Reporter
+		checkupStatus     status.Status
+	)
+
+	setup := func() {
+		fakeClient = fake.NewSimpleClientset(newConfigMap(checkupSpecData()))
+		reporterUnderTest = reporter.New(fakeClient, configMapNamespace, configMapName)
+
+		checkupStatus = status.Status{StartTimestamp: time.Now()}
+	}
```
ormergi, line 64: nit: I think it will improve visibility if those were called directly (instead of setup()).

kiagnose/internal/reporter/reporter_test.go
```diff
@@ -0,0 +1,170 @@
...
+	for k, v := range map2 {
+		resultMap[k] = v
+	}
+
+	return resultMap
+}
+
+func timestamp(t time.Time) string {
+	return t.Format(time.RFC3339)
+}
```
ormergi, line 189: nit: I think it will improve visibility `Fromat` would be called directly (instead of timestamp())
and save me few scrolls :wink: .

ormergi (commented):
Please see my inline comments

kiagnose/internal/reporter/reporter_test.go
```diff
@@ -0,0 +1,170 @@
...
+	)
+
+	setup := func() {
+		fakeClient = fake.NewSimpleClientset(newConfigMap(checkupSpecData()))
+		reporterUnderTest = reporter.New(fakeClient, configMapNamespace, configMapName)
+
+		checkupStatus = status.Status{StartTimestamp: time.Now()}
+	}
+
+	t.Run("on checkup start", func(t *testing.T) {
```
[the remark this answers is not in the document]
orelmisan: There are two reports:
1. When the checkup starts (right before `Setup()`).
2. After checkup completion (successful or failed.)

"when called once" and "when called more then once", don't emphasize to the reader, on which report are we talking about.

kiagnose/internal/reporter/reporter.go
orelmisan: When you keep the ConfigMap in the reporter's state, you also keep its version.
An update will fail if someone tampers with your ConfigMap object behind your back.
IMO this is the desired behavior.

kiagnose/internal/reporter/reporter_test.go
orelmisan: Please see @EdDev's [comment](https://github.com/kiagnose/kiagnose/pull/45#discussion_r862496080).

We can change it in a later PR if it will disturb us.

kiagnose/internal/reporter/reporter_test.go
orelmisan: It will cause these lines to be duplicated.

kiagnose/internal/reporter/reporter.go
orelmisan: The inputs an the outputs of the checkup are strongly connected.
When you are looking at the results, you want to know which inputs has cause them.

By fetching the ConfigMap only once, we ensure that the checkup results / outputs are coming from the checkup inputs defined by the user.

This means that its not possible to change the input during the checkup's execution and prevents unexpected behavior.

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -122,9 +157,20 @@ func (s checkupStub) Teardown() error {
...
+	reportErr   error
+	reportCount int
 }
 
-func (r reporterStub) Report() error {
-	return r.failReport
+func (r *reporterStub) Report(_ status.Status) error {
+	r.reportCount++
+	if r.reportCount > 2 {
+		panic("Report was called more than twice")
```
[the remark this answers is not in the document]
orelmisan: Please see @EdDev's [comment](https://github.com/kiagnose/kiagnose/pull/45#discussion_r864696560).

It is done to break the test in case the stub's basic assumptions are broken (for example adding a third report).

kiagnose/internal/launcher/launcher.go
```diff
@@ -56,6 +61,10 @@ func (l Launcher) Run() (runErr error) {
 		}
 	}()
 
+	if err := l.reporter.Report(statusData); err != nil {
+		return err
+	}
+
```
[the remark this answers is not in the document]
orelmisan: You are right.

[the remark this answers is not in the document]
orelmisan: Done.

kiagnose/internal/launcher/launcher_test.go
orelmisan: I don't think it grammatically correct.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/862f725325b0b02295d19bfc825b291bad80d3e4..6fe3f9bbcd129c429e978bea3f417f2e9207fc06): Moved the initial report before the deferred function.

kiagnose/internal/reporter/reporter.go
EdDev: > An update will fail if someone tampers with your ConfigMap object behind your back.
> IMO this is the desired behavior.

That is actually a bad side effect.
Someone may touch the object and there is nothing you can do about it. It is expected.
As a simple example, some other controller may label or annotate the configMap.

The only way you are suppose to protect against touching the data is through either a controller or an admission control.
I do not think we need to handle it at the moment, we could be optimistic for now.

I see 3 alternatives:
- Keep it as is and do not support integration with other entities in the k8s client.
- Do not keep the state, and as @ormergi suggested, just fetch it each time and depend on it as the fact.
- Keep the state and at the same time fetch the persisted version, compare the relevant data to see nothing changes, and proceed.

kiagnose/internal/reporter/reporter.go
EdDev: Commented on the other thread.

kiagnose/internal/reporter/reporter_test.go
EdDev: @ormergi , What data are you missing in the tests that makes it hard to understand it?

kiagnose/internal/reporter/reporter_test.go
EdDev: When I looked at it, I did not understand what `t.Format(time.RFC3339)` is, and that `RFC3339` tells me nothing useful. It actually makes me go and look for it in the documentation.
But it is a nit, usually I would agree that there is no need to wrap simple stuff.

This is up to the author discretion.

kiagnose/internal/reporter/reporter.go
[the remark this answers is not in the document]
orelmisan: > The user ConfigMap `Spec.Data` either empty or nil is semantically the same from the framework perspective, its possible to handle such error (by simply initialize `spec.data`) and there is no need to fail the checkup execution due to that.
> 
> As I mentioned earlier, it will also provide a better UX for the user if it wont fail on such things. Consider the case when a user runs the same checkup twice, between each execution the user will cleanup the config-map data by simply deleting it from the manifest, the framework will then fail due to `Spec.Data` is `nil`.

I see value in keeping the checkup inputs with the checkup outputs.
If the code reaches to the point were the ConfigMap.Data field is nil or empty, it means that someone had deleted the inputs while the checkup is running.
If we go with the approach to just initialize the Data field, in the end there will be only results, with no inputs.

kiagnose/internal/reporter/reporter.go
orelmisan: > * Keep the state and at the same time fetch the persisted version, compare the relevant data to see nothing changes, and proceed.

I favor this approach. We can implement it in a follow-up PR.

EdDev (approved):

ormergi (commented):
There are 3 different threads discussing similar topic, from my perspective all related to each other : [[1]](https://github.com/kiagnose/kiagnose/pull/45/commits/923d48d847248ec10e73bec9b3c6b906f4119aaa#r867492551) [[2]](https://github.com/kiagnose/kiagnose/pull/45/commits/923d48d847248ec10e73bec9b3c6b906f4119aaa#r867492847) [[3]](https://github.com/kiagnose/kiagnose/pull/45/commits/923d48d847248ec10e73bec9b3c6b906f4119aaa#r862428394)

From what I understand, the current implementation tries to ensure the user `ConfigMap` state is kept in order to ensure that the checkup spec (aka "inputs", anything that is not prefix with "status.") is persisted and wont change or deleted from the user ConfigMap.
Is it correct?

If so, please make sure to mention it the commit message because its not an easy thing to understand for the diff.

Also, I think that the current implementation wont achieve that because the `ConfigMap` can still change at any point of time (which is fine and expected).
Practicality this can be achieved with a controller that will reconcile the `ConfigMap` state 
on change events.

Thus I think the reporter should be simplified and care only for updating the `ConfigMap` data with the new status, and not also ensure the spec is as it was initially (aka "inputs").
Suggestions to simplify the reporter:
- Change reporter to not keep the state of the user `ConfigMap`.
- Always fetch `ConfigMap` at stat of `reporter.Report()`
- At `reporter.Report()` If the user `ConfigMap` `spec.data` is nil do not return immediately.
   Instead do best effort to report what possible and then return meaningful error.
   Checkup execution might be expensive for the user thus the framework should do best effort
   to report the checkup results.

There are planes to use a controller and CRD as replacement for the `ConfigMap`,
persisting the checkup spec (aka "inputs")  can be addressed then.

kiagnose/internal/reporter/reporter.go
```diff
@@ -19,13 +19,64 @@
...
-	return Reporter{}
+func New(client kubernetes.Interface, configMapNamespace, configMapName string) *Reporter {
+	return &Reporter{
+		client: client,
+		configMap: &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      configMapName,
+				Namespace: configMapNamespace,
+			},
+			Data: nil,
```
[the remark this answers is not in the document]
orelmisan: Done.

kiagnose/internal/reporter/reporter_test.go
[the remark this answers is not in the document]
orelmisan: Changed "on checkup start" to "on initial report".

EdDev:
> There are 3 different threads discussing similar topic, from my perspective all related to each other : [[1]](https://github.com/kiagnose/kiagnose/pull/45/commits/923d48d847248ec10e73bec9b3c6b906f4119aaa#r867492551) [[2]](https://github.com/kiagnose/kiagnose/pull/45/commits/923d48d847248ec10e73bec9b3c6b906f4119aaa#r867492847) [[3]](https://github.com/kiagnose/kiagnose/pull/45/commits/923d48d847248ec10e73bec9b3c6b906f4119aaa#r862428394)

If there are 3 threads speaking on the same topic, opening a 4th one will surely not help and unfocus us.
What I feedback before is [here](https://github.com/kiagnose/kiagnose/pull/45#discussion_r867505854).

If you think it needs to be done differently than what was suggested there, we can discuss it in a PR that tries to improve it.

EdDev (approved):
Thank you @orelmisan.

I think most of the comments have been answered and the state of the PR seems good to me.
We can continue improving it and I expect to revisit most of the places again pretty soon.

@ormergi , if I missed some critical aspect which you think is totally wrong, please feel free to propose an alternative when time allows.

ormergi:
> If there are 3 threads speaking on the same topic, opening a 4th one will surely not help and unfocus us. What I feedback before is [here](https://github.com/kiagnose/kiagnose/pull/45#discussion_r867505854).
> 

On the contrary, its confusing to have 3 threads discussing the same thing from different angle.

PR #46, opened 2022-05-08
launcher: Report actual checkup status

orelmisan:
Replace the hard codded report of "succeeded" and "failureReason" with an actual report.

The "succeeded" field is reported as "true" only when:
Checkup.Setup(), Checkup.Run() and Checkup.Teardown() are successfully completed.

At the moment, because the checkup results are not read, beyond the returned error from Checkup.Run(), there is not enough
data to declare the checkup as failed.

This PR also changes the launcher's tests to test the actual final report (in applicable test cases).

~~Depends on PR #45.~~

### DIFF
--- a/kiagnose/internal/launcher/launcher.go
+++ b/kiagnose/internal/launcher/launcher.go
@@ -56,8 +56,12 @@ func (l Launcher) Run() (runErr error) {
 	}
 
 	defer func() {
-		statusData.Succeeded = false
-		statusData.FailureReason = "Failure: implement me"
+		if runErr == nil {
+			statusData.Succeeded = true
+		} else {
+			statusData.FailureReason = runErr.Error()
+		}
+
 		statusData.CompletionTimestamp = time.Now()
 
 		if reportErr := l.reporter.Report(statusData); reportErr != nil {
--- a/kiagnose/internal/launcher/launcher_test.go
+++ b/kiagnose/internal/launcher/launcher_test.go
@@ -21,21 +21,48 @@ package launcher_test
 
 import (
 	"errors"
+	"strconv"
 	"testing"
+	"time"
 
 	assert "github.com/stretchr/testify/require"
 
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/fake"
+
+	"github.com/kiagnose/kiagnose/kiagnose/internal/config"
+	"github.com/kiagnose/kiagnose/kiagnose/internal/configmap"
 	"github.com/kiagnose/kiagnose/kiagnose/internal/launcher"
+	"github.com/kiagnose/kiagnose/kiagnose/internal/reporter"
 	"github.com/kiagnose/kiagnose/kiagnose/internal/status"
 )
 
+const (
+	configMapNamespace = "kiagnose"
+	configMapName      = "checkup1"
+)
+
 func TestLauncherRunsSuccessfully(t *testing.T) {
+	fakeClient := fake.NewSimpleClientset(newConfigMap(checkupSpecData()))
+
 	testLauncher := launcher.New(
 		checkupStub{},
-		&reporterStub{},
+		reporter.New(fakeClient, configMapNamespace, configMapName),
 	)
 
 	assert.NoError(t, testLauncher.Run())
+
+	actualCheckupData := getCheckupData(t, fakeClient, configMapNamespace, configMapName)
+	zeroTimestamps(actualCheckupData)
+
+	expectedData := checkupSpecData()
+	expectedData[reporter.SucceededKey] = strconv.FormatBool(true)
+	expectedData[reporter.FailureReasonKey] = ""
+	zeroTimestamps(expectedData)
+
+	assert.Equal(t, expectedData, actualCheckupData)
 }
 
 func TestLauncherRunShouldFailWithReportWhen(t *testing.T) {
@@ -175,4 +202,35 @@ func (r *reporterStub) Report(_ status.Status) error {
 	}
 
 	return nil
+}
+
+func checkupSpecData() map[string]string {
+	const (
+		testImageValue   = "mycheckup:v0.1.0"
+		testTimeoutValue = "1m"
+	)
+
+	return map[string]string{config.ImageKey: testImageValue, config.TimeoutKey: testTimeoutValue}
+}
+
+func newConfigMap(data map[string]string) *corev1.ConfigMap {
+	return &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      configMapName,
+			Namespace: configMapNamespace,
+		},
+		Data: data,
+	}
+}
+
+func getCheckupData(t *testing.T, client kubernetes.Interface, configMapNamespace, configMapName string) map[string]string {
+	configMap, err := configmap.Get(client.CoreV1(), configMapNamespace, configMapName)
+	assert.NoError(t, err)
+
+	return configMap.Data
+}
+
+func zeroTimestamps(data map[string]string) {
+	data[reporter.StartTimestampKey] = time.Time{}.Format(time.RFC3339)
+	data[reporter.CompletionTimestampKey] = time.Time{}.Format(time.RFC3339)
 }
--- a/kiagnose/internal/launcher/launcher_test.go
+++ b/kiagnose/internal/launcher/launcher_test.go
@@ -38,26 +38,8 @@ func TestLauncherRunsSuccessfully(t *testing.T) {
 	assert.NoError(t, testLauncher.Run())
 }
 
-func TestLauncherShould(t *testing.T) {
-	t.Run("fail when report on checkup start is failing", func(t *testing.T) {
-		testLauncher := launcher.New(
-			checkupStub{},
-			&reporterStub{reportErr: errorFailOnInitialReport},
-		)
-
-		assert.ErrorContains(t, testLauncher.Run(), errorFailOnInitialReport.Error())
-	})
-
-	t.Run("fail when report on checkup completion is failing", func(t *testing.T) {
-		testLauncher := launcher.New(
-			checkupStub{},
-			&reporterStub{reportErr: errorFailOnFinalReport},
-		)
-
-		assert.ErrorContains(t, testLauncher.Run(), errorFailOnFinalReport.Error())
-	})
-
-	t.Run("fail when setup is failing", func(t *testing.T) {
+func TestLauncherRunShouldFailWithReportWhen(t *testing.T) {
+	t.Run("setup is failing", func(t *testing.T) {
 		testLauncher := launcher.New(
 			checkupStub{failSetup: errorSetup},
 			&reporterStub{},
@@ -66,7 +48,45 @@ func TestLauncherShould(t *testing.T) {
 		assert.ErrorContains(t, testLauncher.Run(), errorSetup.Error())
 	})
 
-	t.Run("fail when setup and report on checkup completion are failing", func(t *testing.T) {
+	t.Run("run is failing", func(t *testing.T) {
+		testLauncher := launcher.New(
+			checkupStub{failRun: errorRun},
+			&reporterStub{},
+		)
+
+		assert.ErrorContains(t, testLauncher.Run(), errorRun.Error())
+	})
+
+	t.Run("teardown is failing", func(t *testing.T) {
+		testLauncher := launcher.New(
+			checkupStub{failTeardown: errorTeardown},
+			&reporterStub{},
+		)
+
+		assert.ErrorContains(t, testLauncher.Run(), errorTeardown.Error())
+	})
+}
+
+func TestLauncherRunShouldFailWithoutReportWhen(t *testing.T) {
+	t.Run("report on checkup start is failing", func(t *testing.T) {
+		testLauncher := launcher.New(
+			checkupStub{},
+			&reporterStub{reportErr: errorFailOnInitialReport},
+		)
+
+		assert.ErrorContains(t, testLauncher.Run(), errorFailOnInitialReport.Error())
+	})
+
+	t.Run("report on checkup completion is failing", func(t *testing.T) {
+		testLauncher := launcher.New(
+			checkupStub{},
+			&reporterStub{reportErr: errorFailOnFinalReport},
+		)
+
+		assert.ErrorContains(t, testLauncher.Run(), errorFailOnFinalReport.Error())
+	})
+
+	t.Run("setup and report on checkup completion are failing", func(t *testing.T) {
 		testLauncher := launcher.New(
 			checkupStub{failSetup: errorSetup},
 			&reporterStub{reportErr: errorFailOnFinalReport},
@@ -77,25 +97,7 @@ func TestLauncherShould(t *testing.T) {
 		assert.ErrorContains(t, err, errorFailOnFinalReport.Error())
 	})
 
-	t.Run("fail when run is failing", func(t *testing.T) {
-		testLauncher := launcher.New(
-			checkupStub{failRun: errorRun},
-			&reporterStub{},
-		)
-
-		assert.ErrorContains(t, testLauncher.Run(), errorRun.Error())
-	})
-
-	t.Run("fail when teardown is failing", func(t *testing.T) {
-		testLauncher := launcher.New(
-			checkupStub{failTeardown: errorTeardown},
-			&reporterStub{},
-		)
-
-		assert.ErrorContains(t, testLauncher.Run(), errorTeardown.Error())
-	})
-
-	t.Run("fail when run and report on checkup completion are failing", func(t *testing.T) {
+	t.Run("run and report on checkup completion are failing", func(t *testing.T) {
 		testLauncher := launcher.New(
 			checkupStub{failRun: errorRun},
 			&reporterStub{reportErr: errorFailOnFinalReport},
@@ -106,7 +108,7 @@ func TestLauncherShould(t *testing.T) {
 		assert.ErrorContains(t, err, errorFailOnFinalReport.Error())
 	})
 
-	t.Run("fail when teardown and report on checkup completion are failing", func(t *testing.T) {
+	t.Run("teardown and report on checkup completion are failing", func(t *testing.T) {
 		testLauncher := launcher.New(
 			checkupStub{failTeardown: errorTeardown},
 			&reporterStub{reportErr: errorFailOnFinalReport},
@@ -117,7 +119,7 @@ func TestLauncherShould(t *testing.T) {
 		assert.ErrorContains(t, err, errorFailOnFinalReport.Error())
 	})
 
-	t.Run("fail when run, teardown and report on checkup completion are failing", func(t *testing.T) {
+	t.Run("run, teardown and report on checkup completion are failing", func(t *testing.T) {
 		testLauncher := launcher.New(
 			checkupStub{failRun: errorRun, failTeardown: errorTeardown},
 			&reporterStub{reportErr: errorFailOnFinalReport},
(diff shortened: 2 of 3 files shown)


kiagnose/internal/launcher/launcher.go
```diff
@@ -52,8 +52,12 @@ func (l Launcher) Run() (runErr error) {
 	statusData := status.Status{StartTimestamp: time.Now()}
 
 	defer func() {
-		statusData.Succeeded = false
-		statusData.FailureReason = "Failure: implement me"
+		if runErr == nil {
+			statusData.Succeeded = true
```
EdDev, line 56: I think this needs to be set only after the `Run()` execution.
Here, I would expect to see it set to `false` with the reason in case of an error.

In practice, there is no need to set the success boolean at all. It should be the default in case the value was not raised before.

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -21,34 +21,103 @@ package launcher_test
...
-			&reporterStub{},
+			reporter.New(fakeClient, configMapNamespace, configMapName),
 		)
 
 		assert.NoError(t, testLauncher.Run())
+
+		checkupData := getCheckupData(t, fakeClient, configMapNamespace, configMapName)
+
+		assert.Equal(t, strconv.FormatBool(true), checkupData[reporter.SucceededKey])
+		assert.Empty(t, checkupData[reporter.FailureReasonKey])
```
EdDev, line 60: Can you compare the full data? That would be preferable over multi-asserts.

EdDev (changes requested):
Commit:launcher: Report succeeded and failureReason

I find this commit hard to review.
It includes too many changes, some not related directly with the production code addition.

Can you please perform the refactor of test structure without changing their logic (and what they test) as a pre-step? Then, it should be easy to see what was changed and what was added more easily.

As an example, I am not sure I understand what does it mean "with report" and "without report" in the failure scenarios.

EdDev (commented):
Passed through the "prefix" commits.

General note: You are using in the commit message a lot of `This` wording. I would recommend reducing its usage to the minimum. It does not add information and looks redundant.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/0a605bbaea9cfaa80f5de4cdfed48a0b81330360..6925c373534baef3367bd3412c54c9cba2efdb67): Added intermediate commits to slowly show the change to `launcher_test.go`.

kiagnose/internal/launcher/launcher.go
orelmisan: This code is called after the calls to `Run()` and `Teardown()`.

The success boolean is only set to `true` in case of success, and remains as the default `false` otherwise.

kiagnose/internal/launcher/launcher_test.go
orelmisan: It is a bit problematic because the timestamps are set inside a struct local to the `Launcher.Run()` method.
We can expose them, but I don't think the justification is very strong.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/6925c373534baef3367bd3412c54c9cba2efdb67..06480039d92c2e1a1a0758cc4c0a4d0064b78f64): Changed the commit messages to reduce the usage of the word "this".

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -40,64 +40,66 @@ func TestLauncherRunShouldSucceed(t *testing.T) {
 	})
 }
 
-func TestLauncherShould(t *testing.T) {
-	t.Run("fail when report on checkup start is failing", func(t *testing.T) {
+func TestLauncherRunShouldFailWithReport(t *testing.T) {
```
ormergi, line 64: nit: Please consider renaming to something like `TestLauncherShouldFailWhen` (and remove "when" from tests description), having functioning reporter is the default state.

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -107,12 +107,14 @@ func TestLauncherRunShouldFailWithReport(t *testing.T) {
 
 func TestLauncherRunShouldFailWithoutReport(t *testing.T) {
 	t.Run("when report on checkup start fails", func(t *testing.T) {
+		fakeClient := fake.NewSimpleClientset()
+
 		testLauncher := launcher.New(
 			checkupStub{},
-			&reporterStub{reportErr: errorFailOnInitialReport},
+			reporter.New(fakeClient, configMapNamespace, configMapName),
```
ormergi, line 114: I think it worth to create another test function when the "production" reporter is being used, 
IMO its confusion to see one test using the "production" component and the rest uses the stub component.

ormergi (commented):
Hi Orel, thanks for the PR.
Please see my inline comments.

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -178,3 +199,29 @@ func (r *reporterStub) Report(_ status.Status) error {
...
 }
+
+func checkupSpecData() map[string]string {
+	const (
+		testImageValue   = "mycheckup:v0.1.0"
+		testTimeoutValue = "1m"
+	)
+
+	return map[string]string{config.ImageKey: testImageValue, config.TimeoutKey: testTimeoutValue}
+}
```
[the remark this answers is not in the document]
orelmisan: Please see @EdDev's [comment](https://github.com/kiagnose/kiagnose/pull/45#discussion_r863616213) in PR #45.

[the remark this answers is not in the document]
EdDev: @orelmisan , I suggested a function instead of a variable. In both cases the end result is hiding the details, and Or was talking about that.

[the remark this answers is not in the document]
orelmisan: These details are representing the minimal checkup spec.
Why in your opinion are they important to the comprehension of the test?

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -29,16 +29,18 @@ import (
 	"github.com/kiagnose/kiagnose/kiagnose/internal/status"
 )
 
-func TestLauncherShould(t *testing.T) {
-	t.Run("run successfully", func(t *testing.T) {
+func TestLauncherRunShouldSucceed(t *testing.T) {
+	t.Run("when checkup runs successfully", func(t *testing.T) {
```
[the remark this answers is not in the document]
EdDev: I think the intention is to check that the launcher runs successfully, stubbing (or not) the needed to make it so.

Anyway, lets not over-complicate things... keep it simple.
There is no real need for `t.Run`, there is only one test.

And the test can be called `TestLauncherRunsSuccessfully`.

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -29,16 +29,18 @@ import (
...
 		testLauncher := launcher.New(
 			checkupStub{},
 			&reporterStub{},
 		)
 
 		assert.NoError(t, testLauncher.Run())
 	})
+}
 
+func TestLauncherShould(t *testing.T) {
```
EdDev: You can place in the name all the common wording from each test below.
I think in this case it is: `TestLauncherShouldFailWhen`.

EdDev (commented):
Reviewed commit `launcher, tests: Move success scenario to a dedicated function`.

EdDev (commented):
> launcher, tests: Split failure cases to two functions

launcher, tests: Split failure cases to two groups

> Split the failure cases into two categories:
> 1. Failures that are expected to be reported to the user.
> 2. Failures that are expected to be logged.

You need some overview so one can understand what is this all about.
Something like:
```
When the launcher execution fails, it can result in either the outcome is successfully reported to the user (through the API sever) or no (e.g. unable to access the API server).

To make this distinction clear and allow it to evolve with future scenarios, the two are grouped under different test hierarchies.
```

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/06480039d92c2e1a1a0758cc4c0a4d0064b78f64..56f078e848176071e7856385c1de1d2f043ad5e1): Rebased on PR #45.
Changed the order of the commits, so refactor commits comes first.
Kept the old test names.
Retained the reporterStub's ability to fail on the initial report.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/56f078e848176071e7856385c1de1d2f043ad5e1..917f2b820f92b6e2e523000b259b16e0c204631f): Rebase

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/917f2b820f92b6e2e523000b259b16e0c204631f..e5a5961c01da9c58c144f852223d3bce55dea342): Updated the 6th commit message and the PR description to explain the logic behind the "succeeded" field.

kiagnose/internal/launcher/launcher.go
orelmisan: Updated the commit message and the PR description to explain the logic behind "succeeded" field.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/e5a5961c01da9c58c144f852223d3bce55dea342..5901de3f6d8ab95b87a24849d0febf6737a09bdc): Zero the actual reported timestamps before asserting equality with expected values.

kiagnose/internal/launcher/launcher_test.go
orelmisan: The actual timestamps are now zeroed before asserting the actual report equality with the expected results.

kiagnose/internal/launcher/launcher.go
orelmisan: @ormergi IMO the Status struct should stay a simple data struct.
The logic added to Launcher.Run() is a simple `if` statement.

kiagnose/internal/launcher/launcher_test.go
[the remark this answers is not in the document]
orelmisan: Removed the redundant `t.Run()`.

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -40,64 +40,66 @@ func TestLauncherRunShouldSucceed(t *testing.T) {
...
 			&reporterStub{},
 		)
 
-		assert.ErrorContains(t, testLauncher.Run(), errorSetup.Error())
+		assert.ErrorContains(t, testLauncher.Run(), errorTeardown.Error())
 	})
+}
 
-	t.Run("fail when setup and report on checkup completion are failing", func(t *testing.T) {
+func TestLauncherRunShouldFailWithoutReport(t *testing.T) {
```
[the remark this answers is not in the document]
orelmisan: Renamed this to `TestLauncherRunShouldFailWithoutReportWhen`.

kiagnose/internal/launcher/launcher_test.go
orelmisan: Renamed to `TestLauncherRunShouldFailWithReportWhen`.

kiagnose/internal/launcher/launcher_test.go
orelmisan: The tests division is based on whether there will be a report to the user at the end or not.

kiagnose/internal/launcher/launcher_test.go
orelmisan: Changed this to:
1. `TestLauncherRunShouldFailWithReportWhen`
2. `TestLauncherRunShouldFailWithoutReportWhen`

kiagnose/internal/launcher/launcher.go
EdDev: Ok, you convinced me it is enough for now.

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -21,21 +21,49 @@ package launcher_test
...
 	testLauncher := launcher.New(
 		checkupStub{},
-		&reporterStub{},
+		reporter.New(fakeClient, configMapNamespace, configMapName),
 	)
 
 	assert.NoError(t, testLauncher.Run())
+
+	actualCheckupData := getCheckupData(t, fakeClient, configMapNamespace, configMapName)
+	zeroTimestamps(actualCheckupData)
```
EdDev, line 58: It will be nicer if you call this on `expectedData` as well.

orelmisan: Done.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/5901de3f6d8ab95b87a24849d0febf6737a09bdc..6af425456486da401b1ce5c577bdd47f67be386e): Answered review comment.

EdDev (dismissed):
Thank you!

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -29,16 +29,16 @@ import (
 	"github.com/kiagnose/kiagnose/kiagnose/internal/status"
 )
 
-func TestLauncherShould(t *testing.T) {
-	t.Run("run successfully", func(t *testing.T) {
-		testLauncher := launcher.New(
-			checkupStub{},
-			&reporterStub{},
-		)
+func TestLauncherRunsSuccessfully(t *testing.T) {
```
ormergi, line 47: nit: redundant 's' at Runs.

kiagnose/internal/launcher/launcher.go
ormergi: This defer function already bloated, it will be nicer if we can reuse some noise from it.
Note that what I suggested can also be static functions at launcher/launcher.go.

kiagnose/internal/launcher/launcher_test.go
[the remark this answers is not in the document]
ormergi: It will improve visibility and make it easier to understand how the "expectedData" should look like,
The "expectedData" can be defined locally where needed.

This is nice to have..

kiagnose/internal/launcher/launcher.go
```diff
@@ -56,8 +56,12 @@ func (l Launcher) Run() (runErr error) {
...
+		} else {
+			statusData.FailureReason = runErr.Error()
+		}
+
```
ormergi, line 64: regarding line [67 line](https://github.com/kiagnose/kiagnose/pull/46/commits/0516103cedea0e2314893be7b3733768688a650b#diff-57e7e8036316bd96561318e6d20d9487345e4ac53cb9c3867a324bffeb4d564eR67), it seems that that status "succeeded" is not affected by the reporter, please add a note about it in the commit message.

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -66,32 +66,52 @@ func TestLauncherRunsSuccessfully(t *testing.T) {
...
-			&reporterStub{},
-		)
+			assert.ErrorContains(t, testLauncher.Run(), testCase.expectedError.Error())
 
-		assert.ErrorContains(t, testLauncher.Run(), errorTeardown.Error())
-	})
+			actualCheckupData := getCheckupData(t, fakeClient, configMapNamespace, configMapName)
+			zeroTimestamps(actualCheckupData)
+
+			expectedData := checkupSpecData()
```
ormergi, line 107: nit: It would be easier to understand what is the expected results if it was defined inline:
```
expectedData := map[string]string {
    reporter.SucceededKey: ..
    reporter.FailureReasonKey: ..
    config.StartTimestampKey: ..
    config.CompletionTimestampKey: ..
}
```

ormergi (commented):
Regarding "launcher, tests: Check report on failure cases" commit:
What about successful tests are you going to change them to assert on the 
status that reported?

Regarding "launcher, tests: Change test when initial report fails" commit 
the commit message title is not clean, do you think you can add more info about what is going to change?

kiagnose/internal/launcher/launcher_test.go
orelmisan: This is grammatically correct.

kiagnose/internal/launcher/launcher.go
orelmisan: This is only a temporary step in the way to PR #48.

kiagnose/internal/launcher/launcher_test.go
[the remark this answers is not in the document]
orelmisan: This could be done in a follow-up PR.

kiagnose/internal/launcher/launcher.go
orelmisan: The line you are referring to is the final report. 
If it fails - the user will not get a report.
There is no point in updating a value that will not be reported.

kiagnose/internal/launcher/launcher_test.go
orelmisan: I agree that it would be nicer, but `expectedData` is comprised of two maps:
1. The minimal spec.
2. All the reported data.

ormergi: It seems (2) is not reflected in tests, did I miss something?

orelmisan: Let me rephrase, section 2 is all the expected reported data.
Please see [here](https://github.com/kiagnose/kiagnose/pull/46/files/0516103cedea0e2314893be7b3733768688a650b..4eafc85ed21521162dceae3e2861acfd6acf589c#diff-5bfeeac3b26c7f73c22a1fa3ef5b79ca2f0f2a69eb8d9c27ce1aaf04b40af719R108-R110).

kiagnose/internal/launcher/launcher_test.go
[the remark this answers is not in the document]
EdDev: To be clear: Visibility makes sense if the data is relevant to the test. If the data is not relevant to the test, it is noise and we do not want to know about it.

orelmisan:
> Regarding "launcher, tests: Check report on failure cases" commit: What about successful tests are you going to change them to assert on the status that reported?

It was done in the [previous commit](https://github.com/kiagnose/kiagnose/pull/46/commits/0516103cedea0e2314893be7b3733768688a650b) (and stated in its commit message).

> Regarding "launcher, tests: Change test when initial report fails" commit the commit message title is not clean, do you think you can add more info about what is going to change?

The commit message says:
```
Replace the reporter stub with the production reporter code,
in case when the initial report fails.
```

What are you missing?

EdDev:
> > Regarding "launcher, tests: Change test when initial report fails" commit the commit message title is not clean, do you think you can add more info about what is going to change?
> 
> The commit message says:
> 
> ```
> Replace the reporter stub with the production reporter code,
> in case when the initial report fails.
> ```
> 
> What are you missing?

Maybe he is missing the reasoning behind it.
Also, I am not sure if `in case when the initial report fails` is a valid sentence (that `when` part does not fit).

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/6af425456486da401b1ce5c577bdd47f67be386e..50872a90de1cc4f10dff18d41f3f40b85446b193): Updated last commit message.

EdDev:
> This is done in order to use the production reporter whenever possible.

I am not sure if this qualifies to reasoning.
How about this commit:
```
launcher, tests: Use production reporter instead of stub

The test that covers the scenario in which the report fails to
post the start timestamp is currently using a reporter stub.

In order to cover a more realistic scenario (and setup), the
stub is replaced with the production reporter.
```

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/50872a90de1cc4f10dff18d41f3f40b85446b193..6f563a7ba21053ef518ffa931586f6f0001831c4): Changed last commit message.

kiagnose/internal/launcher/launcher_test.go
ormergi: Now I realized it doesnt mention `launcher.Run()`.

kiagnose/internal/launcher/launcher.go
ormergi: Alright, fair enough.

kiagnose/internal/launcher/launcher_test.go
ormergi: It doenst seem to be a problem, the expectedData can include what necessary in any case.

We can address it later this is nice to have.

kiagnose/internal/launcher/launcher.go
ormergi: It will matter when we will log the status that is about to be written to the ConfigMap, but this is not the scope of this PR.

ormergi (approved):
Thanks

PR #47, opened 2022-05-08
[WIP] config: Read from previously fetched ConfigMap

orelmisan:
Change mainflow to read raw ConfigMap object, so it could be passed
to config.ReadFromConfigMap(), and to the future Reporter.New().

This change reduces the amount of failure cases handled by config.

### DIFF
--- a/kiagnose/internal/config/config.go
+++ b/kiagnose/internal/config/config.go
@@ -28,7 +28,6 @@ import (
 
 	"k8s.io/client-go/kubernetes"
 
-	"github.com/kiagnose/kiagnose/kiagnose/internal/configmap"
 	"github.com/kiagnose/kiagnose/kiagnose/internal/rbac"
 )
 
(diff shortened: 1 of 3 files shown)


EdDev (commented):
I am not exactly convinced this is a good thing.
You have just exposed up the stack that we use a configmap.
Our original intention was to hide these details so changing them will have as little as possible effect on the upper layers.

orelmisan:
The advantages as I see them:
1. The ConfigMap is fetched only once (and not twice by the config package and later by the reporter package).
This means that the data seen by the config and the reporter packages is the same.

2. The reporter could assume that it receives a valid ConfigMap object, which was already checked by the config.
This simplifies the reporter and its unit tests.

The disadvantage is that mainflow is doing this fetch.

EdDev:
> The advantages as I see them:
> 
>     1. The ConfigMap is fetched only once (and not twice by the config package and later by the reporter package).
>        This means that the data seen by the config and the reporter packages is the same.

The current abstraction separates the config (input) and the report (output).
The two also hide the implementation backend (configmap).

You could keep both by adding another entity, like `State` that encapsulates the ConfigMap.

>     2. The reporter could assume that it receives a valid ConfigMap object, which was already checked by the config.
>        This simplifies the reporter and its unit tests.

You are describing a tight coupling between the config and report.
I do think this can be achieved by other means (perhaps the `Status` mentioned earlier).

> The disadvantage is that mainflow is doing this fetch.

orelmisan:
Not relevant at the moment.

PR #49, opened 2022-05-10
kubevirt-vm-latency: Integrate K8S API client

ormergi:
Currently the Kubevirt VM latency cannot interact with the cluster API and thus wont actually update the results
`ConfigMap` with the checkup results.

This PR adds Kubevirt client-go dependencies, integrate it to `client` package and changes
`reporter` code and tests enables it to update the results `ConfigMap`.

With this PR changes Kubevirt VM latency checkup can be tested end-to-end with the framework.
Note that it will always fail with the following error `Kubevirt VM latency checkup failed: run: not implemented`

~~Depends on #41 (first 5 commits)~~


ormergi:
Rebased

checkups/kubevirt-vm-latency/go.mod
```diff
@@ -4,8 +4,66 @@ go 1.17
 
 require github.com/stretchr/testify v1.7.1
 
+// Kubevirt dependencies
```
orelmisan, line 7: Is it possible to switch the order between the first and second commits?
So the vendored dependencies will come before the usage.

checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
```diff
@@ -32,6 +38,20 @@ func New() (*Client, error) {
 	return &Client{c}, nil
 }
 
-func (c *Client) UpdateConfigMap(_, _ string, _ map[string]string) error {
+func (c *Client) UpdateConfigMap(namespace, name string, date map[string]string) error {
```
orelmisan, line 41: Please consider moving this function to the reporter.
Only the reporter has a use for it.

checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
```diff
@@ -32,6 +38,20 @@ func New() (*Client, error) {
...
+	const updateConfigMapTimeout = time.Minute * 3
+	ctx, cancel := context.WithTimeout(context.Background(), updateConfigMapTimeout)
+	defer cancel()
```
orelmisan, line 44: Please consider removing it for now, and adding it in the future if there will be a need.
I suggest for now to use `context.Background` for simplicity.

checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
```diff
@@ -19,10 +19,17 @@
 
 package client
 
-type Client struct{}
+import "kubevirt.io/client-go/kubecli"
+
+type Client struct{ kubecli.KubevirtClient }
```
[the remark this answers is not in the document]
EdDev: I think it was done to allow stubbing.
The questions is... what is the cost of doing this now (it probably affects tests and production code substantially).

checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
EdDev: It would be helpful if you explain why.

@ormergi , from your side, I am interesting also to understand why this is needed.
Are you saying that any of these calls are blocking? It does not seem so to me, but I am not really sure.

orelmisan: All our calls to the K8s API were called with the `context.Background`.
We haven't considered a policy for timeouts yet.
Why is this case different?

checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
ormergi: The intention of `client` is to consolidate interactions with the cluster including updating results ConfigMap (and VM lifecyce that introduced at #55), it enables replacing the client with a stub in tests instead of using k8s fake client.

checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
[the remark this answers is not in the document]
ormergi: The cost is that the `client` will need to implement each operation to k8s API, such as updating ConfigMap, create VM's etc..
When working on #55 I noticed that kubvirt.io/client-go wont provide fake client for VM's (it provides mocks instead),
thus creating this `client` seemed better choose than using k8s fake client.

I suggest to start with this implementation and when there will be a need for fancier stuff we can change the implementation to how it is at kiagnose and use k8s fake client, WDYT?

checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
EdDev: > it enables replacing the client with a stub in tests instead of using k8s fake client.

Which is a bit odd, considering this is what we did in kiagnose core.

checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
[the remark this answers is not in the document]
EdDev: Well, it is a mess. `KubevirtClient` is an interface, and now it is embedded in a concrete struct, which has a method.
I do not know how to solve it now, if any of you have a concrete solution, please try out and then suggest it.

It feels we leave a tail of mud behind us, but we also have time restraints and should try and do things in parallel: Go forward and continue and cleanup while in motion (i.e. as an after-step).

checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
ormergi: 1. From what I saw in the code of the client both operations are blocking, in general it better to limit such operations
    with a timeout and fail faster.

2. We do set a timeouts at the framework [[1]](https://github.com/kiagnose/kiagnose/blob/main/kiagnose/internal/checkup/checkup.go#L290) [[2]](https://github.com/kiagnose/kiagnose/blob/main/kiagnose/internal/checkup/checkup.go#L309) (there is a tech death to use context and timeouts to allow graceful shutdown)

3. IMO there is no need to come up with formal timeouts policy, time framing operations against an API endpoint
is pretty standard as far as I know.

ormergi:
Rebased in order to take #61 changes.

checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
orelmisan: > 1. From what I saw in the code of the client both operations are blocking, in general it better to limit such operations
>    with a timeout and fail faster.

The operations are indeed blocking.

> 2. We do set a timeouts at the framework [[1]](https://github.com/kiagnose/kiagnose/blob/main/kiagnose/internal/checkup/checkup.go#L290) [[2]](https://github.com/kiagnose/kiagnose/blob/main/kiagnose/internal/checkup/checkup.go#L309) (there is a tech death to use context and timeouts to allow graceful shutdown)

There is no timeout on low level K8s API calls in the kiagnose core.

> 3. IMO there is no need to come up with formal timeouts policy, time framing operations against an API endpoint
>    is pretty standard as far as I know.

I agree with the need, but suggest to add it in a later stage.

ormergi: I prefer to do it now and not drag more death for later, as you can see its pretty straight forward (passing context with timeout down the chain..)

checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
ormergi: @EdDev note that later on we saw that we need to hack around the k8s fake client for our needs, 
that eventually made tests setup not trivia (normalizing the client objects tracker, injecting errors).
It seems that we can avoid it for now by working with stub client.

checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
orelmisan: The way I see it is introducing a new thing to the project, same as the error type that was discussed yesterday in PR #41.

EdDev: I have seen no usage of timeouts to API calls that are expected to perform a simple operation. The only usage I have seen this used was for `wait*` operations, which by definition are long blocking (waiting for a condition to be reached).

I think we should avoid adding new patterns implicitly. Please try to suggest such improvements through dedicated PR/s.
The basic guideline is what was merged so far and if there is a problem that requires changing the existing, a change suggesting that should be placed, fixing it.

ormergi: Thanks for explaining, I will remove it.

Regarding adding new patterns implicitly, I think its a bit excessive when you need to introduce some new functionality
when there is no similar example for it in the code base.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/53a2227361d8aa86da75b42b421b1354472c083a..03e9fb90455a5c4eb3e8f935ba155fab67d4e438): remove timeouts

checkups/kubevirt-vm-latency/go.mod
orelmisan: Thank you for the answer.

orelmisan (approved):
Thank you @ormergi.
In the near future will have to see how can we improve the client with regard to the usage of it in the tests.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/03e9fb90455a5c4eb3e8f935ba155fab67d4e438..34ea6c6ba050c430e461abc4b75f01fc8adf8721): Split first commit to two, one for adding Kubveirt client-go module, and another for adding replace rules.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/34ea6c6ba050c430e461abc4b75f01fc8adf8721..c9b941ac6139cd3355056e67d3626dc662ae1427): Add missing dependency and reorder the commits so go-modules changes and vendoring added after code changes.

checkups/kubevirt-vm-latency/go.mod
```diff
@@ -4,8 +4,66 @@ go 1.17
 
 require github.com/stretchr/testify v1.7.1
 
+// Kubevirt client-go dependencies
 require (
-	github.com/davecgh/go-spew v1.1.0 // indirect
+	k8s.io/apimachinery v0.23.5
```
orelmisan, line 9: Why is `k8s.io/apimachinery` in version v0.23.5 while `k8s.io/api` and `k8s.io/client-go` are pinned to v0.23.1?

checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
```diff
@@ -19,7 +19,12 @@
...
 
-import "kubevirt.io/client-go/kubecli"
+import (
+	"context"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"kubevirt.io/client-go/kubecli"
+)
 
 type Client struct{ kubecli.KubevirtClient }
```
orelmisan, line 29:
```suggestion
type Client struct{ 
  c kubecli.KubevirtClient,
}
```

orelmisan (commented):
To be honest, I don't understand the last three commits.

checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
EdDev: @orelmisan , this is not the same... It is not needed.

orelmisan: I see, thanks for the clarification.

checkups/kubevirt-vm-latency/go.mod
ormergi: Sorry this is a mistake, I updated go.mod aligned the versions and  removed unnecessary replacement rules, please take another look.

ormergi:
> To be honest, I don't understand the last three commits.

@orelmisan following our discussion with @EdDev, it was concluded that its better to split the changes of go.mod
to different commit where each will explain what is being done and why, here is a brief description of last three commits:
- First one is a result of `go get kubeirt.io/client-go & go get k8s.io/apimachiney`.
- Second commit, pins the version of the module that Kubevirt client-go uses.
- Third commit, results of `go mod tidy & go mod vendor`.

Let me know if it helps.

orelmisan:
> > To be honest, I don't understand the last three commits.
> 
> @orelmisan following our discussion with @EdDev, it was concluded that its better to split the changes of go.mod to different commit where each will explain what is being done and why, here is a brief description of last three commits:
> 
> * First one is a result of `go get kubeirt.io/client-go & go get k8s.io/apimachiney`.
> * Second commit, pins the version of the module that Kubevirt client-go uses.
> * Third commit, results of `go mod tidy & go mod vendor`.
> 
> Let me know if it helps.

Can it please be reflected in the commit message?

EdDev (approved):
Thank you!

orelmisan (approved):
Thank you!

PR #51, opened 2022-05-10
linter: Increase hugeParam threshold size (to 1024)

EdDev:
The hugeParam linter warns against passing a large amount of data to a
function, causing a large copy.

The cost of a copy is not always obvious.
There are several reasons against converting data to a pointer:
- Readability: From the caller stand of view, passing a pointer may be
  interpreted as if the function is going to mutate the data.
- The compiler may optimize the machine code such that there is no copy.
- Passing a pointer may imply usage of the heap, which is also costly
  (compare to the stack).

Increasing the threshold value higher, gives more control to
the code author to decide what fits as a pointer and what not.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -24,6 +24,9 @@ linters-settings:
       - octalLiteral
       - whyNoLint
       - wrapperFunc
+    settings:
+      hugeParam:
+        sizeThreshold: 1024
   gocyclo:
     min-complexity: 15
   goimports:


orelmisan (commented):
Thank you for the PR @EdDev.

As we've discussed offline, can you please rephrase the following sentence in the commit message and the PR description?
```
The cost of a copy is not trivial..
```

I initially understood as as "The cost of a copy is expensive", but you've meant something like "The cost of a copy is not always obvious".

orelmisan (approved):
Thank you!

PR #48, opened 2022-05-10
reporter: Report checkup results

orelmisan:
Until now, the results reported by the checkup hasn't been read by the framework.

Read and report the checkup's reported results:
1. `succeeded`.
2. `failureReason`.
3. Checkup custom results (optional).

The "succeeded" field value is a "true", when no errors had occurred in the framework or in the checkup.

The `failureReason` field is a concatenation of:
1. The value reported by the checkup (if exists).
2. `Checkup.Run()` error (if occurred).
3. `Checkup.Results()` error (if occurred).
4. `Checkup.Teardown()` error (if occurred).

~~Depends on PR #46.~~

PR #53, opened 2022-05-10
Do not use pointers when not needed

EdDev:
When function receivers or arguments are defined as pointers, beyond the
functional need it also exposes an intent.

When reading the code, an expression like `foo(&myvar)` has the
potential to be interpreted as: "foo is mutating myvar".
On the other hand, seeing `foo(myvar)` is more likely to express to the
reader that the intent is to keep the variable immutable.

The PR reverts the move to pointers due to a linter requirement [1].

[1] https://github.com/kiagnose/kiagnose/pull/48

~Depends on #51 (for this one to pass the checks).~

### DIFF
--- a/kiagnose/internal/launcher/launcher.go
+++ b/kiagnose/internal/launcher/launcher.go
@@ -35,7 +35,7 @@ type workload interface {
 }
 
 type reporter interface {
-	Report(*status.Status) error
+	Report(status.Status) error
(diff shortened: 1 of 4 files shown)


orelmisan (approved):
Thank you!

EdDev:
change: Rebase

PR #54, opened 2022-05-11
configmap: Use default K8s clientset

orelmisan:
Until now, Checkup held as a state, a custom "client" interface which was a small subset of the default "kubernetes.Interface"
client set.

The custom client interface was set up because we didn't know in advanced if K8s fake client will be suitable for our needs.

The custom client interface was forcing its users to use a specific client version (for example CoreV1) instead of having
a choice to select a client from the default client set.

This had a system-wide affect, affecting all functions that were called from Checkup, and using the client.

Because the "configmap" package was affected, other packages were affected as well.

Holding a client from type "kubernetes.Interface", allows its users a choice to select the most suitable client.

### DIFF
--- a/kiagnose/internal/checkup/checkup.go
+++ b/kiagnose/internal/checkup/checkup.go
@@ -288,7 +288,7 @@ func (c *Checkup) Run() error {
 }
 
 func (c *Checkup) Results() (results.Results, error) {
-	return results.ReadFromConfigMap(c.client.CoreV1(), c.resultConfigMap.Namespace, c.resultConfigMap.Name)
+	return results.ReadFromConfigMap(c.client, c.resultConfigMap.Namespace, c.resultConfigMap.Name)
 }
 
 func (c *Checkup) SetTeardownTimeout(duration time.Duration) {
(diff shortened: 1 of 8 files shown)


EdDev (dismissed):
Nice, thank you!

It is worth mentioning that the reason we can do this is due to the usage of the client `fake`.
Without it, the smaller interfaces are a better fit as it allows stubbing only the needed parts (and not all the big interface).

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/43e91bcb04696718cd845bca83f5feba7aa379e9..ea64692b8e90508c1de102735422e08d7fd3a22e): Updated commit message and PR description.

EdDev (approved):

ormergi (commented):
I prefer to keep using `client.Corv1Interface`  in `configmap` helpers (which is embedded in `kubernetes.Interface`), it improve the visibility and let one realize which API actually being used.
Having it like that makes the code cleaner as only the objects that are necessary being passed to the helpers IMO (similar to not passing VMI object just for using its Name attribute).
Same goes for 56-59 PRs which introduce similar changes to different packages.

orelmisan:
@ormergi Checkup is a higher level component, it is not supposed to be aware of lower level details such as client types.

lets say for example that tomorrow there will be a ConfigMap in the coreV2 group that we'll want to use, without this PR, we will have to change a lot of code to get the corev2 client.

When lower-level functions receives the `kubernetes.Interface` they can select a client, without higher level components caring about it.

So this change actually increases the decoupling between higher and lower level components.

ormergi (approved):
LGTM

PR #56, opened 2022-05-11
checkup, namespace: Use default K8s client-set

orelmisan:
Change the client type to default client-set (`kubernetes.Interface`) in functions belonging to the checkup.namespace package.

Depends on PR #54.


EdDev:
LGTM, thanks.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/9d55b45d90b1e9b8f544c19bd2a208cdb6e37b5f..f3231175181fe037833927cfde83434bc7f3d6ca): Rebase.

PR #57, opened 2022-05-11
checkup, job: Use default K8s client-set

orelmisan:
Change the client type to default client-set (kubernetes.Interface) in functions belonging to the checkup.job package.

~~Depends on PR #54.~~


orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/3de5aa706b2652b2b3bde4617ea0e856076034f5..6e60f90a5baca326c2ec5ae7ac98ed825f6465d5): Rebase.

ormergi (approved):
LGTM

PR #58, opened 2022-05-11
checkup, ServiceAccount: Use default K8s client-set

orelmisan:
Change the client type of `serviceaccount.Create()` to the default client-set (kubernetes.Interface).

~~Depends on PR #54.~~

### DIFF
--- a/kiagnose/internal/checkup/checkup.go
+++ b/kiagnose/internal/checkup/checkup.go
@@ -217,7 +217,7 @@ func (c *Checkup) Setup() error {
 		}
 	}()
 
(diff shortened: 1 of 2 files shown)


orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/6b694374cc677964c08658322807a4b0bce89e64..76352ee3c8dd627f245ef2e708bd642eb52de9de): Rebase.

ormergi (approved):
LGTM

EdDev (approved):

PR #59, opened 2022-05-11
rbac: Use default K8s client-set

orelmisan:
Change the client type to the default client-set (kubernetes.Interface) in functions belonging to the rbac package (except for `CreateServiceAccount()` that is handled in PR #58).

Depends on PR #54.


orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/98ce423e5c11620673b1768c2c4772659b651080..ee69e0d62dfa5b1b743e9891630d58f973c85eae): Rebase.

EdDev (approved):
Thanks!

PR #61, opened 2022-05-11
CI, checkup/vm-latency: Update linter version and args

EdDev:
The timeout is increased due to the expected codebase size.
The scanned folders are now explicitly specified.

The linter configuration is now in sync with the one from kiagnose core.


orelmisan (approved):
Thank you!

PR #63, opened 2022-05-16
config: Add check if ConfigMap is already in use

orelmisan:
At the moment, it is possible to run two (or more) concurrent  framework Jobs, which refer to the same ConfigMap object as an input.
The results of such a scenario are undefined because there will be a race when writing the results.

It is also possible to run two (or more) consecutive framework Jobs, which refer to the same ConfigMap object as an input.
In this case, the results of the first run will be overwritten with the results of the second run.

Add a check if the input ConfigMap object is already in use by another checkup Job, and fail if it is.

The check is looking for the "StartTimestamp" field reported by the reporter.

In order to not duplicate the "reporter.StartTimestampKey" constant, config is now dependent on reporter.

Because the API is about to change from a ConfigMap to a CRD in the near future, it seems reasonable to read the constant from the `reporter` package instead of starting a new package which will include only one constant.


EdDev (commented):
The reporter caches the `ConfigMap` state and populates that cache once on the first call (when it needs to post the start timestamp).
I think you can detect this issue there (on first call to `Report()`, saving use the burden to do all these hoops.

Could you check if it works?

orelmisan:
> The reporter caches the `ConfigMap` state and populates that cache once on the first call (when it needs to post the start timestamp). I think you can detect this issue there (on first call to `Report()`, saving use the burden to do all these hoops.
> 
> Could you check if it works?

The suggested change catches the issue before the config is parsed, and `Checkup`, `Reporter` and `Launcher` are instantiated.
Your suggestion could work, but after the list of objects above has been instantiated, and `Launcher.Run()` had also begun.
Furthermore, to catch this condition when parsing the input, seems more natural than trying to catch it on first output.

As a compromise, I suggest to create a new package that will contain all the constants from `config` and `reporter` packages, so `config ` won't be dependent on `reporter`.
WDYT?

EdDev:
> As a compromise, I suggest to create a new package that will contain all the constants from `config` and `reporter` packages, so `config` won't be dependent on `reporter`.
> WDYT?

You could export the constants of the key names to a `types` package. I am fine with that.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/af6447089d17bcaebf2a476fd7285b89720f4bc7..bf91a0aa5cd5d53ea484e2c7bc8429c460156c8f): 
1. Rebase
2. Created the `types` package to hold the field names of `config` and `reporter`.
3. Moved the check if the ConfigMap is already in use to `cmparser.go`.

kiagnose/internal/types/types.go
```diff
@@ -0,0 +1,28 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright 2022 Red Hat, Inc.
+ *
+ */
+
+package types
```
EdDev, line 20: I think this is suppose to be public, not internal.
It is part of the kiagnose API, no?

kiagnose/internal/config/cmparser.go
```diff
@@ -55,6 +56,10 @@ func (cmp *configMapParser) Parse() error {
 		return ErrConfigMapDataIsNil
 	}
 
+	if cmp.isConfigMapAlreadyInUse() {
```
EdDev, line 59: IMO it should be placed in a validation step, not hidden in the parser.
The parser should do its thing (parse) and the result should be validated (at `mainflow`).

If am not going to insist, but I think it makes more sense.
Let me know your thoughts.

kiagnose/internal/types/types.go
orelmisan: Up until now, these constants were in the `config` and `reporter` packages, which are also internal.
I think that we should have a dedicated API package when we'll make the transition to a CRD.

EdDev: I am saying that these are keys that interest anyone that wants to work with it programmatically.
The CR will have these public for sure.

orelmisan: Moved the `types` package up one level.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/0078f2962f18c770c734ad3fe2172c7305226966..f5add576702afa588026eb26258562b71e9373f7): Moved the`types` package up one level to make it public.

PR #65, opened 2022-05-17
automation, e2e: Set default binaries with full path

EdDev:
The default binaries used in the e2e (sanity) tests (`kind` and
`kubectl`) have been set with a relative path (`./`).

This does not work well when running locally.
To solve this, an explicit full path is used instead.


orelmisan (approved):
Thank you!

PR #71, opened 2022-05-18
kubevirt-vm-latency: Do not set node-selector when node name is empty

ormergi:
In order to ensure a VM is scheduled on specific node, the VM node
selector is set with "kubernetes.io/hostname" label.

Setting this label with an empty string (i.e: kubernetes.io/hostname="")
cause the VMs to stuck on 'Scheduling' phase as it looks for a node
that does not exists.
It seems that the value of "kubernetes.io/hostname" label is no validated.


ormergi:
> typo in the commit message: `is no validated` should be `is not validated`.

@EdDev thanks for the heads up, fixed.

EdDev (approved):
Thanks!

PR #70, opened 2022-05-18
kubevirt-vm-latency: Use fedora image

ormergi:
A VM considred ready when AgentConnected condition is met,
meaning that the VM had booted and QEMU Guest Agent (GA) is running.

It seems that the GA wont start on alpine image causing the checkup to
always with VM not read error.

Signed-off-by: Or Mergi <ormergi@redhat.com>

PR #74, opened 2022-05-18
latency, vmi, spec: Rename NewAlpine() to NewFedora()

orelmisan:
This PR is a follow up to PR #70 - the OS image was switched from Alpine to Fedora.
Reflect this change in the name of the function that creates this VMI.


EdDev (approved):
Missed it, thanks!

PR #80, opened 2022-05-23
latency, CI, e2e: Move general targets to main e2e.sh

orelmisan:
The following targets where performed from the latency checkup's `e2e.sh` file:
```
--install-kind
--install-kubectl
--create-cluster
--deploy-kiagnose
```

These targets are general to all checkups, so they were moved to the main `e2e.sh` file.

The `--delete-cluster` target in the checkup's `e2e.sh` script was redundant, because it is also present at the main `e2e.sh` script.


EdDev (approved):
Thank you!

PR #82, opened 2022-05-23
checkup, job: Remove redundant event log

EdDev:
Events received about the checkup job are currently logged by kiagnose,
but there seem to be no useful reason for such a debug level logging.
(it may be worth the effort once issues are found in this area and a
proper logging library is used to support setting levels of logs)


orelmisan (approved):
Thank you!

PR #12, opened 2022-04-04
docs: Add VM latency checkup design document

orelmisan:
This adds a design document for the VM latency checkup.

Signed-off-by: Orel Misan <omisan@redhat.com>

PR #11, opened 2022-04-03
[WIP] docs: Add design document

orelmisan:
This adds the project's primary design document.

Signed-off-by: Orel Misan <omisan@redhat.com>

PR #83, opened 2022-05-23
kubevirt-vm-latency: Add README

ormergi:
Add README file for Kubevirt VM latency checkup


checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,116 @@
+# Kubevirt VM latnecy checkup
+
+Kubevirt VM latency enables performing network latency measurement between two VMs
```
orelmisan, line 3:
```suggestion
[Kubevirt](https://kubevirt.io/) VM latency checkup performs network latency measurement between two virtual machines 
```

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+4. `MAX_DESIRED_LATENCY_MILLISECONDS`
+
+   Maximal accepted network latency, if the actual latency is higher the checkup
+   will be considered as failed.
+
+
+5. `SOURCE_NODE`, `TARGET_NODE`
+
+   The two ends of the network latency check.
```
orelmisan, line 31: Please consider using a table.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,116 @@
...
+# build image 
+$ ./automation/make.sh --build-checkup-image
+
+# Using Docker to build the image:
+$ CRI=docker ./automation/make.sh --build-checkup-image
+```
+
+## Manual Execution Instructions
+### Prerequisites
+- Kubernetes cluster configured with SR-IOV network card (including [sriov-device-plugin](https://github.com/k8snetworkplumbingwg/sriov-network-device-plugin) and [sriov-cni](https://github.com/k8snetworkplumbingwg/sriov-cni)):
```
orelmisan, line 85: AFAIU SR-IOV support is optional.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,116 @@
...
+- [Kubevirt](https://github.com/kubevirt/kubevirt)
+- [Multus](https://github.com/k8snetworkplumbingwg/multus-cni)
+
+```bash
+# from Kiagnose root directory
+# deploy Kiagnose
+$ kubectl apply -f ./manifests/kiagnose.yaml
+
+# create SR-IOV NetworkAttachmentDefinition
+$ kubectl apply -f ./checkups/kubevirt-vm-latency/manifests/sriov-network-attachment-definition.yaml
```
orelmisan, line 95: Please consider to start with a bridge example, and explain usage over SR-IOV later.
Please consider giving a note regarding in which namespaces its optional to put the NetworkAttachmentDefinition object.

orelmisan (commented):
Thank you for the PR @ormergi!

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,116 @@
+# Kubevirt VM latnecy checkup
+
+Kubevirt VM latency enables performing network latency measurement between two VMs 
+over a given network.
+
```
orelmisan, line 5: Please also consider adding:
1. A description of the problem.
2. A description of the solution.
3. A block diagram of the objects participating in the checkup (VMIs, NetworkAttachmentDefinition, etc).

ormergi:
rebased

checkups/kubevirt-vm-latency/README.md
ormergi: The motivation for this checkup is to test SR-IOV networks,
we enable bridged network for CI and development proposes.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,116 @@
+# Kubevirt VM latnecy checkup
+
+Kubevirt VM latency enables performing network latency measurement between two VMs 
+over a given network.
```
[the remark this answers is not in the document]
ormergi: I prefer to not get into those details, the user should care for the network he wants to test.

checkups/kubevirt-vm-latency/README.md
ormergi: The motivation for this checkup is to test SR-IOV networks,
I prefer describe bridged networks as another option and not
the default.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,116 @@
+# Kubevirt VM latnecy checkup
```
EdDev, line 1: typo

checkups/kubevirt-vm-latency/README.md
EdDev: > The motivation for this checkup is to test SR-IOV networks

I do not remember this appearing in the [design proposal](https://github.com/kiagnose/kiagnose/pull/12/files).

checkups/kubevirt-vm-latency/README.md
ormergi: Done

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+> **_Note:_**
+> `NetworkAttachmentDefinition` is namespaced object and can be created on any other namespace.
+
+create cluster roles with the permissions required by the checkup 
+```bash
+$ kubectl apply -f ./manifests/clusterroles.yaml
+```
+deploy checkup
+```bash
+$ kubectl apply -f ./manifests/kubevirt-vm-latency-checkup.yaml
```
EdDev, line 169: I think it will be better to use `cat <<`. I prefer not to main such manifests.
Maybe we will have stuff like that in future e2e tests, then it may make sense to use them.

checkups/kubevirt-vm-latency/README.md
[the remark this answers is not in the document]
ormergi: I added examples both for bridge and SR-IOV networks

checkups/kubevirt-vm-latency/README.md
ormergi: I added examples both for bridge and SR-IOV networks

checkups/kubevirt-vm-latency/README.md
ormergi: At the current state SR-IOV networks are also supported.
I added examples both for bridge and SR-IOV networks.

checkups/kubevirt-vm-latency/README.md
ormergi: Thanks for the heads up, fixed.

checkups/kubevirt-vm-latency/README.md
ormergi: Do you expect all manifests to be applied like this? (including ConfigMap and ClusterRoles)

checkups/kubevirt-vm-latency/README.md
EdDev: Supported, yes. But not required.
The only required is to have a NAD defined in the system that is expected to work with the KubeVirt VMI.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+| `MAX_DESIRED_LATENCY_MILLISECONDS`                                                 | Maximal network latency accepted, if the actual latency <br/> is higher the checkup will be considered as failed. |
+| `SOURCE_NODE`<br/>`TARGET_NODE`                                                    | Two ends in the of the network latency measurement.                                                               |
+
+## Results
+Following Kiagnose APIm, the checkup will report the results to the results ConfigMap object:
+```yaml
+spec.image: quay.io/kiagnose/kubevirt-vm-latency-checkup:main
+spec.timeout: 5m
+spec.clusterRoles: |
+  kubevirt-vmis-manager
```
orelmisan, line 25: This name was changed in PR #97 to `kubevirt-vm-latency-checker`.

checkups/kubevirt-vm-latency/README.md
orelmisan: > 3. A block diagram of the objects participating in the checkup (VMIs, NetworkAttachmentDefinition, etc). 

The block diagram is missing.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+status.succeeded: "false"
+status.failureReason: "run: failed to run check: failed due to connectivity issue: 5 packets transmitted, 0 packets received"
+```
+
+## Quickstart
+## Prerequisites
+- Kubernetes cluster, nodes have [bridge CNI](https://github.com/containernetworking/plugins) installed.
+- [Multus](https://github.com/k8snetworkplumbingwg/multus-cni#quickstart-installation-guide)
+- [Kiagnose](https://github.com/kiagnose/kiagnose/blob/main/README.install.md/#kiagnose-installation)
+- [Kubevirt](https://kubevirt.io//quickstart_minikube/#deploy-kubevirt)
```
orelmisan, line 59: Please consider putting Kiagnose last in the list.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+create NetworkAttachmentDefinition specifying bridge CNI plugin.
+```bash
+$ cat <<EOF | kubectl apply -f -
```
orelmisan, line 63: @phoracek had [suggested](https://github.com/kiagnose/kiagnose/pull/93#discussion_r882380034) in PR #93, to remove the dollar sign, so copying and pasting these commands are easier.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+              "bridge": "kgnsbr0"
+          }
+      ]
+    }
+EOF
+```
+
+> **_Note:_**
+> `NetworkAttachmentDefinition` is namespaced object and can be created on any other namespace.
+
```
orelmisan, line 85: Please consider using sub-headings for the main steps such as:
1. Permissions deployment
2. Configuration
3. Execution

The main README has these steps written.

checkups/kubevirt-vm-latency/README.md
ormergi: The order shouldn't matter all components are mandatory.

checkups/kubevirt-vm-latency/README.md
ormergi: I think there is no need to sub-heading for those, each command is expected to run one after similar to bash script.
Its not that each command is from different domain.

ormergi (commented):
[Changes](https://github.com/kiagnose/kiagnose/compare/9648c828ac505131f774e9e8e163ac8ecffb4cdf..cb667ca16d5d4fb98b33008478b6b510761d0494): addressed code-review comments.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+Kubernetes cluster administrators are often required to check network connectivity and performance on various networks.
+Such checks might be required to do immediately after the cluster is deployed, as an acceptance test, when new networks are introduced or just as part of a troubleshooting effort.
+
+[Kubevirt](https://kubevirt.io/) VM latency checkup performs network connectivity and latency measurement between two virtual machines over a given network.
+
+It can greatly improve the life of cluster administrators by removing the burden of manually testing network connectivity and performance, reduce mistakes and generally save time.
+
+## Parameters
+| Environment Variable Name                                                          | Remark                                                                                                            |
```
orelmisan, line 11: Please consider replacing `Remark` with `Description`.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+|:-----------------------------------------------------------------------------------|:------------------------------------------------------------------------------------------------------------------|
```
orelmisan, line 12: Why are there two colon signs (`:`) in this line?

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+| `RESULT_CONFIGMAP_NAMES`<br/>`RESULT_CONFIGMAP_NAMESPACE`                          | Results `ConfigMap` object.                                                                                       |
+| `NETWORK_ATTACHMENT_DEFINITION_NAME`<br/>`NETWORK_ATTACHMENT_DEFINITION_NAMESPACE` | `NetworkAttachmentDefinition` object on which <br/> latency measurement is performed on.                          |
+| `SAMPLE_DURATION_SECONDS`                                                          | Network latency measurement sample time.                                                                          |
+| `MAX_DESIRED_LATENCY_MILLISECONDS`                                                 | Maximal network latency accepted, if the actual latency <br/> is higher the checkup will be considered as failed. |
+| `SOURCE_NODE`<br/>`TARGET_NODE`                                                    | Two ends in the of the network latency measurement.                                                               |
+
+## Results
+Following Kiagnose APIm, the checkup will report the results to the results ConfigMap object:
```
orelmisan, line 20: Typo.
```suggestion
Following Kiagnose API, the checkup will report the results to the results ConfigMap object:
```

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+              "type": "bridge",
+              "bridge": "kgnsbr0"
+          }
+      ]
+    }
+EOF
+```
+
+> **_Note:_**
+> `NetworkAttachmentDefinition` is namespaced object and can be created on any other namespace.
```
orelmisan, line 84:
```suggestion
`NetworkAttachmentDefinition` is a namespaced object and can be created on any other namespace.
```

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+## Quickstart
+## Prerequisites
+- Kubernetes cluster, nodes have [bridge CNI](https://github.com/containernetworking/plugins) installed.
+- [Multus](https://github.com/k8snetworkplumbingwg/multus-cni#quickstart-installation-guide)
+- [Kubevirt](https://kubevirt.io//quickstart_minikube/#deploy-kubevirt)
+- [Kiagnose](https://github.com/kiagnose/kiagnose/blob/main/README.install.md/#kiagnose-installation)
+
+create NetworkAttachmentDefinition specifying bridge CNI plugin.
```
orelmisan, line 61:
```suggestion
Create NetworkAttachmentDefinition specifying bridge CNI plugin.
```

checkups/kubevirt-vm-latency/README.md
orelmisan: > I do not think we want to show explicit examples with SR-IOV NAD, except just referencing one. If you want to show an example, the use something that will work regardless of the HW.
> 
> I am also not aware of any such formal motivation, see the previous response.

Following PR #106, I agree with @EdDev's comment.
IMO we just need to list the supported CNI types that are parsed from the NAD.

checkups/kubevirt-vm-latency/README.md
orelmisan: @ormergi there is no block diagram showing how the checkup creating two VMIs, writing to the results ConfigMap, etc.

What does it have to do with a user knowing or not knowing how a NetworkAttachmentDefinition works?

checkups/kubevirt-vm-latency/README.md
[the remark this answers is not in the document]
orelmisan: I think it worth mentioning (somewhere in the document) that the Pod network is disabled on purpose, and that the ping is running on a secondary network.

checkups/kubevirt-vm-latency/README.md
orelmisan: As a user first seeing this project, I guess the first thing I'll read is the main `README.md` file.
In this file, there are several steps to the checkup life-cycle:
1. Checkup Installation
2. Checkup Configuration
3. Checkup Execution
4. Checkup Results Retrieval
5. Checkup Removal

As a new reader, I would expect this steps to be reflected explicitly in every checkup's README file.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
+# Kubevirt VM latency checkup
+
+Kubernetes cluster administrators are often required to check network connectivity and performance on various networks.
```
orelmisan, line 3: This checkup is relevant to Kubevirt users, it is not applicable in a k8s cluster without Kubevirt.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+Such checks might be required to do immediately after the cluster is deployed, as an acceptance test, when new networks are introduced or just as part of a troubleshooting effort.
```
orelmisan, line 4:
```suggestion
Such checks are probably required:
1. After cluster's initial deployment, as an acceptance test.
2. When new networks are introduced.
3. As part of a troubleshooting effort.
```

checkups/kubevirt-vm-latency/README.md
orelmisan, line 20: Please add a table describing the names, description and other relevant data regarding the output parameters.
This table already exists in PR #12.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+| `RESULT_CONFIGMAP_NAMES`<br/>`RESULT_CONFIGMAP_NAMESPACE`                          | Results `ConfigMap` object.                                                                                       |
+| `NETWORK_ATTACHMENT_DEFINITION_NAME`<br/>`NETWORK_ATTACHMENT_DEFINITION_NAMESPACE` | `NetworkAttachmentDefinition` object on which <br/> latency measurement is performed on.                          |
+| `SAMPLE_DURATION_SECONDS`                                                          | Network latency measurement sample time.                                                                          |
+| `MAX_DESIRED_LATENCY_MILLISECONDS`                                                 | Maximal network latency accepted, if the actual latency <br/> is higher the checkup will be considered as failed. |
+| `SOURCE_NODE`<br/>`TARGET_NODE`                                                    | Two ends in the of the network latency measurement.                                                               |
+
+## Results
+Following Kiagnose APIm, the checkup will report the results to the results ConfigMap object:
+```yaml
```
orelmisan, line 21: I suggest to add a sub-tittle `Example` above this example.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+| `RESULT_CONFIGMAP_NAMES`<br/>`RESULT_CONFIGMAP_NAMESPACE`                          | Results `ConfigMap` object.                                                                                       |
+| `NETWORK_ATTACHMENT_DEFINITION_NAME`<br/>`NETWORK_ATTACHMENT_DEFINITION_NAMESPACE` | `NetworkAttachmentDefinition` object on which <br/> latency measurement is performed on.                          |
+| `SAMPLE_DURATION_SECONDS`                                                          | Network latency measurement sample time.                                                                          |
+| `MAX_DESIRED_LATENCY_MILLISECONDS`                                                 | Maximal network latency accepted, if the actual latency <br/> is higher the checkup will be considered as failed. |
+| `SOURCE_NODE`<br/>`TARGET_NODE`                                                    | Two ends in the of the network latency measurement.                                                               |
+
```
orelmisan, line 18: Please add an example yaml configuration.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+status.succeeded: "true"
+status.failureReason: ""
+status.completionTimestamp: "2022-01-01T09:00:00Z"
+status.startTimestamp: "2022-01-01T09:00:07Z"
+status.result.avgLatencyNanoSec: "177000"
+status.result.maxLatencyNanoSec: "244000"
+status.result.measurementDurationSec: "7"
+status.result.minLatencyNanoSec: "135000"
+```
+
```
orelmisan, line 41: Please consider adding a sub-title `Failure Reasons`.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+In case of a connectivity issues between the VMs:
+```yaml
+status.succeeded: "false"
+status.failureReason: "run: failed to run check: failed due to connectivity issue: 5 packets transmitted, 0 packets received"
+```
+
+## Quickstart
+## Prerequisites
+- Kubernetes cluster, nodes have [bridge CNI](https://github.com/containernetworking/plugins) installed.
```
orelmisan, line 56:
```suggestion
- Kubernetes cluster, nodes have at least one of the supported CNIs installed.
```

With a link to the list of supported CNIs.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+          }
+      ]
+    }
+EOF
+```
+
+> **_Note:_**
+> `NetworkAttachmentDefinition` is namespaced object and can be created on any other namespace.
+
+create cluster roles with the permissions required by the checkup
```
orelmisan, line 86:
```suggestion
Apply the checkup's ClusterRole manifest, in order to supply the checkup the permissions it requires:
```

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+> **_Note:_**
+> `NetworkAttachmentDefinition` is namespaced object and can be created on any other namespace.
+
+create cluster roles with the permissions required by the checkup 
+```bash
+ kubectl apply -f ./manifests/clusterroles.yaml
+```
+deploy checkup
```
orelmisan, line 90:
```suggestion
Start checkup's execution:
```

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+```bash
+ kubectl apply -f ./manifests/kubevirt-vm-latency-checkup.yaml
+```
+wait for checkup to finish
```
orelmisan, line 94:
```suggestion
Wait for checkup to finish:
```

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+```
+view results
+```bash
+ kubectl get configmap kubevirt-vm-latency-checkup-config -n kiagnose -o yaml
+```
+clean up:
+```bash
+ kubectl delete -f ./manifests/clusterroles.yaml
+ kubectl delete -f ./manifests/kubevirt-vm-latency-checkup.yaml
+ kubectl delete -f ./manifests/sriov-network-attachment-definition.yaml
```
orelmisan, line 108:
```suggestion

```

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+  sleep 5
+done 
+```
+view results
+```bash
+ kubectl get configmap kubevirt-vm-latency-checkup-config -n kiagnose -o yaml
+```
+clean up:
+```bash
+ kubectl delete -f ./manifests/clusterroles.yaml
```
orelmisan, line 106: A user should remove the checkup's ClusterRole objects:
1. After the removal of the Jobs using it.
2. Only on checkup's removal.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+ ./automation/make.sh
+
+# build image 
+ ./automation/make.sh --build-checkup-image
+
+# Using Docker to build the image:
+ CRI=docker ./automation/make.sh --build-checkup-image
+```
+
+## SR-IOV networks
```
orelmisan, line 126: I think we can drop this section. Just mention that this checkup also support SR-IOV.

checkups/kubevirt-vm-latency/README.md
ormergi: It controls text alignment (left, right, center)

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+    }
+EOF
+```
+
+> **_Note:_**
+> `NetworkAttachmentDefinition` is namespaced object and can be created on any other namespace.
+
+create cluster roles with the permissions required by the checkup 
+```bash
+ kubectl apply -f ./manifests/clusterroles.yaml
```
orelmisan, line 88:
```suggestion
kubectl apply -f ./manifests/clusterroles.yaml
```

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+```
+deploy checkup
+```bash
+ kubectl apply -f ./manifests/kubevirt-vm-latency-checkup.yaml
```
orelmisan, line 92:
```suggestion
kubectl apply -f ./manifests/kubevirt-vm-latency-checkup.yaml
```

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,186 @@
...
+```
+wait for checkup to finish
+```bash
+until [ -n "$(kubectl get configmap kubevirt-vm-latency-checkup-config -n kiagnose -o jsonpath='{.data}' | grep completionTimestamp)" ]; do
+  sleep 5
+done 
+```
+view results
+```bash
+ kubectl get configmap kubevirt-vm-latency-checkup-config -n kiagnose -o yaml
```
orelmisan, line 102:
```suggestion
kubectl get configmap kubevirt-vm-latency-checkup-config -n kiagnose -o yaml
```

checkups/kubevirt-vm-latency/README.md
ormergi: I am no sure what list you expect, this checkup supports sriov-cni and bridge-cni it is reflected by each example.

orelmisan (commented):
There are redundant spaces before bash commands.
I've marked some, but there are more.

checkups/kubevirt-vm-latency/README.md
orelmisan:
```md
# Supported CNI types
This checkup supports the following CNI types:
1. bridge
2. cnv-bridge
3. sr-iov
...
```

checkups/kubevirt-vm-latency/README.md
ormergi: How about:
Create the ClusterRole with required permissions, the framework will later on grant the checkup with.

checkups/kubevirt-vm-latency/README.md
ormergi: I think it has value, note that the bridge CNI case exists for CI and dev proposes..

checkups/kubevirt-vm-latency/README.md
ormergi: I think that describing how result are reported deserves its own diagram at the framework readme/docs.
You mentioned that such diagram should include VMIs and NetAttachDef, so I assumed this is what you mean.

checkups/kubevirt-vm-latency/README.md
ormergi: There is one at line 22, isnt that what you mean?

checkups/kubevirt-vm-latency/README.md
ormergi: I changed it to what you suggest.

checkups/kubevirt-vm-latency/README.md
ormergi: I removed the snippets that creates NetAttachDefs.

checkups/kubevirt-vm-latency/README.md
ormergi: Done

checkups/kubevirt-vm-latency/README.md
ormergi: Done

checkups/kubevirt-vm-latency/README.md
ormergi: Done

checkups/kubevirt-vm-latency/README.md
ormergi: I changed the structure, now failure reasons are described under `Results`.

checkups/kubevirt-vm-latency/README.md
ormergi: I changed the structure, now the example you suggest exists.

checkups/kubevirt-vm-latency/README.md
ormergi: Done

checkups/kubevirt-vm-latency/README.md
ormergi: Done

checkups/kubevirt-vm-latency/README.md
ormergi: Done

checkups/kubevirt-vm-latency/README.md
ormergi: Done

checkups/kubevirt-vm-latency/README.md
ormergi: Done

ormergi:
@orelmisan I removed  redundant spaces all over, please take another look.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/a8c5e0eac7b746fbb8f2a7f790bd543288d2eb3a..e773a58b11992293e830e504a738977237a24366): Added block diagram

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,195 @@
+# Kubevirt VM latency checkup
+
+[Kubevirt](https://kubevirt.io/) users are often required to check network connectivity and performance of their virtualized workloads.
```
orelmisan, line 3:
```suggestion
[Kubevirt](https://kubevirt.io/) users are often required to check network connectivity and performance between their virtualized workloads.

```

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,195 @@
...
+Such checks might be required to do immediately after a cluster with Kubevirt is deployed, as an acceptance test, when new networks are introduced or just as part of a troubleshooting effort.
+
+Kubevirt VM latency checkup performs network connectivity and latency measurement between two virtual machines over a given network.
+
+It can improve the life of cluster administrators by removing the burden of manually testing network connectivity and performance, reduce mistakes and generally save time.
+
+![block diagram](./docs/kubevirt-vm-latency-diagram.png)
```
orelmisan, line 10: Thank you for adding the block diagram.
1. Can you please move the image to `docs/images`?
2. I've received a [comment](https://github.com/kiagnose/kiagnose/pull/11#discussion_r857284660) in PR #11 to change the image to svg format, can you please change it as well?

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Prerequisites
+- [Kiagnose](https://github.com/kiagnose/kiagnose/blob/main/README.install.md/#kiagnose-installation)
```
orelmisan, line 13: As I've answered in [this](https://github.com/kiagnose/kiagnose/pull/109#discussion_r890885971) thread in PR#109, can you please link the file in the repo?
When linking to main, a user using an older version - will see the installation instructions of the latest version, instead of the old one.

Furthermore, I suggest to put `Kiagnose` last in the list.
I think its fair to assume that an existing cluster already have:
K8s, kubevirt, multus and a NetworkAttachmentDefinition.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,195 @@
...
+- [Kubevirt](https://kubevirt.io//quickstart_minikube/#deploy-kubevirt)
+- [Multus](https://github.com/k8snetworkplumbingwg/multus-cni#quickstart-installation-guide)
+- Cluster node have one of the desired CNI plugins installed.
```
orelmisan, line 16:
```suggestion
- Cluster node have one of the desired [CNI](https://www.cni.dev/) plugins installed.
```

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,195 @@
...
+- `NetworkAttachmentDefinition` object to exists.
+
+## Supported CNI plugins
+- [bridge-cni](https://www.cni.dev/plugins/v0.7/main/bridge/)
```
orelmisan, line 20:
```suggestion
- [bridge-cni](https://www.cni.dev/plugins/current/main/bridge/)
```

Is there a reason to link to a specific version?

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+![block diagram](./docs/kubevirt-vm-latency-diagram.png)
+
+## Prerequisites
+- [Kiagnose](https://github.com/kiagnose/kiagnose/blob/main/README.install.md/#kiagnose-installation)
+- [Kubevirt](https://kubevirt.io//quickstart_minikube/#deploy-kubevirt)
+- [Multus](https://github.com/k8snetworkplumbingwg/multus-cni#quickstart-installation-guide)
+- Cluster node have one of the desired CNI plugins installed.
+- `NetworkAttachmentDefinition` object to exists.
+
```
orelmisan, line 18: Can you please add the SR-IOV requirements as optional?

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,195 @@
...
+## Supported CNI plugins
+- [bridge-cni](https://www.cni.dev/plugins/v0.7/main/bridge/)
+- [sriov-cni](https://github.com/k8snetworkplumbingwg/sriov-cni)
+
+> **_Note_**:
+> The SR-IOV CNI can be deployed through the [sriov-network-operator](https://github.com/k8snetworkplumbingwg/sriov-network-operator) or manually.
+
+## Permissions
+The checkup requires some additional permissions in order to operate
```
orelmisan, line 27:
```suggestion
The checkup requires some additional permissions in order to operate, in order to install them:
```

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,195 @@
...
+```
+
+## Configuration
+The checkup is by the following parameters:
+
+| Name                                                                               | Description                                                                                                                  |
+|:-----------------------------------------------------------------------------------|:-----------------------------------------------------------------------------------------------------------------------------|
+| `spec.image`                                                                       | The checkup container image.                                                                                                 |
+| `spec.timeout`                                                                     | Overall time the checkup can run.                                                                                            |
+| `spec.clusterRoles`                                                                | ClusterRole name with the required permission (optional).                                                                    |
```
orelmisan, line 39: > ClusterRole name with the required permission (optional).

`spec.clusterRoles` should have a specific value, and its mandatory.
Without it, the checkup will fail its preflight/setup.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,195 @@
...
+| `network_attachment_definition_namespace`<br/>`network_attachment_definition_name` | `NetworkAttachmentDefinition` object on which <br/> the VMs are connected to and measure network latency.                    |
+| `sample_duration_seconds`                                                          | Network latency measurement sample time (optional).<br/> Default is 5 seconds.                                               |
+| `max_desired_latency_milliseconds`                                                 | Maximal network latency accepted, if the actual latency <br/> is higher the checkup will be considered as failed (optional). |
+| `source_node`<br/>`target_node`                                                    | Two ends in the of the network latency measurement (optional).                                                               |
```
orelmisan, line 43: What happens if you only put one of them?

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+> **_Note_**:
+>`timeout` should be greater than `sample_duration_seconds`.
```
orelmisan, line 46: I suggest to write a minimum time for `timeout` (like 5m or so).
This time should take into account the time it takes from start to finish including the VMs spin up and teardown.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Clean up
+```bash
+kubectl delete job -n kiagnose kubevirt-vm-latency-checkup
+kubectl delete config-map -n kubevirt-vm-latency-checkup-config
+```
+Once the checkup is finished it's safe to remove the ClusterRole:
+```bash
+kubectl delete -f https://raw.githubusercontent.com/kiagnose/kiagnose/main/checkups/kubevirt-vm-latency/manifests/clusterroles.yaml
+```
```
orelmisan, line 178: This is actually the removal instructions.

orelmisan (commented):
Thank you for the changes @ormergi.

checkups/kubevirt-vm-latency/README.md
orelmisan: @EdDev WDYT?

checkups/kubevirt-vm-latency/README.md
EdDev: I expect not to have manifest files which we do not distribute.
And example manifest makes sense, bit it needs to be place under a different location (the example can include the job, but not the dependencies, e.g. NAD).

checkups/kubevirt-vm-latency/README.md
ormergi: Right, this checkup does require some additional permissions it is mentioned under `Permission`.
This is a general description of `spec.clusterRoles` which is not mandatory.

checkups/kubevirt-vm-latency/README.md
ormergi: Sorry I dont understand to which parameter you mean, please elaborate.

checkups/kubevirt-vm-latency/README.md
orelmisan: The general description of `spec.clusterRoles` is in the main README.
In this context, it is mandatory, and it has a specific value.

checkups/kubevirt-vm-latency/README.md
ormergi: Well this is a mistake I was going to change it to point v1.1.1.
I think its better to point to a version you verified that it works, on CI we use CNOA v0.74.0 which deploys bridbe-cni v1.1.1.

checkups/kubevirt-vm-latency/README.md
ormergi: Usage `of ` is right, the users should check connectivity and performance of the visualized workloads no necessary between them.

checkups/kubevirt-vm-latency/README.md
ormergi: The VM is scheduled on arbitrary node, as mentioned in the design doc.

checkups/kubevirt-vm-latency/README.md
ormergi: I changed the link to local file.

IMO the order doesn't really matter.

checkups/kubevirt-vm-latency/README.md
ormergi: Thanks for the heads up, I added the colon.

checkups/kubevirt-vm-latency/README.md
ormergi: I am not sure its the right place for it, not that at the example under `Configuration` there is a value for timeout.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,195 @@
...
+In case of a connectivity issues between the VMs:
+```yaml
+status.succeeded: "false"
+status.failureReason: "run: failed to run check: failed due to connectivity issue: 5 packets transmitted, 0 packets received"
+```
+
+### Example:
+Retrieve the checkup results:
+```bash
+kubectl get configmap kubevirt-vm-latency-checkup-config -n kiagnose -o yaml
```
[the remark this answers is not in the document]
ormergi: I find it make more sense to have the command right next to the result, WDYT?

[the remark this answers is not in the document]
ormergi: It also keeps the same format similar to `Configuration`

checkups/kubevirt-vm-latency/README.md
ormergi: I dont think removal is the right word (similar to install example), few objects are created in the cluster and this describes how to clean them up, WDYT?

checkups/kubevirt-vm-latency/README.md
ormergi: My intention was to add add a table that describes which parameter controls what.
Each parameter value is specified in the yaml of Example sections along this doc.

checkups/kubevirt-vm-latency/README.md
ormergi: Which requirements would you expect there to be?
Note that there is a reference for SR-IOV CNI and operator.

checkups/kubevirt-vm-latency/README.md
orelmisan: Connectivity is by definition between two or more entities.

checkups/kubevirt-vm-latency/README.md
orelmisan: The current wording is fine by me.

checkups/kubevirt-vm-latency/README.md
ormergi: Noting "between" means that it checks connectivity and performance only between the virtualized workloads which is not accurate.

checkups/kubevirt-vm-latency/README.md
ormergi: I changed the structure of the readme, now there is no explicit example for SR-IOV.

checkups/kubevirt-vm-latency/README.md
ormergi: The readme structure is changed, now there are no explicit examples.
I also added supported CNI types list.

checkups/kubevirt-vm-latency/README.md
ormergi: Thanks for explaining, I removed all NAD yamls for the readme.

checkups/kubevirt-vm-latency/README.md
orelmisan: There is no API results fields that gives information regarding on which nodes the VMs ran on.
I've opened an issue about it #113.

checkups/kubevirt-vm-latency/README.md
ormergi: Done

checkups/kubevirt-vm-latency/README.md
ormergi: Done

checkups/kubevirt-vm-latency/README.md
ormergi: Done

checkups/kubevirt-vm-latency/README.md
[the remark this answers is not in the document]
ormergi: The fact that the ping is preformed over the secondary network expressed well all over the readme, the diagram also describes it.
I didnt mention pod network to not district the reader from the main idea.

checkups/kubevirt-vm-latency/README.md
ormergi: In my opinion numbering add noise in this case, I replaced the apposition with parenthesis I think its easier to understand now.

ormergi:
Following offline discussion I resolved some of the conversations.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/e9d064480d4e3b05cb889ae82a87cca768b46c54..a9bd11c05e3d624c6685cd52b27abfce09b38a97): address https://github.com/kiagnose/kiagnose/pull/83#discussion_r887711634

orelmisan:
There are changes in the second commit that are not supposed to be there.
Anyway, I suggest to squash the two commits.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,195 @@
+# Kubevirt VM latency checkup
+
+[Kubevirt](https://kubevirt.io/) users are often required to check network connectivity and performance of their virtualized workloads.
+Such checks might be required to do immediately after a cluster with Kubevirt is deployed (as an acceptance test), when new networks are introduced or just as part of a troubleshooting effort.
+
+Kubevirt VM latency checkup performs network connectivity and latency measurement between two virtual machines over a given network.
+
+It can improve the life of cluster administrators by removing the burden of manually testing network connectivity and performance, reduce mistakes and generally save time.
+
+![block diagram](./docs/images/kubevirt-vm-latency-diagram.svg)
```
orelmisan, line 10: Is the image shown on your computer?
I think there is a problem with the path:
```suggestion
![block diagram](../docs/images/kubevirt-vm-latency-diagram.svg)
```

checkups/kubevirt-vm-latency/README.md
orelmisan: I think we can just leave it.
As long as the user provides the `NetworkAttachmentDefinition` that matches the supported CNI, its outside of the checkup's scope.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,195 @@
...
+- [Multus](https://github.com/k8snetworkplumbingwg/multus-cni#quickstart-installation-guide)
+- Cluster node have one of the desired [CNI](https://www.cni.dev/) plugins installed.
+- `NetworkAttachmentDefinition` object to exists.
+
+## Supported CNI plugins
+- [bridge-cni](https://www.cni.dev/plugins/v1.1/main/bridge/)
+- [sriov-cni](https://github.com/k8snetworkplumbingwg/sriov-cni)
+
+> **_Note_**:
+> The SR-IOV CNI can be deployed through the [sriov-network-operator](https://github.com/k8snetworkplumbingwg/sriov-network-operator) or manually.
```
orelmisan, line 24: nit: Please consider dropping this note.
As said in a previous comment, its outside of the checkup's scope.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Permissions
+The checkup requires some additional permissions in order to operate:
+```bash
+kubectl apply -f ./manifests/clusterroles.yaml
+```
+
+## Configuration
+The checkup is by the following parameters:
```
orelmisan, line 33:
```suggestion
The checkup is configured by the following parameters:
```

checkups/kubevirt-vm-latency/README.md
orelmisan:
```suggestion
| `spec.clusterRoles`                                                                | ClusterRole name with the required permissions.                                                                    |
```

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+> **_Note_**:
+> `CONFIGMAP_NAMESPACE` and `CONFIGMAP_NAME` environment variables are required in order to pass the checkup configuration to Kiagnose.
+
+Wait for the checkup to finish:
+```bash
+until [ -n "$(kubectl get configmap kubevirt-vm-latency-checkup-config -n kiagnose -o jsonpath='{.data}' | grep completionTimestamp)" ]; do
+  sleep 5
+done 
+```
```
orelmisan, line 104: This code snippet will run forever if the kiagnose or the checkup's image are not fetched correctly.
Please consider:
```suggestion
Wait for Kiagnose Job to finish.
```

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Results
+When the checkup is finished the checkup ConfigMap is updated with the following results:
```
orelmisan, line 107:
```suggestion
When the Kiagnose Job is finished, the user-supplied ConfigMap is updated with the following results:
```

checkups/kubevirt-vm-latency/README.md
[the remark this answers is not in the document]
orelmisan: In the configuration section its logical to first understand what configuration options there are, and then to apply them.
Here I find it more logical to first fetch the results and then to understand what you see.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Clean up
+```bash
+kubectl delete job -n kiagnose kubevirt-vm-latency-checkup
+kubectl delete config-map -n kubevirt-vm-latency-checkup-config
+```
+Once the checkup is finished it's safe to remove the ClusterRole:
+```bash
+kubectl delete -f ./manifests/clusterroles.yaml
+```
```
orelmisan, line 178: These instructions are relevant only if you want to remove the checkup.
There is no need to do it every time the checkup finishes its execution.

ormergi: It is not mentioned that the ClusterRoles needs to be deleted every time.
Those instructions intention is to clean up after the steps that were described earlier, the way I see it its up to the reader what to delete and when.

checkups/kubevirt-vm-latency/README.md
ormergi: I find using "Kiagnose Job"it a little bit confusing, 
eventually a checkup is containerized app (by definition) when it finish we can check the results, plain and simple.

I think using "checkup ConfigMap" is more accurate, I wanted to call it checkup config ConfigMap by it sounds awkward.

checkups/kubevirt-vm-latency/README.md
ormergi: It will run until the timeout (specified at the checkup ConfigMap).
I think its worth to have this example, it may teach the reader how to work with Kiagnose and this checkup API.

checkups/kubevirt-vm-latency/README.md
ormergi: Yes it looks fine, this is the right path (`README.md` and `docs/` are at the same level).

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,195 @@
+# Kubevirt VM latency checkup
+
+[Kubevirt](https://kubevirt.io/) users are often required to check network connectivity and performance of their virtualized workloads.
+Such checks might be required to do immediately after a cluster with Kubevirt is deployed (as an acceptance test), when new networks are introduced or just as part of a troubleshooting effort.
```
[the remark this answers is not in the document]
ormergi: I think its a matter of taste this rephrasing looks good to me.

checkups/kubevirt-vm-latency/README.md
ormergi: As I mentioned before this field is optional by the API, not mentioning its optional may confuse the reader since its mentioned as optional in other places.

Yes this checkup requires additional permissions, its described at `Permissions` at the very beginning.

How about adding  a note right next to the table, something like:
> Although `spec.clusterRoles` is optional filed, specifying the checkup clusterRole is mandatory for it to work as metioned at <pemissions section>

WDYT?

checkups/kubevirt-vm-latency/README.md
orelmisan: Sorry for the confusion.

checkups/kubevirt-vm-latency/README.md
ormergi: Done

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,193 @@
+# Kubevirt VM latency checkup
+
+[Kubevirt](https://kubevirt.io/) users are often required to check network connectivity and performance of their virtualized workloads.
+Such checks might be required to do immediately after a cluster with Kubevirt is deployed (as an acceptance test), when new networks are introduced or part of a troubleshooting effort.
```
orelmisan, line 4:
```suggestion
Such checks might be required to do immediately after a cluster with Kubevirt is deployed (as an acceptance test), when new networks are introduced or as part of a troubleshooting effort.
```

checkups/kubevirt-vm-latency/README.md
orelmisan: It is optional in a general sense, but in the context of this checkup it is mandatory and has a specific value (`kubevirt-vm-latency-checker`).
If a user does not put this specific value, the checkup won't work.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,193 @@
...
+| `spec.clusterRoles`                                                                | ClusterRole name with the required permission (optional).                                                                    |
+| `network_attachment_definition_namespace`<br/>`network_attachment_definition_name` | `NetworkAttachmentDefinition` object on which <br/> the VMs are connected to and measure network latency.                    |
+| `sample_duration_seconds`                                                          | Network latency measurement sample time (optional).<br/> Default is 5 seconds.                                               |
+| `max_desired_latency_milliseconds`                                                 | Maximal network latency accepted, if the actual latency <br/> is higher the checkup will be considered as failed (optional). |
+| `source_node`<br/>`target_node`                                                    | Two ends in the of the network latency measurement (optional).                                                               |
+
+> **_Note_**:
+>`timeout` should be greater than `sample_duration_seconds`.
+
+### Example:
```
orelmisan, line 46:
```suggestion
### Example
```

checkups/kubevirt-vm-latency/README.md
orelmisan: Please see the [example](https://github.com/kiagnose/kiagnose/blob/61b92c98e6a3c76e5bea1e4398652e22a20b1f2a/checkups/echo/README.md?plain=1#L65) in the echo checkup's README.

checkups/kubevirt-vm-latency/README.md
orelmisan: A user directly runs `kiagnose`.
`user-supplied ConfigMap` is a term that is used throughout our documentation.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,193 @@
+# Kubevirt VM latency checkup
+
+[Kubevirt](https://kubevirt.io/) users are often required to check network connectivity and performance of their virtualized workloads.
+Such checks might be required to do immediately after a cluster with Kubevirt is deployed (as an acceptance test), when new networks are introduced or part of a troubleshooting effort.
+
+Kubevirt VM latency checkup performs network connectivity and latency measurement between two virtual machines over a given network.
+Currently, the measurement is done using standard linux [ping](https://en.wikipedia.org/wiki/Ping_(networking_utility)) utility.
```
orelmisan, line 7:
```suggestion
Currently, the measurement is done using the standard [ping](https://en.wikipedia.org/wiki/Ping_(networking_utility)) utility.
```

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+It can improve the life of cluster administrators by removing the burden of manually testing network connectivity and performance, reduce mistakes and generally save time.
+
+![block diagram](./docs/images/kubevirt-vm-latency-diagram.svg)
+
+## Prerequisites
+- [Kiagnose](../../README.install.md)
+- [Kubevirt](https://kubevirt.io//quickstart_minikube/#deploy-kubevirt)
+- [Multus](https://github.com/k8snetworkplumbingwg/multus-cni#quickstart-installation-guide)
+- Cluster node have one of the desired [CNI](https://www.cni.dev/) plugins installed.
```
orelmisan, line 17: `Cluster nodes have one of the desired [CNI](https://www.cni.dev/) plugins installed.`

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,195 @@
+# Kubevirt VM latency checkup
+
+[Kubevirt](https://kubevirt.io/) users are often required to check network connectivity and performance of their virtualized workloads.
+Such checks might be required to do immediately after a cluster with Kubevirt is deployed (as an acceptance test), when new networks are introduced or just as part of a troubleshooting effort.
+
+Kubevirt VM latency checkup performs network connectivity and latency measurement between two virtual machines over a given network.
+
```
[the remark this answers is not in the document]
ormergi: Done

checkups/kubevirt-vm-latency/README.md
ormergi: Done

checkups/kubevirt-vm-latency/README.md
ormergi: Note that the example that use `kubectl wait job ...` doesn't cover failing jobs, meaning that if the checkup fails it will wait until timeout instead of failing immediately.

orelmisan: I am aware of it.
Because AFAIK the `kubectl wait` doesn't support multiple `--for` options, there is a need to write more than one line of code in order to achieve this.
My intent was to have a one-liner that will do a best effort.

checkups/kubevirt-vm-latency/README.md
ormergi: Done (all over)

checkups/kubevirt-vm-latency/README.md
ormergi: Done

checkups/kubevirt-vm-latency/README.md
RamLavi: both would work in my book. the question whether to keep things more complicated-but-accurate as oppose to simplified-but-less-accurate is niche in this particular case.
I would say, whichever snippet/oneliner you choose - keep things the same everywhere. whatever you choose, use it on both README's

checkups/kubevirt-vm-latency/README.md
ormergi: Thank you both for the feedback, I removed it.

checkups/kubevirt-vm-latency/README.md
ormergi: +1 on keeping things consistent.

@orelmisan this readme (like any other) is the entrypoint for new comers, it should introduce best practices for working with this checkup.
Waiting for the job will work but in my opinion its not the right way, there is an API that provides an indication of when the checkup is done (`completionTimestamp`) I suggest to use it instead of monitoring underlying moving parts (the Job).
This snippet is also used as an example to teach the users how to work with Kiagnose API.

orelmisan: The reason I favor to wait for the kiagnose Job is because its the central place where you can find all the relevant data.
For example, if a user did not configure correctly the two environment variables which links the ConfigMap to the kiagnose Job, there will be no indication that a failure had occurred at the ConfigMap.
At this case (which I think will be pretty common), the suggested snippet will run forever.

orelmisan: Another scenario where I think the suggested snippet will give false results - when a user tries to run with a previously used ConfigMap.
On this scenario - the framework Job exits without reporting.
1. If there is no `completionTimestamp` - the snippet can run forever (or until the other kiagnose instance will finish).
2. If there is `completionTimestamp` - the snippet could exit immediately, before the framework even started.

ormergi: > The reason I favor to wait for the kiagnose Job is because its the central place where you can find all the relevant data.

Actually the ConfigMap have all the information including checkup completion.

Regarding the example you describe, this is basically a missconfiguration scenario (of the Job),
its no different then using `kubectl --wait ..` with wrong parameters (e.g: wrong condition).

About the other scenario, 
(1) I think that this is not likely to happen unless the cluster is not healthy ("reporting" wouldn't work like any other interaction with the cluster API)
(2) Similar behavior could happen with `kubectl wait ..` when the job did not started yet (`containerCreating` error).

At the current state I wouldn't recommend to reuse the `ConfigMap` (Job cannot be reused).
Note that even when there is controller and CRD, one will would probably need to delete the CR (or remove its status) in order to trigger another execution, same with `ConfigMap` it is expected.

Note that this readme intention is to provide some quick start instruction to run this checkup.
It describe the most simple execution flow, it does not consider leftovers from previous execution as its is not expected
in the context of this readme. As long that clean up instructions are preformed there should be no issue.

ormergi: @orelmisan 
> The reason I favor to wait for the kiagnose Job is because its the central place where you can find all the relevant data.

Actually the ConfigMap have all the information including checkup completion.

Regarding the example you describe, this is basically a missconfiguration scenario (of the Job),
its no different then using `kubectl --wait ..` with wrong parameters (e.g: wrong condition).

About the other scenario, 
(1) I think that this is not likely to happen unless the cluster is not healthy ("reporting" wouldn't work like any other interaction with the cluster API)
(2) Similar behavior could happen with `kubectl wait ..` when the job did not started yet (`containerCreating` error).

At the current state I wouldn't recommend to reuse the `ConfigMap` (Job cannot be reused).
Note that even when there is controller and CRD, one will would probably need to delete the CR (or remove its status) in order to trigger another execution, same with `ConfigMap` it is expected.

Note that this readme intention is to provide some quick start instruction to run this checkup.
It describe the most simple execution flow, it does not consider leftovers from previous execution as its is not expected
in the context of this readme. As long that clean up instructions are preformed there should be no issue.

ormergi: For the sake of getting this in, I will change this snippet to use `kubectl wait ...` as suggested and move this discussion to another PR.

Thank you for the feedback I truly appreciate it.

orelmisan (dismissed):
Thank you!

EdDev (commented):
Thanks.
I only had a few small comments, see inline.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,191 @@
...
+## Prerequisites
+- [Kiagnose](../../README.install.md)
+- [Kubevirt](https://kubevirt.io//quickstart_minikube/#deploy-kubevirt)
+- [Multus](https://github.com/k8snetworkplumbingwg/multus-cni#quickstart-installation-guide)
+- Cluster nodes have one of the desired [CNI](https://www.cni.dev/) plugins installed.
+- `NetworkAttachmentDefinition` object to exists.
+
+## Supported CNI plugins
+- [bridge-cni](https://www.cni.dev/plugins/v1.1/main/bridge/)
+- [sriov-cni](https://github.com/k8snetworkplumbingwg/sriov-cni)
```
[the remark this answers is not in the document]
ormergi: These are the supported CNI plugins this checkup supports, if the user pick different CNIs we cannot guarantee it will works.
This is why I did not mentioned the binding method, the way I see it its an implementation detail that may change.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,191 @@
+# Kubevirt VM latency checkup
+
+[Kubevirt](https://kubevirt.io/) users are often required to check network connectivity and performance of their virtualized workloads.
+Such checks might be required to do immediately after a cluster with Kubevirt is deployed (as an acceptance test), when new networks are introduced or as part of a troubleshooting effort.
+
+Kubevirt VM latency checkup performs network connectivity and latency measurement between two virtual machines over a given network.
+Currently, the measurement is done using standard [ping](https://en.wikipedia.org/wiki/Ping_(networking_utility)) utility.
+
+It can improve the life of cluster administrators by removing the burden of manually testing network connectivity and performance, reduce mistakes and generally save time.
```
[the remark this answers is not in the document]
ormergi: Done

checkups/kubevirt-vm-latency/README.md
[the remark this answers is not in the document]
EdDev: Please elaborate on why we can guarantee these two CNI plugins and not others?

The code base does not support this claim, nor I know of Kubevirt limiting the supported CNI types.
AFAIK, we are limited only by what kubevirt is limited.

Regarding the binding, this is what this checkup uses (for good reasons). I am not sure what does "implementation details" means in this regard. This specific checkup checks these bindings, and if the user will setup the secondary networks using a different binding (masquerade, SLIRP, macvtap) on hos VMI/s, this checkup will not reflect his VMI setup connectivity.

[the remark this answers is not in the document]
ormergi: Yes simmilar to Kubevirt, the checkup wont actually limit which CNI can be used,
I added this section in order to mention what the checkup supports, in other words what we tested and can vouch for.

What do you expect from this section to include?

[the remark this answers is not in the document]
EdDev: I hinted what I think should be here by mentioning

>  Maybe you should refer to the binding options from kubevirt and not the CNI.

In other words, we support two binding types (SR-IOV and birdge).
This is reflected also in the code, and I remember I had the same comment there as well.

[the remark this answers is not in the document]
orelmisan: I was also raised [here](https://github.com/kiagnose/kiagnose/pull/12/files#diff-bbe57debe1510ab9407eca2922d081a04d7e6babacc15ecaf1d4275400db7d6dR258).

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/b90808535eae67b223209cfa8a4a16e01615c33d..17e5e07f5d883dd9fabc5218082f90c15afee366): Resolve code review comment https://github.com/kiagnose/kiagnose/pull/83#discussion_r901086737 https://github.com/kiagnose/kiagnose/pull/12/files#diff-bbe57debe1510ab9407eca2922d081a04d7e6babacc15ecaf1d4275400db7d6dR258

checkups/kubevirt-vm-latency/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+It can ease the maintainability effort required from a cluster administrator by removing the burden of manually testing network connectivity and performance, reduce mistakes and generally save time.
+
+![block diagram](./docs/images/kubevirt-vm-latency-diagram.svg)
+
+## Supported network binding method
+The checkup binds the VMs to the given `NetworkAttachmentDefinition` using one of the
+following biding methods:
+- [bridge](https://kubevirt.io/user-guide/virtual_machines/interfaces_and_networks/#bridge)
+- [sriov](hhttps://kubevirt.io/user-guide/virtual_machines/interfaces_and_networks/#sriov)
```
orelmisan, line 17: Thank you for the change @ormergi.
Can you please state that we are using sriov binding for SR-IOV networks and bridge binding for all the rest?

I also suggest to change the heading to `## VM network binding` or something alone these lines.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/17e5e07f5d883dd9fabc5218082f90c15afee366..765c165f8026ead3b18e7dc7d24f327e3f7a1793): address comment https://github.com/kiagnose/kiagnose/pull/83#discussion_r902291069

PR #84, opened 2022-05-23
kiagnose, namespace: Use watcher to wait for deletion

EdDev:
Waiting for the namespace to be deleted was done using polling.
This has now been improved by using a watcher.

To avoid a race where the deletion event arrives before a watcher is
created for the namespace, the watcher is created before the delete
action.


kiagnose/internal/checkup/namespace/namespace.go
```diff
@@ -43,30 +43,48 @@ func Create(client corev1client.CoreV1Interface, namespace *corev1.Namespace) (*
...
+		return fmt.Errorf("failed to wait for Namespace %q deletion: %v", nsName, err)
+	}
 
-	const pollInterval = time.Second * 5
-	return wait.PollImmediate(pollInterval, timeout, func() (bool, error) {
-		_, err := client.Namespaces().Get(context.Background(), nsName, metav1.GetOptions{})
-		namespaceNotFound := errors.IsNotFound(err)
-		if err != nil && !namespaceNotFound {
-			log.Printf("failed to get namespace %q while waiting for it to dispose: %v", nsName, err)
+	if errAction := action(ctx); errAction != nil {
```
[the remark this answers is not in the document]
EdDev: It allows splitting the "wait" and "operation", each doing its thing.
In order to re-use it, some more work is needed to generalize the `wait*` one.

If you'll insist, I can flatten it and collapses everything into `DeleteAndWait`.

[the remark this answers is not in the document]
orelmisan: Do you plan to make a `wait` function in a future PR?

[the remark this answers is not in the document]
orelmisan: Do you mind putting all the code in `DeleteAndWait()`?
If we don't have a near future plan for the separation of "action" and "wait" parts, I think the code will be more readable flat.
Another reasoning is that `waitForNamespaceDeletion()` actually requests the the Namespace deletion.

kiagnose/internal/checkup/namespace/namespace.go
```diff
@@ -43,30 +43,37 @@ func Create(client corev1client.CoreV1Interface, namespace *corev1.Namespace) (*
...
 	}
-	log.Printf("namespace %q successfully deleted", nsName)
-
-	return nil
-}
-
-// waitForDeletion waits until the given namespace is disposed.
-func waitForDeletion(client corev1client.CoreV1Interface, nsName string, timeout time.Duration) error {
-	log.Printf("waiting for namespace %q to dispose", nsName)
+	log.Printf("deleted namespace %q request sent", nsName)
```
orelmisan, line 59:
```suggestion
	log.Printf("delete namespace %q request sent", nsName)
```

orelmisan (dismissed):
Thank you for the changes.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -183,7 +183,7 @@ func TestCheckupTeardownShould(t *testing.T) {
 
 		const (
 			getNamespaceError = "failed to get Namespace"
-			expectedErrMatch  = "timed out"
+			expectedErrMatch  = "deadline exceeded"
```
ormergi, line 186: You can use `context.DeadlineExceeded` typed error.

kiagnose/internal/checkup/namespace/namespace.go
```diff
@@ -43,30 +43,37 @@ func Create(client corev1client.CoreV1Interface, namespace *corev1.Namespace) (*
 
 // DeleteAndWait delete and waits for the given namespace to dispose.
 func DeleteAndWait(client corev1client.CoreV1Interface, nsName string, timeout time.Duration) error {
-	if err := client.Namespaces().Delete(context.Background(), nsName, metav1.DeleteOptions{}); err != nil {
-		return err
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	const NamespaceNameLabel = "kubernetes.io/metadata.name"
+	nsLabel := fmt.Sprintf("%s=%s", NamespaceNameLabel, nsName)
```
ormergi, line 50: You can use `corev1.LabelMetadataName` instead.

EdDev:
change: Rebase

kiagnose/internal/checkup/namespace/namespace.go
```diff
@@ -43,30 +43,37 @@ func Create(client corev1client.CoreV1Interface, namespace *corev1.Namespace) (*
...
+	for {
+		select {
+		case event := <-eventsCh:
+			if event.Type != k8swatch.Deleted {
+				continue
+			}
+			ns, ok := event.Object.(*corev1.Namespace)
+			if !ok {
+				continue
+			}
```
ormergi, line 72: I think this check should be moved before the event type check (before line 66).

ormergi (commented):
Thanks for addressing this it looks good, please see my inline comments.

EdDev:
change: Use a zero teardown timeout.

(we should try and get rid of this time dependent tests)

kiagnose/internal/checkup/checkup_test.go
EdDev: Done

kiagnose/internal/checkup/namespace/namespace.go
EdDev: Done

kiagnose/internal/checkup/namespace/namespace.go
EdDev: Why?

EdDev:
change: Answered review comments about the contant strings and prioritized the `ctx` check over the event existence.

EdDev:
Unfortunately, I am not able to pass the unit test that checks what happens when the context expires.
```
    checkup_test.go:187: 
        	Error Trace:	checkup_test.go:187
        	Error:      	An error is expected but got nil.
        	Test:       	TestCheckupTeardownShould/fail_when_Namespace_wont_dispose_on_time
```

All the time-based tests should be replaced with something that we can control.

EdDev:
change: Add `watch` permission.

kiagnose/internal/checkup/namespace/namespace.go
ormergi: I got it wrong, sorry for the noise

EdDev:
change: Added an injection to the fake client to ignore an operation. It allows the `Delete` verb on the namespace to be ignored, therefore, not to create an event, giving opportunity for the context timeout to kick in.

EdDev:
change: Rebase

kiagnose/internal/checkup/namespace/namespace.go
```diff
@@ -44,30 +44,36 @@ func Create(client kubernetes.Interface, namespace *corev1.Namespace) (*corev1.N
 
 // DeleteAndWait delete and waits for the given namespace to dispose.
 func DeleteAndWait(client kubernetes.Interface, nsName string, timeout time.Duration) error {
-	if err := client.CoreV1().Namespaces().Delete(context.Background(), nsName, metav1.DeleteOptions{}); err != nil {
-		return err
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	nsLabel := fmt.Sprintf("%s=%s", corev1.LabelMetadataName, nsName)
+	nsWatcher, err := client.CoreV1().Namespaces().Watch(ctx, metav1.ListOptions{LabelSelector: nsLabel})
```
qinqon, line 51: There is a "labels" package at apimachinery with some helpers for this operation, what we can use here is "Set" We can use https://pkg.go.dev/k8s.io/apimachinery@v0.24.0/pkg/labels#Set so we don't have to create a expression.

```suggestion
	nsLabel := labels.Set(map[string]string{corev1.LabelMetadataName: nsName})
	nsWatcher, err := client.CoreV1().Namespaces().Watch(ctx, metav1.ListOptions{LabelSelector: nsLabel.String()})
```

kiagnose/internal/checkup/namespace/namespace.go
```diff
@@ -44,30 +44,36 @@ func Create(client kubernetes.Interface, namespace *corev1.Namespace) (*corev1.N
...
+	if err != nil {
+		return fmt.Errorf("failed to create namespace %q watcher: %v", nsName, err)
 	}
```
qinqon, line 54: Defer has to be done just after resource instantiation
```suggestion
	}
	defer nsWatcher.Close();
```

kiagnose/internal/checkup/namespace/namespace.go
```diff
@@ -44,30 +44,36 @@ func Create(client kubernetes.Interface, namespace *corev1.Namespace) (*corev1.N
...
-	log.Printf("waiting for namespace %q to dispose", nsName)
-
-	const pollInterval = time.Second * 5
-	return wait.PollImmediate(pollInterval, timeout, func() (bool, error) {
-		_, err := client.CoreV1().Namespaces().Get(context.Background(), nsName, metav1.GetOptions{})
-		namespaceNotFound := errors.IsNotFound(err)
-		if err != nil && !namespaceNotFound {
-			log.Printf("failed to get namespace %q while waiting for it to dispose: %v", nsName, err)
+	eventsCh := nsWatcher.ResultChan()
+	defer nsWatcher.Stop()
```
qinqon, line 62: Added at previous comment
```suggestion
```

kiagnose/internal/checkup/namespace/namespace.go
```diff
@@ -44,30 +44,36 @@ func Create(client kubernetes.Interface, namespace *corev1.Namespace) (*corev1.N
...
+			ns, ok := event.Object.(*corev1.Namespace)
+			if !ok {
+				continue
+			}
+			log.Printf("Namespace %q deleted", ns.Name)
+			return nil
 		}
-		return namespaceNotFound, nil
-	})
+	}
```
qinqon, line 78: This can be implemented using the helper `watch.Until` here https://pkg.go.dev/k8s.io/client-go/tools/watch#Until so we don't have to handle the channel

It would be something like 

```golang
if _, err := watch.Until(timeout, nsWatcher, func(event watch.Event) (bool, error) {
		        if event.Type != k8swatch.Deleted {
				return false, nil
			}
			ns, ok := event.Object.(*corev1.Namespace)
			if !ok {
				false, nil
			}
			log.Printf("Namespace %q deleted", ns.Name)
			return true, nil
	}); err != nil {
		return err
	}
```

qinqon (changes requested):

kiagnose/internal/checkup/namespace/namespace.go
EdDev: It seems an overkill to me in this case. We are not using any interesting helper method on the `Set`, I think the current simpler (less dependencies) is good enough.

kiagnose/internal/checkup/namespace/namespace.go
EdDev: You mean `Stop()`?

kiagnose/internal/checkup/namespace/namespace.go
EdDev: Nice, I will try it out. Thanks!

kiagnose/internal/checkup/namespace/namespace.go
qinqon: The project is already using apimachinery so the dependency is already there, also I think we can simplify this with 

```
labels.Set{corev1.LabelMetadataName: nsName}
```

EdDev:
[change](https://github.com/kiagnose/kiagnose/compare/fa05c76b90569cae4292763ba972ff82e5f7556c..9dbc9b251d59a5d4d0746783b062e55344791785): Implemented the watching mechanism using the client-go tooling (i.e. `Until`).

EdDev:
The unit tests are still not working... Investigating.

EdDev:
change: There was a need to inject a watch event for the ns deletion (in unit tests).

EdDev:
change: Make linter happy.

kiagnose/internal/checkup/namespace/namespace.go
EdDev: Done

kiagnose/internal/checkup/namespace/namespace.go
EdDev: Done

kiagnose/internal/checkup/namespace/namespace.go
```diff
@@ -43,31 +45,35 @@ func Create(client kubernetes.Interface, namespace *corev1.Namespace) (*corev1.N
...
-		if err != nil && !namespaceNotFound {
-			log.Printf("failed to get namespace %q while waiting for it to dispose: %v", nsName, err)
+	nsLabel := k8slabels.Set{corev1.LabelMetadataName: namespace.Name}
+	w := &cache.ListWatch{
+		WatchFunc: func(options metav1.ListOptions) (k8swatch.Interface, error) {
+			if options.LabelSelector != "" {
+				options.LabelSelector = "," + nsLabel.String()
+			} else {
+				options.LabelSelector = nsLabel.String()
+			}
```
qinqon, line 64: You don't need this here, there is no extra options passed when you consume the ListWatch since it's a very controller context.

qinqon (changes requested):
Jus a nit to simplify a little ListWatch

kiagnose/internal/checkup/namespace/namespace.go
EdDev: This is the only way to inject custom options, so I am not sure I follow your comment here.
The options are required in order to offload to the server side the events we are interested in listening to (we do not want to get events of the whole cluster).

qinqon: I mean the comma part is not needed we are not calling eat with extra options

EdDev:
change: Simplify the resource-version update in tests (no need to use unstructured data).

kiagnose/internal/checkup/namespace/namespace.go
EdDev: Maybe you mean the condition if it is already populated with some data or not?
I think this is safer, as we do not know if this will change or not in the future. I also had the impression that `Until` adds labels, so by the time this is reached, it is already populated with the ones it adds.

qinqon (dismissed):

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -404,6 +435,35 @@ func (c *testsClient) injectResourceManipulationError(verb, resourceName, err st
...
+// injectIgnoreOperation causes the operation described by the verb and resourceName to be ignored.
+// e.g. no events are triggered, datastore is not changed.
+func (c *testsClient) injectIgnoreOperation(verb, resourceName string) {
+	c.PrependReactor(verb, resourceName, func(action clienttesting.Action) (bool, runtime.Object, error) {
+		return true, nil, nil
+	})
+}
+
+func (c *testsClient) injectResourceVersionUpdateOnNamespaceCreation() {
+	c.injectResourceVersionUpdate("create", "namespaces")
```
ormergi, line 447: Here we can use `namespaceResource`.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -404,6 +435,35 @@ func (c *testsClient) injectResourceManipulationError(verb, resourceName, err st
 	c.PrependReactor(verb, resourceName, reactionFn)
 }
 
+// injectIgnoreOperation causes the operation described by the verb and resourceName to be ignored.
+// e.g. no events are triggered, datastore is not changed.
+func (c *testsClient) injectIgnoreOperation(verb, resourceName string) {
```
ormergi, line 440: nit: I would omit `inject`.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -179,15 +192,13 @@ func TestCheckupTeardownShould(t *testing.T) {
...
+		testClient.injectResourceVersionUpdateOnNamespaceCreation()
+
 		assert.NoError(t, testCheckup.Setup())
 
-		const (
-			getNamespaceError = "failed to get Namespace"
-			expectedErrMatch  = "timed out"
-		)
-		testClient.injectGetErrorForResource(namespaceResource, getNamespaceError)
+		testClient.injectIgnoreOperation("delete", "namespaces")
```
ormergi, line 199: Here we can use `namespaceResource`.

ormergi (commented):
Hi Edy, thanks for the PR  it nice to see that client-go provide watchers helpers.

Could you please elaborate on the commit message about why its necessary to 
set the namespace ResourceVersion in tests?

kiagnose/internal/checkup/checkup_test.go
EdDev: Sorry, I do not understand.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -373,7 +404,7 @@ type testsClient struct{ *fake.Clientset }
 // between such cases.
 func newNormalizedFakeClientset() *testsClient {
 	f := fake.NewSimpleClientset(
-		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "o"}},
+		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "o", ResourceVersion: "123"}},
```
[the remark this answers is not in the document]
EdDev: It is not needed, it is a leftover. Thanks.

kiagnose/internal/checkup/checkup_test.go
EdDev: It makes sense (to me) to keep the same language for consistency.

kiagnose/internal/checkup/checkup_test.go
EdDev: Same, I do not understand what you mean.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -404,6 +435,35 @@ func (c *testsClient) injectResourceManipulationError(verb, resourceName, err st
...
+	c.PrependReactor(verb, resourceName, func(action clienttesting.Action) (bool, runtime.Object, error) {
+		return true, nil, nil
+	})
+}
+
+func (c *testsClient) injectResourceVersionUpdateOnNamespaceCreation() {
+	c.injectResourceVersionUpdate("create", "namespaces")
+}
+
+func (c *testsClient) injectResourceVersionUpdate(verb, resourceName string) {
```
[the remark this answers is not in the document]
EdDev: I wanted to keep the `verb` and `resourceName` as input arguments, but the linter does not like it if all are using the same exact values. This is how to overcome it.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -449,6 +509,22 @@ func (c *testsClient) listNamespaces() ([]corev1.Namespace, error) {
 	return nil, nil
 }
 
+func (c *testsClient) injectWatchWithNamespaceDeleteEvent() {
```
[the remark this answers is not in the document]
EdDev: I am having hard time decoding the suggested name. It has "watch", "delete" and "event" with no clear relation between each other. I find the existing option better.

EdDev:
[change](https://github.com/kiagnose/kiagnose/compare/593253fcaf75b7a14dc4a51b5c19a01033b40e27..8929baa5cf2cf1b4e410a8f2d84815795504583f): Removed a leftover from the tests and updated the commit message with a note about the need for the resource version.

kiagnose/internal/checkup/checkup_test.go
ormergi: I meant to use this variable instead https://github.com/kiagnose/kiagnose/blob/main/kiagnose/internal/checkup/checkup_test.go#L46

kiagnose/internal/checkup/checkup_test.go
ormergi: I meant to use this variable instead https://github.com/kiagnose/kiagnose/blob/main/kiagnose/internal/checkup/checkup_test.go#L46

kiagnose/internal/checkup/checkup_test.go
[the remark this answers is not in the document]
ormergi: It seems that it wont be necessary to receive`verb` and `resourceName` parameters, if both functions are merged to one.

kiagnose/internal/checkup/namespace/namespace.go
```diff
@@ -44,32 +46,34 @@ func Create(client kubernetes.Interface, namespace *corev1.Namespace) (*corev1.N
...
+	if err := client.CoreV1().Namespaces().Delete(ctx, namespace.Name, metav1.DeleteOptions{}); err != nil {
 		return err
 	}
-	log.Printf("deleted namespace %q request sent", nsName)
+	log.Printf("delete namespace %q request sent", namespace.Name)
 
-	if err := waitForDeletion(client, nsName, timeout); err != nil {
-		return err
+	nsLabel := k8slabels.Set{corev1.LabelMetadataName: namespace.Name}
+	w := &cache.ListWatch{
```
orelmisan, line 58: What is the difference between `ListWatch` and something like:
```go
client.CoreV1().Namespaces().Watch(ctx, metav1.ListOptions{
  LabelSelector: ...,
})
```

kiagnose/internal/checkup/checkup_test.go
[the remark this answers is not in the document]
EdDev: And still, I want to keep the arguments. I think it is the right thing to do, serving other resources (e.g. jobs) in the very near future. It also allows me to reuse it on other projects.

EdDev:
change: Use the existing const instead of `"namespaces"` resource.

kiagnose/internal/checkup/namespace/namespace.go
orelmisan: What is the difference in functionality?

EdDev: Sorry, I do not understand you.
Maybe you can try reducing the confusion by providing a code snippet you propose here.

What I answered before is that you cannot replace the `w` with the thing the `Watch` returns, they are not the same.
I can add information like: The `Until()` is creating a watcher internally, it does not accept one that is already instantiated.
But I am not sure it helps, it may open more questions on how this thing works.

ormergi (approved):
Thanks for the changes, LGTM.

qinqon (approved):

PR #85, opened 2022-05-23
automation, e2e: Format results

ormergi:
Format the checkup results and make them look human friendly.

### DIFF
--- a/checkups/kubevirt-vm-latency/automation/e2e.sh
+++ b/checkups/kubevirt-vm-latency/automation/e2e.sh
@@ -210,5 +210,7 @@ EOF
     echo
     echo "Result:"
     echo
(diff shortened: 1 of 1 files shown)


orelmisan (commented):
Thank you for the PR @ormergi!
Can you please show in the PR description the output before and after this PR?

EdDev:
> @EdDev well I need to rephrase it because using `-o yaml` is human readable sweat_smile
> my intention was to reduce noise from the results.

I find it useful to see the whole object. It does not make it too difficult to read it in its full form and adds the meta data information which may be useful when analyzing the result.
This is really borderline IMO.

orelmisan:
I agree with @EdDev, I also think that `-o yaml` is fine for our needs.
Furthermore, the `sed` command's input is not very readable to the common person.

ormergi:
Closing this PR since its nice to have and there is no agreement to accept it.

PR #88, opened 2022-05-24
Add shellcheck fixes

RamLavi:
This PR is fixing some shellcheck issues on the automation scripts.

### DIFF
--- a/automation/make.sh
+++ b/automation/make.sh
@@ -106,5 +106,5 @@ if [ -n "${OPT_PUSH_CORE_IMAGE}" ]; then
 fi
 
 if [ -n "${OPT_E2E}" ]; then
-    "${SCRIPT_PATH}"/e2e.sh $@
+    "${SCRIPT_PATH}"/e2e.sh "$@"
 fi
--- a/automation/make.sh
+++ b/automation/make.sh
@@ -21,7 +21,7 @@ set -e
 
 ARGCOUNT=$#
 
-SCRIPT_PATH=$(dirname $(realpath -s "$0"))
+SCRIPT_PATH=$(dirname "$(realpath -s "$0")")
 
 CRI=${CRI:-podman}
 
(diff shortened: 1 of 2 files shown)


automation/make.sh
```diff
@@ -102,9 +102,9 @@ fi
 
 if [ -n "${OPT_PUSH_CORE_IMAGE}" ]; then
     echo "Pushing \"${CORE_IMAGE}\"..."
-    ${CRI} push ${CORE_IMAGE}
+    ${CRI} push "${CORE_IMAGE}"
 fi
 
 if [ -n "${OPT_E2E}" ]; then
-    ${SCRIPT_PATH}/e2e.sh $@
+    "${SCRIPT_PATH}"/e2e.sh "$@"
```
EdDev, line 109: > "$@"

This does not make sense to me. And I wander how many other places this may also be like this.

RamLavi:
```
Double quotes around $@ (and similarly, ${array[@]}) prevents globbing and word splitting of individual elements, while still expanding to multiple separate arguments.

Let’s say you have four arguments: baz, foo bar, * and /*/*/*/*

"$@" will expand into exactly that: baz, foo bar, * and /*/*/*/*

$@ will expand into multiple other arguments: baz, foo, bar, file.txt, otherfile.jpg, and (eventually) a list of most files on the system
```
Since we are passing flags like `--this` and `--that` then we do not desire globbing/splitting, then I think this is the right approach.

RamLavi:
> Thank you for this one.
> 
> I'm not sure if this is correct for or cases (see inline comment), but nevertheless, do you think it is worth adding a linter to cover this?

I believe shellcheck is a good helping tool but it doesn't mean it's worth adding it as a check in the ci. 
Like you said, **every** case needs to be examined (for example, do I want this var to be globbed? or splitted?) and often times one would decide to ignore the shellcheck. keeping it as a ci check will make us add multiple "shellcheck ignore" lines which will not be pretty.

But In the case of `make.sh`, `e2e.sh` scripts, the shellcheck indicate issues of good practice, so I would vote to just use it :)

automation/make.sh
EdDev: Thank you for the explanation.
Quoting `$@` is special and not that trivial to me. I did not know it expands with only the individual elements kept unified.

Issue #17, opened 2022-04-17
Generate random names for the checkup Namespace and ClusterRoleBindings


ormergi:
Following the discussion https://github.com/kiagnose/kiagnose/pull/9/commits/e1dbdaaad820b27d167086464344e4e7a3a6990c#r844968232, 
a checkup `Namespace` and `ClusterRoleBindings` should be created with random names, in order to prevent the framework from blocking other checkups to run in case:
- A checkup hangs due to failing teardown.
- Two or more checkups require the exact same ClusterRole, meaning two 
  ClusterRoleBindings will be created with the same name.

orelmisan:
After PRs #89 and #91 were merged, this issue is fixed.

PR #89, opened 2022-05-25
checkup: Randomize Namespace's name

orelmisan:
In order to run multiple checkups simultaneously, it is required to run them in Namespaces with a unique name.
Create checkup namespaces with names in the following format:
`kiagnose-checkup-<random string>`

Partially fixes issue #17:
This PR does not handle name randomization of ClusterRoleBindings, which is also required to run multiple checkups simultaneously. It will be done in a follow-up PR.

~~Depends on PR #101.~~

### DIFF
--- a/kiagnose/internal/checkup/checkup.go
+++ b/kiagnose/internal/checkup/checkup.go
@@ -55,7 +55,7 @@ type Checkup struct {
 }
 
 const (
-	NamespaceName                  = "checkup-workspace"
+	NamespaceName                  = "kiagnose-checkup"
 	ServiceAccountName             = "checkup-sa"
 	ResultsConfigMapName           = "checkup-results"
 	ResultsConfigMapWriterRoleName = "results-configmap-writer"
@@ -65,18 +65,23 @@ const (
 	ResultsConfigMapNameEnvVarNamespace = "RESULT_CONFIGMAP_NAMESPACE"
 )
 
-func New(c kubernetes.Interface, checkupConfig *config.Config) *Checkup {
-	checkupRoles := []*rbacv1.Role{NewConfigMapWriterRole(ResultsConfigMapWriterRoleName, NamespaceName, ResultsConfigMapName)}
+type namer interface {
+	Name(string) string
+}
 
-	subject := newServiceAccountSubject(ServiceAccountName, NamespaceName)
+func New(c kubernetes.Interface, checkupConfig *config.Config, namer namer) *Checkup {
+	nsName := namer.Name(NamespaceName)
+	checkupRoles := []*rbacv1.Role{NewConfigMapWriterRole(ResultsConfigMapWriterRoleName, nsName, ResultsConfigMapName)}
+
+	subject := newServiceAccountSubject(ServiceAccountName, nsName)
 	var checkupRoleBindings []*rbacv1.RoleBinding
 	for _, role := range checkupRoles {
-		checkupRoleBindings = append(checkupRoleBindings, NewRoleBinding(role.Name, NamespaceName, subject))
+		checkupRoleBindings = append(checkupRoleBindings, NewRoleBinding(role.Name, nsName, subject))
 	}
 
 	checkupEnvVars := []corev1.EnvVar{
 		{Name: ResultsConfigMapNameEnvVarName, Value: ResultsConfigMapName},
-		{Name: ResultsConfigMapNameEnvVarNamespace, Value: NamespaceName},
+		{Name: ResultsConfigMapNameEnvVarNamespace, Value: nsName},
 	}
 	checkupEnvVars = append(checkupEnvVars, checkupConfig.EnvVars...)
 
@@ -84,16 +89,16 @@ func New(c kubernetes.Interface, checkupConfig *config.Config) *Checkup {
 	return &Checkup{
 		client:              c,
 		teardownTimeout:     defaultTeardownTimeout,
-		namespace:           NewNamespace(NamespaceName),
-		serviceAccount:      NewServiceAccount(ServiceAccountName, NamespaceName),
-		resultConfigMap:     NewConfigMap(ResultsConfigMapName, NamespaceName),
+		namespace:           NewNamespace(nsName),
+		serviceAccount:      NewServiceAccount(ServiceAccountName, nsName),
+		resultConfigMap:     NewConfigMap(ResultsConfigMapName, nsName),
 		roles:               checkupRoles,
 		roleBindings:        checkupRoleBindings,
 		jobTimeout:          checkupConfig.Timeout,
-		clusterRoleBindings: NewClusterRoleBindings(checkupConfig.ClusterRoles, ServiceAccountName, NamespaceName),
+		clusterRoleBindings: NewClusterRoleBindings(checkupConfig.ClusterRoles, ServiceAccountName, nsName),
 		job: NewCheckupJob(
 			JobName,
-			NamespaceName,
+			nsName,
 			ServiceAccountName,
 			checkupConfig.Image,
 			int64(checkupConfig.Timeout.Seconds()),
--- a/kiagnose/internal/checkup/checkup_test.go
+++ b/kiagnose/internal/checkup/checkup_test.go
@@ -80,15 +80,20 @@ func TestCheckupWith(t *testing.T) {
 	for _, testCase := range checkupCreateTestCases {
 		t.Run(testCase.description, func(t *testing.T) {
 			c := fake.NewSimpleClientset()
-			testCheckup := checkup.New(c, &config.Config{
-				Image:        testImage,
-				Timeout:      testTimeout,
-				EnvVars:      testCase.envVars,
-				ClusterRoles: testCase.clusterRole,
-				Roles:        testCase.roles,
-			})
+			nameGen := nameGeneratorStub{}
+			testCheckup := checkup.New(
+				c,
+				&config.Config{
+					Image:        testImage,
+					Timeout:      testTimeout,
+					EnvVars:      testCase.envVars,
+					ClusterRoles: testCase.clusterRole,
+					Roles:        testCase.roles,
+				},
+				nameGen,
+			)
 
-			checkupNamespaceName := checkup.NamespaceName
+			checkupNamespaceName := nameGen.Name(checkup.NamespaceName)
 
 			assert.NoError(t, testCheckup.Setup())
 			assertNamespaceCreated(t, c, checkupNamespaceName)
@@ -116,13 +121,17 @@ func TestCheckupSetupShouldFailWhen(t *testing.T) {
 			testClient := newNormalizedFakeClientset()
 			expectedErr := fmt.Sprintf("failed to create resource %q object", testCase.resource)
 			testClient.injectCreateErrorForResource(testCase.resource, expectedErr)
-			testCheckup := checkup.New(testClient, &config.Config{
-				Image:        testImage,
-				Timeout:      testTimeout,
-				EnvVars:      testCase.envVars,
-				ClusterRoles: testCase.clusterRole,
-				Roles:        testCase.roles,
-			})
+			testCheckup := checkup.New(
+				testClient,
+				&config.Config{
+					Image:        testImage,
+					Timeout:      testTimeout,
+					EnvVars:      testCase.envVars,
+					ClusterRoles: testCase.clusterRole,
+					Roles:        testCase.roles,
+				},
+				nameGeneratorStub{},
+			)
 
 			assert.ErrorContains(t, testCheckup.Setup(), expectedErr)
 
@@ -137,7 +146,11 @@ func TestCheckupSetupShould(t *testing.T) {
 		expectedErr := fmt.Sprintf("failed to create resource %q object", clusterRoleBindingResource)
 		expectedClusterRoles := newTestClusterRoles()
 		testClient.injectClusterRoleBindingCreateError(expectedClusterRoles[1].Name, expectedErr)
-		testCheckup := checkup.New(testClient, &config.Config{Image: testImage, Timeout: testTimeout, ClusterRoles: expectedClusterRoles})
+		testCheckup := checkup.New(
+			testClient,
+			&config.Config{Image: testImage, Timeout: testTimeout, ClusterRoles: expectedClusterRoles},
+			nameGeneratorStub{},
+		)
 
 		assert.ErrorContains(t, testCheckup.Setup(), expectedErr)
 
@@ -147,7 +160,7 @@ func TestCheckupSetupShould(t *testing.T) {
 
 func TestCheckTeardownShouldSucceed(t *testing.T) {
 	testClient := newNormalizedFakeClientset()
-	testCheckup := checkup.New(testClient, &config.Config{Image: testImage, Timeout: testTimeout})
+	testCheckup := checkup.New(testClient, &config.Config{Image: testImage, Timeout: testTimeout}, nameGeneratorStub{})
 
 	testClient.injectResourceVersionUpdateOnNamespaceCreation()
 	testClient.injectWatchWithNamespaceDeleteEvent()
@@ -159,7 +172,11 @@ func TestCheckTeardownShouldSucceed(t *testing.T) {
 func TestCheckupTeardownShould(t *testing.T) {
 	t.Run("fail when failed to delete ClusterRoleBinding", func(t *testing.T) {
 		testClient := newNormalizedFakeClientset()
-		testCheckup := checkup.New(testClient, &config.Config{Image: testImage, Timeout: testTimeout, ClusterRoles: newTestClusterRoles()})
+		testCheckup := checkup.New(
+			testClient,
+			&config.Config{Image: testImage, Timeout: testTimeout, ClusterRoles: newTestClusterRoles()},
+			nameGeneratorStub{},
+		)
 
 		testClient.injectResourceVersionUpdateOnNamespaceCreation()
 		testClient.injectWatchWithNamespaceDeleteEvent()
@@ -175,7 +192,7 @@ func TestCheckupTeardownShould(t *testing.T) {
 
 	t.Run("fail when failed to delete Namespace", func(t *testing.T) {
 		testClient := newNormalizedFakeClientset()
-		testCheckup := checkup.New(testClient, &config.Config{Image: testImage, Timeout: testTimeout})
+		testCheckup := checkup.New(testClient, &config.Config{Image: testImage, Timeout: testTimeout}, nameGeneratorStub{})
 
 		testClient.injectResourceVersionUpdateOnNamespaceCreation()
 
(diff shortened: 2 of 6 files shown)


kiagnose/internal/checkup/checkup.go
```diff
@@ -28,7 +28,9 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+
```
RamLavi, line 31: why are the `k8s.io` imports grouped apart?

RamLavi (changes requested):
like!, adding an import nit

kiagnose/internal/checkup/checkup.go
orelmisan: This was done in order to emphasize the difference between the `k8s.io/api` repo and `k8s.io/apimachinery` repo.

RamLavi: ah, the reason I'm asking is that I am used to the convention used on kubevirt where there should be 5 import groups:
1. standard go libraries
2. test libraries
3. third part libraries (like k8s.io...)
4. org libraries (for kubevirt, it's all kubevirt related repos)
5. own repo's libraries.

But TBH it is not so meticulously enforced, so I leave it as a suggestion.

RamLavi (approved):
/lgtm

kiagnose/internal/checkup/checkup.go
orelmisan: Thank you for the explanation.

ormergi (commented):
Hi Orel, thanks for addressing this

IMO: commits 1-7 can be merged to single atomic commit that actually bring meaningful change, each commit does stands by its own but its all variables substitutions (until commit 7).
Same goes for commits 8-9.

kiagnose/internal/checkup/checkup.go
```diff
@@ -296,3 +298,9 @@ func concentrateErrors(errs []error) error {
 
 	return errors.New(sb.String())
 }
+
+func randomizeName(name string) string {
+	const randomStringLen = 6
+
+	return fmt.Sprintf("%s-%s", name, rand.String(randomStringLen))
```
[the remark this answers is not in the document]
orelmisan: What is the reasoning behind your proposal?

The randomization code used in this PR is also used by [kubevirt's libvmi](https://github.com/kubevirt/kubevirt/blob/5bf9394089f6790279aad9ac2001a7a230d34ec3/tests/libvmi/vmi.go#L49).

kiagnose/internal/checkup/checkup.go
```diff
@@ -55,7 +57,7 @@ type Checkup struct {
 }
 
 const (
-	NamespaceName                  = "checkup-workspace"
+	NamespaceName                  = "kiagnose-checkup"
```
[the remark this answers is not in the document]
orelmisan: It is still being used by `newTestRoles()` in `checkup_test.go`.
I'll fix it in a follow-up PR.

orelmisan:
> IMO: commits 1-7 can be merged to single atomic commit that actually bring meaningful change, each commit does stands by its own but its all variables substitutions (until commit 7). Same goes for commits 8-9.

I agree that commits 1-7 could be squashed, but I've found it more comprehensible to depict the change gradually.
Same for commits 8-9.

kiagnose/internal/checkup/checkup.go
```diff
@@ -28,7 +28,9 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
```
EdDev, line 33: Please alias it with  `k8s` or `util`.

kiagnose/internal/checkup/checkup.go
EdDev, line 60: Won't this change the documentation/manifests?

EdDev:
> > IMO: commits 1-7 can be merged to single atomic commit that actually bring meaningful change, each commit does stands by its own but its all variables substitutions (until commit 7). Same goes for commits 8-9.
> 
> I agree that commits 1-7 could be squashed, but I've found it more comprehensible to depict the change gradually. Same for commits 8-9.

I tend to agree that this is the same subject, under the same area, so having so many commits may be an overkill.

EdDev (commented):
| This change was done in preparation for the Namespace's name randomization.

This is repeating on each commit, which is too much IMO.

kiagnose/internal/checkup/checkup.go
orelmisan: This is not reflected in the manifests, because they are deployed to the `kiagnose` Namespace.

The documentation is an open PR (#93).

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/024c5f7fb15d352d2b86cc4fb78baa6e89dff96f..31f73839dcfb046cfed7cd6f54288fa8250fe6fc): Squased commits 1-7, and 8-9.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/31f73839dcfb046cfed7cd6f54288fa8250fe6fc..2f7b95d4f33e96852e6fa4df6ba8697a32cf9149): Answered review [comment](https://github.com/kiagnose/kiagnose/pull/89#discussion_r884238686).

kiagnose/internal/checkup/checkup.go
orelmisan: Aliased it `k8srand`

EdDev (approved):
Thank you!

kiagnose/internal/checkup/checkup.go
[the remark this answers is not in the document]
ormergi: The fact that one project solves a problem in one way doesn't mean other projects should follow the same solution :slightly_smiling_face: 
In this case its easy to come up with our own implementation instead of bringing another dependency with code we dont use (one liner vs. 120).

kiagnose/internal/checkup/checkup.go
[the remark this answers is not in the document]
ormergi: I think #91 is the right place.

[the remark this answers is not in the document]
orelmisan: PR #91 deals with ClusterRoleBinding names.
The problem is with a function that creates Role objects.

kiagnose/internal/checkup/checkup.go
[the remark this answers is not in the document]
orelmisan: > The fact that one project solves a problem in one way doesn't mean other projects should follow the same solution slightly_smiling_face 

It is the same behavior like Pod names in a `ReplicaSet` or a random VMI Pod name.
I think that there is value in the same "look and feel".

> In this case its easy to come up with our own implementation instead of bringing another dependency with code we dont use (one liner vs. 120).

Our `vendor` directory has ~1,000,000 lines ;)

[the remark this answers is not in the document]
EdDev: > The simplest thing to do would be appending Unix time to the namespace name

If apimachinery maintainers accept it, we will adopt it as well.
There is a good chance that your proposal is not safe for the long run or even today. The existing library can prove by usage that it is safe, yours requires proof.

This is not a syntax sugar thing that is imported, but a complex logic thing that can cause a lot of problems if you are wrong. I suggest to keep using the apimachinery version.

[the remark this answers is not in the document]
EdDev: > The fact that one project solves a problem in one way doesn't mean other projects should follow the same solution

One project? Maybe it is the other way around.
We have k8s and its satellite projects, then we have kiagnose who will invent a new randomization. You will need to be a bit more convincing on this.

EdDev (commented):
Ok, after looking at both naming randomization for the namespace and cluster role bindings, I find the approach from the test perspective problematic. What I mean but this is, that it is hard to test and complicates the API.

My suggestion is to use a `NameGenerator` object, with a `Name` method that generates a name. A `Namer` interface can be used so in tests a stub with a fixed name can be set.

The only unit tests that this new NG requires is that calling `Name()` twice, does not generate the same name.

WDYT?

ormergi:
> 

@EdDev I think its possible thing to do, from what I understand it requires passing `NameGenerator` object to checkup.New()
that later on can be switched with a stub in tests.
What was the blocker you tackled?

kiagnose/internal/checkup/checkup.go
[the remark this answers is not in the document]
ormergi: @orelmisan 
> Our vendor directory has ~1,000,000 lines ;)

Yep, not optimal, but if we can make avoid adding more to it easily why not? :slightly_smiling_face: 

@EdDev 
> If apimachinery maintainers accept it, we will adopt it as well.
There is a good chance that your proposal is not safe for the long run or even today. The existing library can prove by usage that it is safe, yours requires proof.

I do understand `apimachinery` is proofed by it usage, but I am not sure what kind of proof do you expect.
Also this code is 5 year old, I dont think anyone would bother to touch also my suggestion doesnt seem 
to add value for `apimachinery` implementation, but I can give it a try..

Note that in `apimachinery` implementation uses the standard go `rand` packaged and the randomization seed
that being used is `time.Now().UnixNano()`.
Eventually the results will be derivative of the randomization seed, what I suggest is basically using the string directly.

> One project? Maybe it is the other way around.
We have k8s and its satellite projects, then we have kiagnose who will invent a new randomization. You will need to be a bit more convincing on this.

What I meant is that we should not `always` stick with the exact same solutions we see on other projects
when we can simplify them to our needs.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/2f7b95d4f33e96852e6fa4df6ba8697a32cf9149..ac8d6f2a389232263ec00054819005870d597037): Rebase
Introduced the `namer` interface and the `NameGenerator` struct that implements it.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/ac8d6f2a389232263ec00054819005870d597037..4811fbc3bfc938d8109a08138dd4f8372080ed80): Changed `checkup.NamespaceName` value to `kiagnose-checkup`.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/4811fbc3bfc938d8109a08138dd4f8372080ed80..7a97625a3e3ef0b5406415c3a3813c23b3d28e2c): Generate name using `NameGenerator.Name()` in `nameGeneratorStub.Name()`.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -555,3 +588,25 @@ func assertNoClusterRoleBindingExists(t *testing.T, testClient *testsClient) {
...
+	givenNames map[string]string
+}
+
+func newNameGeneratorStub() nameGeneratorStub {
+	return nameGeneratorStub{
+		givenNames: map[string]string{},
+	}
+}
+
+func (ngs nameGeneratorStub) Name(prefix string) string {
```
EdDev, line 603: Why do we need this complicated logic? Isn't it enough to return a specific pre-defined string? We do not need it to be random in these tests.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/7a97625a3e3ef0b5406415c3a3813c23b3d28e2c..ac0c4572609e80d2ea238f2983accbcc69bbf0f6): Changed `nameGeneratorStub` receiver to an object instead of a pointer.

kiagnose/internal/checkup/namegenerator/namegenerator.go
```diff
@@ -0,0 +1,34 @@
...
+
+	k8srand "k8s.io/apimachinery/pkg/util/rand"
+)
+
+type NameGenerator struct{}
+
+func (ng NameGenerator) Name(prefix string) string {
+	const randomStringLen = 5
+
+	return fmt.Sprintf("%s-%s", prefix, k8srand.String(randomStringLen))
```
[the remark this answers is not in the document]
EdDev: I won't insist. IMO it is a matter of minimalism.
No need to invest in it further.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/ac0c4572609e80d2ea238f2983accbcc69bbf0f6..0f9d2691b4fe540ff7d57e5ca7df852d91d5658c): Removed random name generation from unit tests.

kiagnose/internal/checkup/checkup_test.go
orelmisan: This function now returns: `<input>-12345`

EdDev (approved):

ormergi (commented):
Thanks for the changes, other then the random string generator implementation 
https://github.com/kiagnose/kiagnose/pull/89#discussion_r884635351
LGTM.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/0f9d2691b4fe540ff7d57e5ca7df852d91d5658c..45af7ff838339fb340d1a6681db9c62036954c0e): Rebase

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/45af7ff838339fb340d1a6681db9c62036954c0e..0fd76854918716d4f4c0314f718a91588dd9f02f): Removed two redundant newlines.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -146,19 +161,21 @@ func TestCheckupSetupShould(t *testing.T) {
...
 		testClient.injectWatchWithNamespaceDeleteEvent()
 
 		assert.NoError(t, testCheckup.Setup())
 		assert.NoError(t, testCheckup.Teardown())
 	})
-
 	t.Run("fail when failed to delete ClusterRoleBinding", func(t *testing.T) {
 		testClient := newNormalizedFakeClientset()
-		testCheckup := checkup.New(testClient, &config.Config{Image: testImage, Timeout: testTimeout, ClusterRoles: newTestClusterRoles()})
-
```
EdDev, line 161: Please preserve the newline.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -146,19 +161,21 @@ func TestCheckupSetupShould(t *testing.T) {
...
-		testCheckup := checkup.New(testClient, &config.Config{Image: testImage, Timeout: testTimeout})
+		testCheckup := checkup.New(testClient, &config.Config{Image: testImage, Timeout: testTimeout}, nameGeneratorStub{})
 
 		testClient.injectResourceVersionUpdateOnNamespaceCreation()
 		testClient.injectWatchWithNamespaceDeleteEvent()
 
 		assert.NoError(t, testCheckup.Setup())
 		assert.NoError(t, testCheckup.Teardown())
 	})
-
```
EdDev, line 157: Please preserve the newline.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -170,10 +187,9 @@ func TestCheckupTeardownShould(t *testing.T) {
 		assert.ErrorContains(t, testCheckup.Teardown(), expectedErr)
 		assertNoNamespaceExists(t, testClient)
 	})
-
```
EdDev, line 173: Please preserve the newline.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -185,13 +201,11 @@ func TestCheckupTeardownShould(t *testing.T) {
 		assert.ErrorContains(t, testCheckup.Teardown(), expectedErr)
 		assertNoClusterRoleBindingExists(t, testClient)
 	})
-
```
EdDev, line 188: Please preserve the newline.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -201,11 +215,13 @@ func TestCheckupTeardownShould(t *testing.T) {
 		assert.ErrorContains(t, testCheckup.Teardown(), wait.ErrWaitTimeout.Error())
 		assertNoClusterRoleBindingExists(t, testClient)
 	})
-
```
EdDev, line 204: Please preserve the newline.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -201,11 +215,13 @@ func TestCheckupTeardownShould(t *testing.T) {
...
 	t.Run("fail when ClusterRoleBinding wont dispose on time", func(t *testing.T) {
 		testClient := newNormalizedFakeClientset()
-		testCheckup := checkup.New(testClient, &config.Config{Image: testImage, Timeout: testTimeout, ClusterRoles: newTestClusterRoles()})
-
```
EdDev, line 208: Please preserve the newline.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -219,13 +235,14 @@ func TestCheckupTeardownShould(t *testing.T) {
 		assert.ErrorContains(t, testCheckup.Teardown(), expectedErrMatch)
 		assertNoNamespaceExists(t, testClient)
 	})
-
```
EdDev, line 222: Please preserve the newline.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -219,13 +235,14 @@ func TestCheckupTeardownShould(t *testing.T) {
...
 	t.Run("fail when failed to delete both Namespace and ClusterRoleBindings", func(t *testing.T) {
 		testClient := newNormalizedFakeClientset()
-		testCheckup := checkup.New(testClient, &config.Config{Image: testImage, Timeout: testTimeout, ClusterRoles: newTestClusterRoles()})
-
```
EdDev, line 226: Please preserve the newline.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -219,13 +235,14 @@ func TestCheckupTeardownShould(t *testing.T) {
...
+		testCheckup := checkup.New(
+			testClient,
+			&config.Config{Image: testImage, Timeout: testTimeout, ClusterRoles: newTestClusterRoles()},
+			nameGeneratorStub{},
+		)
 		testClient.injectResourceVersionUpdateOnNamespaceCreation()
-
```
EdDev, line 228: Please preserve the newline.

EdDev (commented):
Some newlines have been removed which I prefer untouched.

orelmisan:
@EdDev The linter complains about `TestCheckupTeardownShould()` being longer than 100 lines.

RamLavi (dismissed):
/lgtm

kiagnose/internal/checkup/checkup_test.go
orelmisan: Done.

kiagnose/internal/checkup/checkup_test.go
orelmisan: Done.

kiagnose/internal/checkup/checkup_test.go
orelmisan: Done.

kiagnose/internal/checkup/checkup_test.go
orelmisan: Done.

kiagnose/internal/checkup/checkup_test.go
orelmisan: Done.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/0fd76854918716d4f4c0314f718a91588dd9f02f..48d926b000209a17f46961360097052aad73be47): Rebased on PR #101.

EdDev (approved):
Thanks

PR #90, opened 2022-05-25
latency, e2e: Add clean-run target

orelmisan:
Add a `--clean-run` target to vm latency checkup `e2e.sh` script.
The target removes the user-supplied Job and ConfigMap objects.

### DIFF
--- a/checkups/kubevirt-vm-latency/automation/e2e.sh
+++ b/checkups/kubevirt-vm-latency/automation/e2e.sh
@@ -35,8 +35,12 @@ CNAO_VERSION=${CNAO_VERSION:-v0.74.0}
 FRAMEWORK_IMAGE="quay.io/kiagnose/kiagnose:devel"
 CHECKUP_IMAGE="quay.io/kiagnose/kubevirt-vm-latency:devel"
 
+KIAGNOSE_NAMESPACE=kiagnose
+KIAGNOSE_JOB=kubevirt-vm-latency-checkup
+VM_LATENCY_CONFIGMAP=kubevirt-vm-latency-checkup
+
 options=$(getopt --options "" \
-    --long deploy-kubevirt,deploy-cnao,deploy-checkup,define-nad,run-tests,help\
+    --long deploy-kubevirt,deploy-cnao,deploy-checkup,define-nad,run-tests,clean-run,help\
     -- "${@}")
 eval set -- "$options"
 while true; do
@@ -56,9 +60,12 @@ while true; do
     --run-tests)
         OPT_RUN_TEST=1
         ;;
+    --clean-run)
+        OPT_CLEAN_RUN=1
+        ;;
     --help)
         set +x
-        echo -n "$0 [--deploy-kubevirt] [--deploy-cnao] [--deploy-checkup] [--define-nad] [--run-tests]"
+        echo -n "$0 [--deploy-kubevirt] [--deploy-cnao] [--deploy-checkup] [--define-nad] [--run-tests] [--clean-run]"
         exit
         ;;
     --)
@@ -160,10 +167,6 @@ if [ -n "${OPT_DEPLOY_CHECKUP}" ]; then
 fi
 
 if [ -n "${OPT_RUN_TEST}" ]; then
-    KIAGNOSE_NAMESPACE=kiagnose
-    KIAGNOSE_JOB=kubevirt-vm-latency-checkup
-    VM_LATENCY_CONFIGMAP=kubevirt-vm-latency-checkup
-
     echo
     echo "Deploy ConfigMap with input data: "
     echo
@@ -224,4 +227,9 @@ EOF
       echo "Kubevirt VM latency checkup failed: ${failureReason}"
       exit 1
     fi
+fi
+
+if [ -n "${OPT_CLEAN_RUN}" ];then
+  ${KUBECTL} delete job ${KIAGNOSE_JOB} -n ${KIAGNOSE_NAMESPACE} --ignore-not-found
+  ${KUBECTL} delete configmap ${VM_LATENCY_CONFIGMAP} -n ${KIAGNOSE_NAMESPACE} --ignore-not-found
 fi


RamLavi:
Not a must, but do we want to add this as an extra step to `--run-tests` as well (around [here](https://github.com/kiagnose/kiagnose/blob/main/checkups/kubevirt-vm-latency/automation/e2e.sh#L166))?

imo it kinda makes sense that before a tests starts then the remnants of the former one are removed.

orelmisan:
> Not a must, but do we want to add this as an extra step to `--run-tests` as well (around [here](https://github.com/kiagnose/kiagnose/blob/main/checkups/kubevirt-vm-latency/automation/e2e.sh#L166))?
> 
> imo it kinda makes sense that before a tests starts then the remnants of the former one are removed. (AFAIU it's not a must because with #89 merged then we won't get cross-test issues.)

PR #89 will not solve this problem.
The user-supplied ConfigMap and Job objects are deployed to the `kiagnose` namespace which exists as long as the kiagnose framework is installed.

For this we can use other measures:
1. Add `ttlSecondsAfterFinished` field to the user-supplied Job - the Job will be automatically deleted after this amount of seconds.
2. Add a cron job to periodically delete ConfigMap objects of completed checkups @EdDev had suggested it in the past.

IMO the cleanup should be manual, so people will not unintentionally delete data (checkup results, logs, etc).

RamLavi (approved):
/lgtm

checkups/kubevirt-vm-latency/automation/e2e.sh
```diff
@@ -218,3 +221,9 @@ EOF
     echo
     ${KUBECTL} get configmap ${VM_LATENCY_CONFIGMAP} -n ${KIAGNOSE_NAMESPACE} -o yaml
 fi
+
+
+if [ -n "${OPT_CLEANUP}" ];then
+  ${KUBECTL} delete job ${KIAGNOSE_JOB} -n ${KIAGNOSE_NAMESPACE}
+  ${KUBECTL} delete configmap ${VM_LATENCY_CONFIGMAP} -n ${KIAGNOSE_NAMESPACE}
```
EdDev, line 228: What about the namespace? I thought that deleting it, will also cause the other stuff to get removed.
I am also missing ClusterRoles, NAD, etc...

orelmisan: The `cleanup` target (as I see it) is supposed to help you quickly re-run tests during local development (without deleting the whole cluster).

Example usage:
```bash
./checkups/kubevirt-vm-latency/automation/make.sh --e2e -- --run-tests
./checkups/kubevirt-vm-latency/automation/make.sh --e2e -- --cleanup
# Optional rebuild and re-deploy checkup / framework image
./checkups/kubevirt-vm-latency/automation/make.sh --e2e -- --run-tests
```

We can clean all the objects created in the cluster, but it will take more time, and require more work to re-run a test.

EdDev: The target name hints it is a cleanup for what other targets in this script do. So if it is not the kiagnose ns, then all the rest.

If the objective is just to run the tests quickly, then there is no need for a target. It is just a step in the test target. But IMO this is not the common case needed. Changing the code of the checkup requires rebuild, reload and the rest.

ormergi: Note that at the current state, the checkup namespace (checkup-workspace) may exist from previous execution (in case e2e.sh is terminated), I think it should also be deleted.

Its seems that it is not the scope of this PR, but it worth to do similar target for the framework.

RamLavi: my 2 cents: the goal IMO of this PR is to be able to run `--run-tests` repeatedly without it getting stuck on leftovers. 
however it seems that removing the `kiagnose` namespace entirely fails when we reran `--run-tests` cause the test expected the `kiagnose` namespace to be there (for the cm for example).

So I would vote that the framework namespace (aka `kiagnose`) stays when cleaning up. It also makes sense, cause the framework doesn't necessarily have to be removed every test run.

checkups/kubevirt-vm-latency/automation/e2e.sh
```diff
@@ -218,3 +221,9 @@ EOF
     echo
     ${KUBECTL} get configmap ${VM_LATENCY_CONFIGMAP} -n ${KIAGNOSE_NAMESPACE} -o yaml
 fi
+
+
+if [ -n "${OPT_CLEANUP}" ];then
+  ${KUBECTL} delete job ${KIAGNOSE_JOB} -n ${KIAGNOSE_NAMESPACE}
```
RamLavi, line 227:
```suggestion
  ${KUBECTL} delete job ${KIAGNOSE_JOB} -n ${KIAGNOSE_NAMESPACE} --ignore-not-found
```

checkups/kubevirt-vm-latency/automation/e2e.sh
RamLavi, line 228:
```suggestion
  ${KUBECTL} delete configmap ${VM_LATENCY_CONFIGMAP} -n ${KIAGNOSE_NAMESPACE} --ignore-not-found
```

RamLavi (changes requested):

checkups/kubevirt-vm-latency/automation/e2e.sh
orelmisan: Done.
Thanks for the suggestion!

checkups/kubevirt-vm-latency/automation/e2e.sh
orelmisan: Done.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/4cc2187e1da43c08bdcd782e0e1dee259f551194..794b805cefbef883d65bb1daf3a903ca4902ecb4): Added `--ignore-not-found` option to `kubectl delete` commands.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/794b805cefbef883d65bb1daf3a903ca4902ecb4..dbafcc258eca51ecf308852528005f36136caf8c): Rebase
Changed target name to `clean-run`.

EdDev (approved):
Thanks!

@ormergi, @RamLavi  please take a look.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/dbafcc258eca51ecf308852528005f36136caf8c..104d9201921b8311c598471a2fda28f8afbd6c05): Rebase
Removed redundant new line.

checkups/kubevirt-vm-latency/automation/e2e.sh
```diff
@@ -225,3 +228,9 @@ EOF
       exit 1
     fi
 fi
+
+
```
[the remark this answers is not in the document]
orelmisan: Done.

RamLavi (approved):
/lgtm

EdDev (approved):
Thanks!

PR #92, opened 2022-05-25
automation: Compile and test with vendor directory

qinqon:
To ensure that the vendor directory is in sync with the go.mod this
change enforece the "-mod=vendor" go flag to use the vendor at build and
also for running tests.

### DIFF
--- a/automation/make.sh
+++ b/automation/make.sh
@@ -34,6 +34,8 @@ CORE_IMAGE="${IMAGE_REGISTRY}/${IMAGE_ORG}/${CORE_IMAGE_NAME}:${CORE_IMAGE_TAG}"
 
 CORE_BINARY_NAME="kiagnose"
 
+export GOFLAGS=-mod=vendor
+
 options=$(getopt --options "" \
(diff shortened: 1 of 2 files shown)


EdDev (dismissed):
Thank you!

automation/make.sh
```diff
@@ -34,6 +34,8 @@ CORE_IMAGE="${IMAGE_REGISTRY}/${IMAGE_ORG}/${CORE_IMAGE_NAME}:${CORE_IMAGE_TAG}"
 
 CORE_BINARY_NAME="kiagnose"
 
+export GOFLAGS=-mod=vendor
```
EdDev, line 37: Don't we need to add it to any existing ones?

EdDev (commented):
Thank you

automation/make.sh
qinqon: You mean the checkups right ?

qinqon: Yes, if vendor directory is living on the project users have to be forced to use it to be sure that they don't push non synced PRs.

EdDev (approved):
Thank you!

PR #96, opened 2022-05-26
vmlatency, automation, e2e: Allow traffic between bridge ports on CI

ormergi:
Following https://github.com/kiagnose/kiagnose/pull/96
Currently kubevirt-vm-latency checkup e2e test fails consistently due to connectivity issues between the VMs.
It seems that on CI environment traffic between bridge ports is blocked causing e2e test to fail 
due to no communication between VMs over the bridge network.

It seems that setting `sysctl -w net.bridge.bridge-nf-call-iptables=0` or adding iptables rule [[1]](https://github.com/kubevirt/cluster-network-addons-operator#configure-bridge-on-node) 
in order to enable traffic between bridge ports doesn't work on Github Workflow environment.

While troubleshooting, it seems that unloading `br_netfilter` removed the restrictions on traffic between bridge 
ports and made the test pass.
Unloading `br_netfilter` kernel module, ensures that the bridge traffic is not inspected (e.g: by iptables)
and thus no packet is dropped.

Thanks to @EdDev for suggesting this.

### DIFF
--- a/.github/workflows/checkup-kubevirt-vm-latency.check.yaml
+++ b/.github/workflows/checkup-kubevirt-vm-latency.check.yaml
@@ -80,6 +80,8 @@ jobs:
       CRI: docker
       KUBEVIRT_USE_EMULATION: true
     steps:
+      - name: Unload the br_netfilter kernel module to remove traffic restriction between bridge ports
+        run:  sudo rmmod br_netfilter
       - name: Check out code
         uses: actions/checkout@v2
       - name: Build kiagnose image
--- a/checkups/kubevirt-vm-latency/automation/e2e.sh
+++ b/checkups/kubevirt-vm-latency/automation/e2e.sh
@@ -216,5 +216,12 @@ EOF
     echo
     echo "Result:"
     echo
-    ${KUBECTL} get configmap ${VM_LATENCY_CONFIGMAP} -n ${KIAGNOSE_NAMESPACE} -o yaml
+    results=$(${KUBECTL} get configmap ${VM_LATENCY_CONFIGMAP} -n ${KIAGNOSE_NAMESPACE} -o yaml)
+    echo "${results}"
+
+    if echo "${results}" | grep 'status.succeeded: "false"'; then
+      failureReason=$(echo ${results} | grep -Po "status.failureReason: \K'.+'")
+      echo "Kubevirt VM latency checkup failed: ${failureReason}"
+      exit 1
+    fi
 fi
--- a/checkups/kubevirt-vm-latency/automation/e2e.sh
+++ b/checkups/kubevirt-vm-latency/automation/e2e.sh
@@ -181,7 +181,7 @@ data:
     kubevirt-vmis-manager
   spec.param.network_attachment_definition_namespace: "default"
   spec.param.network_attachment_definition_name: "bridge-network"
-  spec.param.max_desired_latency_milliseconds: "10"
+  spec.param.max_desired_latency_milliseconds: "100"
   spec.param.sample_duration_seconds: "5"
 EOF
 


checkups/kubevirt-vm-latency/automation/e2e.sh
```diff
@@ -217,4 +217,10 @@ EOF
     echo "Result:"
     echo
     ${KUBECTL} get configmap ${VM_LATENCY_CONFIGMAP} -n ${KIAGNOSE_NAMESPACE} -o yaml
+
+    succeded=$(${KUBECTL} get configmap ${VM_LATENCY_CONFIGMAP} -n ${KIAGNOSE_NAMESPACE} -o jsonpath='{.data}' | \
+      grep -Po "(?<=\"status.succeeded\":\")(false|true)+")
```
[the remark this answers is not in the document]
ormergi: Done

.github/workflows/checkup-kubevirt-vm-latency.check.yaml
```diff
@@ -80,6 +80,8 @@ jobs:
       CRI: docker
       KUBEVIRT_USE_EMULATION: true
     steps:
+      - name: Configure system to allow traffic between bridge ports
+        run:  sudo rmmod br_netfilter
```
orelmisan, line 84: Can you please describe in the commit message what this command does?

checkups/kubevirt-vm-latency/automation/e2e.sh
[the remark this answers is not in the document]
orelmisan: The change was made in the third commit instead of the second.

checkups/kubevirt-vm-latency/automation/e2e.sh
```diff
@@ -217,4 +217,10 @@ EOF
...
+    if [ "${succeded}" == "false" ]; then
```
orelmisan, line 223: Please consider changing this condition to `if [ "${succeded}" != "true" ]; then`.
It could catch more problems this way (different spelling, missing field, etc).

checkups/kubevirt-vm-latency/automation/e2e.sh
```diff
@@ -217,4 +217,10 @@ EOF
...
+      echo "Kubevirt VM latency checkup failed" && exit 1
```
orelmisan: Why is `&&` needed here?

.github/workflows/checkup-kubevirt-vm-latency.check.yaml
ormergi: I didn't add it to the commit message because its pretty standard linux command https://linux.die.net/man/8/rmmod.

checkups/kubevirt-vm-latency/automation/e2e.sh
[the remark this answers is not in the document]
ormergi: Done

checkups/kubevirt-vm-latency/automation/e2e.sh
ormergi: Done

checkups/kubevirt-vm-latency/automation/e2e.sh
ormergi: Just a habit of compact way of writing, I used an `if` clause which isnt compact no I realize its odd.
I moved `exit 1` to new line instead of using `&&`.

ormergi:
@orelmisan @EdDev please take another look.

.github/workflows/checkup-kubevirt-vm-latency.check.yaml
orelmisan: My intention was to describe that this command causes bridge-related traffic no to go through netfilter.

orelmisan:
@ormergi the e2e test fails because of the max allowed latency tolerance.

ormergi:
> @ormergi the e2e test fails because of the max allowed latency tolerance.

Its a proof that this PR works :grin: 

I changed max desired latency to 100ms

.github/workflows/checkup-kubevirt-vm-latency.check.yaml
ormergi: Done, I also updated PR description.

orelmisan: Please consider changing `iptables` to `netfilter` (it is part of the kernel module's name).

ormergi: I guess you meant for the commit message and description, note that `iptables` is mentioned as an example.

EdDev (changes requested):
Thanks.

checkups/kubevirt-vm-latency/automation/e2e.sh
```diff
@@ -217,4 +217,12 @@ EOF
     echo "Result:"
     echo
     ${KUBECTL} get configmap ${VM_LATENCY_CONFIGMAP} -n ${KIAGNOSE_NAMESPACE} -o yaml
+
+    results=$(${KUBECTL} get configmap ${VM_LATENCY_CONFIGMAP} -n ${KIAGNOSE_NAMESPACE} -o jsonpath='{.data}')
```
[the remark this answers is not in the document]
ormergi: Not sure what you mean by multi level script, you mean that there should be another target for it?

This PR initially used `grep`, I was asked to use `jq` instead https://github.com/kiagnose/kiagnose/pull/96#discussion_r882674991.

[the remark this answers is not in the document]
ormergi: Also calling get once more is not too expensive at this case, I can change it to reuse first one if you prefer.

.github/workflows/checkup-kubevirt-vm-latency.check.yaml
```diff
@@ -80,6 +80,8 @@ jobs:
       CRI: docker
       KUBEVIRT_USE_EMULATION: true
     steps:
+      - name: Configure system to allow traffic between bridge ports
```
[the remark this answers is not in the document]
ormergi: Sound good, done.

checkups/kubevirt-vm-latency/automation/e2e.sh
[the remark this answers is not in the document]
EdDev: I meant that you parse the reason out of the output, but this is not needed.
If it fails, we see the full configmap anyway, so there is no need to make it nicer.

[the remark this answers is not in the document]
ormergi: Github Action wont report it as failing job, this is the reason for parsing and exit with retcode 1.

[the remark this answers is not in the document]
EdDev: > This PR initially used grep, I was asked to use jq instead https://github.com/kiagnose/kiagnose/pull/96#discussion_r882674991.

If we are not using it already in some other place, I think we should not add it as a dependency.
We actually have the `kubectl` that can do this partially if you really must.
I would be fine with a grep, but even simpler than what you had before (just grep for the success, anything else is a failure).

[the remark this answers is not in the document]
ormergi: Looking for "succeeded" string only w/o the key is a bit risky, nothing promise that this string will appear only once.

checkups/kubevirt-vm-latency/automation/e2e.sh
```diff
@@ -211,10 +211,16 @@ spec:
               value: ${VM_LATENCY_CONFIGMAP}
 EOF
 
-    ${KUBECTL} wait --for=condition=complete --timeout=10m job.batch/${KIAGNOSE_JOB} -n ${KIAGNOSE_NAMESPACE}
+    results=$(${KUBECTL} get configmap ${VM_LATENCY_CONFIGMAP} -n ${KIAGNOSE_NAMESPACE} -o yaml)
```
RamLavi, line 214: hmm, why is this line removed? don't we want to wait until the KIAGNOSE_JOB finishes the job (pun intended)?

checkups/kubevirt-vm-latency/automation/e2e.sh
```diff
@@ -211,10 +211,16 @@ spec:
               value: ${VM_LATENCY_CONFIGMAP}
```
RamLavi, line 211: commit message is not accurate, cause the checkup can fail if it takes more than the timeout time. I would change it to something like this
```
Currently there is no assertion on the checkup results
and the e2e check job is not checked for actual test outcome.
```

checkups/kubevirt-vm-latency/automation/e2e.sh
RamLavi: Also, I would prefer printing in json format, just so I could use jq to get the results outcome

ormergi: It was a mistake, sorry for the noise.
Previous iteration was with `jq` I was asked to use `grep` https://github.com/kiagnose/kiagnose/pull/96#discussion_r884683719.

checkups/kubevirt-vm-latency/automation/e2e.sh
ormergi: > commit message is not accurate, cause the checkup can fail if it takes more than the timeout time. I would change it to something like this

Right, but even then the e2e job would report it succeeded.

EdDev (approved):
Thanks!

PR #94, opened 2022-05-26
kubevirt-vm-latency: Support bridge networks

ormergi:
Currently the checkup supports SR-IOV networks only, it defines the VMs interface with SR-IOV binding method 
regardless of the `NetworkAttachmentDefinition` (NAD) spec.

By supporting bridge network it will enable running the checkup on environment without SR-IOV hardware
such as CI and dev environments.

With this PR changes, the checkup will detect which CNI plugin is being used by the given NAD and define the 
VMs interfaces accordingly.
If the bridge CNI is being used the VMs interfaces will be defined with bridge binding method.

PR #97, opened 2022-05-26
vmlatency, checkup, preflight: Verify NetworkAttachmentDefinition exists

ormergi:
Currently the `NetworkAttachmentDefinition` (NAD) namespace and name are passed
to the checkup, w/o verifying it is exists.
In case the NAD does not exists, the checkup will hang until the timeout, because Kubevirt does
not validate that the NAD exists ending up with VMs that wont schedule.

With this PR changes the checkup will verify that before starting the checkup setup,
if the NAD does not exist it will fail early and wont hang until the timeout.
The checkup ClusterRoles manifest is updated with the required permissions to fetch the NAD.

Note that `go.mod` file is updated, but no new dependencies were added.

### DIFF
--- a/checkups/kubevirt-vm-latency/automation/e2e.sh
+++ b/checkups/kubevirt-vm-latency/automation/e2e.sh
@@ -178,7 +178,7 @@ data:
   spec.image: ${CHECKUP_IMAGE}
   spec.timeout: 10m
   spec.clusterRoles: |
-    kubevirt-vmis-manager
+    kubevirt-vm-latency-checker
   spec.param.network_attachment_definition_namespace: "default"
   spec.param.network_attachment_definition_name: "bridge-network"
   spec.param.max_desired_latency_milliseconds: "10"
--- a/checkups/kubevirt-vm-latency/go.mod
+++ b/checkups/kubevirt-vm-latency/go.mod
@@ -4,8 +4,11 @@ go 1.17
 
 require (
 	github.com/google/goexpect v0.0.0-20210430020637-ab937bf7fd6f
+	github.com/k8snetworkplumbingwg/network-attachment-definition-client v0.0.0-20191119172530-79f836b90111
 	github.com/stretchr/testify v1.7.1
+	google.golang.org/grpc v1.40.0
 	k8s.io/api v0.23.5
+	k8s.io/client-go v12.0.0+incompatible
 	kubevirt.io/api v0.0.0-20220430221853-33880526e414
 	sigs.k8s.io/yaml v1.3.0
 )
@@ -36,7 +39,6 @@ require (
 	github.com/gorilla/websocket v1.5.0 // indirect
 	github.com/imdario/mergo v0.3.10 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
-	github.com/k8snetworkplumbingwg/network-attachment-definition-client v0.0.0-20191119172530-79f836b90111 // indirect
 	github.com/kubernetes-csi/external-snapshotter/client/v4 v4.2.0 // indirect
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.2 // indirect
@@ -55,13 +57,11 @@ require (
 	golang.org/x/time v0.0.0-20210723032227-1f47c861a9ac // indirect
 	google.golang.org/appengine v1.6.7 // indirect
 	google.golang.org/genproto v0.0.0-20210831024726-fe130286e0e2 // indirect
-	google.golang.org/grpc v1.40.0 // indirect
 	google.golang.org/protobuf v1.27.1 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b // indirect
 	k8s.io/apiextensions-apiserver v0.23.1 // indirect
-	k8s.io/client-go v12.0.0+incompatible // indirect
 	k8s.io/klog/v2 v2.40.1 // indirect
 	k8s.io/kube-openapi v0.0.0-20220124234850-424119656bbf // indirect
 	k8s.io/utils v0.0.0-20211116205334-6203023598ed // indirect
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
@@ -62,6 +62,13 @@ func New(c vmi.KubevirtVmisClient, namespace string, params config.CheckupParame
 }
 
 func (c *checkup) Preflight() error {
+	if _, err := c.client.GetNetworkAttachmentDefinition(
+		c.params.NetworkAttachmentDefinitionNamespace,
+		c.params.NetworkAttachmentDefinitionName,
+	); err != nil {
+		return fmt.Errorf("preflight: %v", err)
+	}
+
 	return nil
 }
 
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
@@ -32,6 +32,8 @@ import (
 	kvcorev1 "kubevirt.io/api/core/v1"
 	"kubevirt.io/client-go/kubecli"
 
+	netattdefv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+
 	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/vmlatency/internal/checkup"
 	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/vmlatency/internal/config"
 )
@@ -42,6 +44,18 @@ const (
 	testSampleDurationSeconds = 1
 	testTimeout               = time.Nanosecond
 )
+
+func TestCheckupPreflightShouldFailWhenNetAttachDefDoesNotExist(t *testing.T) {
+	expectedError := errors.New("get netAttachDef test error")
+	testCheckup := checkup.New(
+		&clientStub{failGetNetAttachDef: expectedError},
+		testNamespace,
+		newTestsCheckupParameters(),
+		&checkerStub{},
+	)
+
+	assert.ErrorContains(t, testCheckup.Preflight(), expectedError.Error())
+}
 
 func TestCheckupSetupShouldFailWhen(t *testing.T) {
 	t.Run("failed to create a VM", func(t *testing.T) {
@@ -143,10 +157,12 @@ func newTestsCheckupParameters() config.CheckupParameters {
 }
 
 type clientStub struct {
-	returnVmi     *kvcorev1.VirtualMachineInstance
-	failGetVmi    error
-	failCreateVmi error
-	failDeleteVmi error
+	returnVmi *kvcorev1.VirtualMachineInstance
+
+	failGetNetAttachDef error
+	failGetVmi          error
+	failCreateVmi       error
+	failDeleteVmi       error
 }
 
 func (c clientStub) GetVirtualMachineInstance(_, _ string) (*kvcorev1.VirtualMachineInstance, error) {
@@ -163,6 +179,10 @@ func (c clientStub) DeleteVirtualMachineInstance(_, _ string) error {
 
 func (c clientStub) SerialConsole(_, _ string, _ time.Duration) (kubecli.StreamInterface, error) {
 	return nil, nil
+}
+
+func (c clientStub) GetNetworkAttachmentDefinition(_, _ string) (*netattdefv1.NetworkAttachmentDefinition, error) {
+	return nil, c.failGetNetAttachDef
 }
 
 type checkerStub struct {
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
@@ -24,6 +24,8 @@ import (
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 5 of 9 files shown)


orelmisan (commented):

As I see it, there are multiple changes made in a single commit:
1. Changing the ClusterRole name to `kubevirt-vm-latency-checker`.
2. Addition of a permission to get NetworkAttachmentDefinition objects.
3. Implementation of a Pre-flight check.

Can you please make these changes gradually over several commits?

checkups/kubevirt-vm-latency/automation/e2e.sh
```diff
@@ -178,7 +178,7 @@ data:
   spec.image: ${CHECKUP_IMAGE}
   spec.timeout: 10m
   spec.clusterRoles: |
-    kubevirt-vmis-manager
+    kubevirt-vm-latency-checker
```
orelmisan, line 181: This change should also occur at `checkups/kubevirt-vm-latency/manifests/kubevirt-vm-latency-checkup.yaml`.

checkups/kubevirt-vm-latency/go.mod
```diff
@@ -4,7 +4,9 @@ go 1.17
 
 require (
 	github.com/google/goexpect v0.0.0-20210430020637-ab937bf7fd6f
+	github.com/k8snetworkplumbingwg/network-attachment-definition-client v0.0.0-20191119172530-79f836b90111
```
orelmisan, line 7: Is there a change required to the `vendor` directory following these changes?

ormergi: No change were required, note that those dependencies were specified under `require` with "indirect" comment, meaning there is a dependency that uses it.
For example: github.com/k8snetworkplumbingwg/network-attachment-definition-client` is a dependency of Kubevirt client-go uses it.

checkups/kubevirt-vm-latency/automation/e2e.sh
ormergi: Done

ormergi:
> Thank you for the PR @ormergi.
> 
> As I see it, there are multiple changes made in a single commit:
> 
> 1. Changing the ClusterRole name to `kubevirt-vm-latency-checker`.
> 2. Addition of a permission to get NetworkAttachmentDefinition objects.
> 3. Implementation of a Pre-flight check.
> 
> Can you please make these changes gradually over several commits?

Note that 1-3 are all part of the same commit because each change dependent on the other (i,e:  (3) requires (2), (1) wont make sense w/o (3)).

orelmisan:
> > Thank you for the PR @ormergi.
> > As I see it, there are multiple changes made in a single commit:
> > 
> > 1. Changing the ClusterRole name to `kubevirt-vm-latency-checker`.
> > 2. Addition of a permission to get NetworkAttachmentDefinition objects.
> > 3. Implementation of a Pre-flight check.
> > 
> > Can you please make these changes gradually over several commits?
> 
> Note that 1-3 are all part of the same commit because each change dependent on the other (i,e: (3) requires (2), (1) wont make sense w/o (3)).

I argue that the ClusterRole rename to `kubevirt-vm-latency-checker` does stand on its own.

ormergi:
> I argue that the ClusterRole rename to `kubevirt-vm-latency-checker` does stand on its own.

Do you insist of extracting the rename changes to new commit?

orelmisan:
> > I argue that the ClusterRole rename to `kubevirt-vm-latency-checker` does stand on its own.
> 
> Do you insist of moving the rename changes to new commit?

At least write the reason for this change it in the commit message if you don't want to split it.

checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
```diff
@@ -70,3 +73,8 @@ func (c *Client) DeleteVirtualMachineInstance(namespace, name string) error {
 func (c *Client) SerialConsole(namespace, vmiName string, timeout time.Duration) (kubecli.StreamInterface, error) {
 	return c.KubevirtClient.VirtualMachineInstance(namespace).SerialConsole(vmiName, &kubecli.SerialConsoleOptions{ConnectionTimeout: timeout})
 }
+
+func (c *Client) GetNetAttachDef(namespace, name string) (*cncfcniv1.NetworkAttachmentDefinition, error) {
+	return c.KubevirtClient.NetworkClient().K8sCniCncfIoV1().NetworkAttachmentDefinitions(namespace).
```
EdDev, line 78: Ohh my, you are telling me that kubevirt client is exposing the NAD client?
I prefer not to depend on this, can we have a direct access to the client?

EdDev (changes requested):
Thanks.

checkups/kubevirt-vm-latency/go.mod
orelmisan: Thank you for the explanation.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
```diff
@@ -32,6 +32,8 @@ import (
 	kvcorev1 "kubevirt.io/api/core/v1"
 	"kubevirt.io/client-go/kubecli"
 
+	cncfcniv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
```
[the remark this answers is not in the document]
ormergi: Done

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
```diff
@@ -43,6 +45,18 @@ const (
 	testTimeout               = time.Nanosecond
 )
 
+func TestCheckupPreflightShouldFailWhenFailToGetNetAttachDef(t *testing.T) {
```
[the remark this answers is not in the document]
ormergi: Looking at this again, I understand why it is confusing :sweat_smile:.
What I wanted to express is "checkup preflight should fail when it fails to get the net-attach-def".
I will change it to what you suggest it sounds good.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
```diff
@@ -143,10 +157,11 @@ func newTestsCheckupParameters() config.CheckupParameters {
 }
 
 type clientStub struct {
-	returnVmi     *kvcorev1.VirtualMachineInstance
-	failGetVmi    error
-	failCreateVmi error
-	failDeleteVmi error
+	failGetNetAttachDef error
```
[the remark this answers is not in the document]
ormergi: Done

checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
```diff
@@ -24,8 +24,11 @@ import (
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	kvcorev1 "kubevirt.io/api/core/v1"
 	"kubevirt.io/client-go/kubecli"
+
+	cncfcniv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
```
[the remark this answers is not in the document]
ormergi: Looks good, done (all over).

checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
```diff
@@ -70,3 +73,8 @@ func (c *Client) DeleteVirtualMachineInstance(namespace, name string) error {
 func (c *Client) SerialConsole(namespace, vmiName string, timeout time.Duration) (kubecli.StreamInterface, error) {
 	return c.KubevirtClient.VirtualMachineInstance(namespace).SerialConsole(vmiName, &kubecli.SerialConsoleOptions{ConnectionTimeout: timeout})
 }
+
+func (c *Client) GetNetAttachDef(namespace, name string) (*cncfcniv1.NetworkAttachmentDefinition, error) {
```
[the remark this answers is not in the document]
ormergi: Done.

ormergi (commented):
@EdDev thanks for the review, I resolved your comments.
Please take another look.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -62,6 +62,14 @@ func New(c vmi.KubevirtVmisClient, namespace string, params config.CheckupParame
 }
 
 func (c *checkup) Preflight() error {
+	var err error
```
EdDev, line 65: Why do you need this?
Can't you do `if _, err := c.client.GetNetworkAttachmentDefinition(` below?

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -62,6 +62,14 @@ func New(c vmi.KubevirtVmisClient, namespace string, params config.CheckupParame
...
+	if _, err = c.client.GetNetworkAttachmentDefinition(
+		c.params.NetworkAttachmentDefinitionNamespace,
+		c.params.NetworkAttachmentDefinitionName,
+	); err != nil {
```
EdDev, line 69: I think the preflight check should be specific about the missing entity, not if there are communication or access problems.
I think you should fail only in one condition, anything else should be ignored.

checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
```diff
@@ -28,9 +28,15 @@ import (
 
 	kvcorev1 "kubevirt.io/api/core/v1"
 	"kubevirt.io/client-go/kubecli"
+
+	netattdefv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+	netattdefclient "kubevirt.io/client-go/generated/network-attachment-definition-client/clientset/versioned/typed/k8s.cni.cncf.io/v1"
```
EdDev, line 33: I see you still use the kubevirt client.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
ormergi: If there are communication / access issues at such early stage why continue the flow?
It is likely that there will be more issues later on that will cause the checkup to fail.
Failing faster will improve the UX, its one of the reasons for having preflight in the first place.

checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
ormergi: Right, `kubevirt.io/client-go` includes generated version of  `NetAttachDef` client [[1]] (https://github.com/kubevirt/kubevirt/tree/main/staging/src/kubevirt.io/client-go/generated/network-attachment-definition-client/clientset/versioned) [[2]](https://github.com/kubevirt/kubevirt/blob/main/hack/generate.sh#L65).
Its different from using `github.com/k8snetworkplumbingwg/network-attachment-definition-client` as it will bring simmilar files to what Kubevirt client-go includes.

Note that once we switch to controller-runtime client we will not need neither of Kubevirt client-go or the `NetAttachDef` client.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
EdDev: The short answer is: This is not the k8s way.
The long answer is: Accessing the NAD directly is not a must to pass the checkup at this moment. Furthermore, it is not the preflight interest about intermediate disconnection or unavailability of the API server.
When the parsing of the NAD is introduced, it makes sense to fail there if the data is not collected, not in a different place.

checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
EdDev: > Its different from using github.com/k8snetworkplumbingwg/network-attachment-definition-client as it will bring simmilar files to what Kubevirt client-go includes.

I do not understand if this means it is not possible or something else.
If you attempt to do so, will something break?

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
ormergi: Initially this commit was part of #94, I was asked to move it to another PR, now it lake context and we have this discussion that could be obsolete.
With #94 PR changes fetching the NAD is a must.

Think about this scenario, the NAD is not exist and there was an issue with accessing the cluster API server,
the preflight will pass (false positive) and the checkup setup will later on fail due to missing data / NAD not found error.
Its wrong, isnt it?

checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
ormergi: > I do not understand if this means it is not possible or something else.
If you attempt to do so, will something break?

It doesn't work,` kubevirt.io/client-go` includes its own generated version of network-attachment-definition client, 
meaning we cannot use `import "github.com/k8snetworkplumbingwg/network-attachment-definition-client/..."`  because there are missing files (of the clientset itself).

EdDev: It sounds wrong to me, but I cannot dig into this further at this time.
I do not like the fact that we use the client provided by kubevirt, as accessing the NAD should be independent.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
EdDev: And at the same time, it is not that accurate to fail due to a glitch on the API server.
There are a wide range of options to choose from, including retry, fail only on things you care about (I guess that is my suggestion) or just fail on first attempt as you suggested.

I do not know how a different action in a different PR will influence this decision.
But it does influence what fails and when.

I guess there is no one best path here, all have costs.
I am fine keeping it as is, thanks.

EdDev (approved):
Thank you!

EdDev:
@orelmisan , please check if all your concerns have been responded.

orelmisan (approved):
Thank you @ormergi!

PR #99, opened 2022-05-29
config: Add a check if custom param name is empty

orelmisan:
Until this commit, it was possible to receive an empty string as a valid custom parameter name.

Add a check if a custom parameter name is empty, and fail if it is.

### DIFF
--- a/kiagnose/internal/config/config_test.go
+++ b/kiagnose/internal/config/config_test.go
@@ -123,6 +123,8 @@ func TestReadFromConfigMapShouldFail(t *testing.T) {
 		expectedError string
 	}
 
+	const emptyParamName = ""
+
 	failureTestCases := []loadFailureTestCase{
 		{
 			description: "when ConfigMap is already in use",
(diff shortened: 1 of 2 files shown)


kiagnose/internal/config/config_test.go
```diff
@@ -177,6 +177,15 @@ func TestReadFromConfigMapShouldFail(t *testing.T) {
 			configMapData: map[string]string{types.ImageKey: imageName, types.TimeoutKey: timeoutValue, types.RolesKey: "illegal name\n"},
 			expectedError: "role name",
 		},
+		{
+			description: "when param name is empty",
+			configMapData: map[string]string{
+				types.ImageKey:           imageName,
+				types.TimeoutKey:         timeoutValue,
+				types.ParamNameKeyPrefix: "some value",
```
ormergi, line 185: Shouldn't it be an empty string?

ormergi (commented):
Hi Orel, thanks for addressing this.

nit about the commit message, I would rephrase it to something like "Currently the checkup parameters are not validated.."

kiagnose/internal/config/config_test.go
orelmisan: The value here doesn't matter - its the key that matters, and here its just the prefix.
This translates to an empty string key.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/b32e065fbdc50da26225ea7da15f04a27dd9da52..cc2983d754f093b57b0ca863a90aebc521272f88): Edited commit message.

kiagnose/internal/config/config_test.go
EdDev: The confusion indeed exist.
I would recommend making this explicit:
`types.ParamNameKeyPrefix + emtyParamName: ...`

EdDev (commented):
Thanks! Just a small point which will make the unit test clearer.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/cc2983d754f093b57b0ca863a90aebc521272f88..751a2426875f1174db08157aa11ca0a3ab130356): Added `emptyParamName` constant to make the test more explicit.

kiagnose/internal/config/config_test.go
orelmisan: Done.

ormergi (approved):
LGTM

EdDev (approved):
Thank you!

PR #100, opened 2022-05-30
kiagnose, checkup: Use k8s client-go `Until` tool for watching the job

EdDev:
Kiagnose creates a checkup job and waits for it to finish.
The current implementation uses a watcher with a self curated logic to
process the events received from it.

The client-go tooling module has a `watch` package [1] that exposes the
function `Until`, implementing all needed mechanics that are required to
process watchers events.

By using this tool, it is now possible to drop our own limited
implementation.
It is worth noting that `Until` also handles possible connectivity
issues between the client and the server, increasing stability in case
of connectivity flakiness.

The change is following the work done for the namespace deletion [2] but
includes an improvements for early termination:
- In case a DELETED event type is detected, fail the waiting earlier.

[1] https://pkg.go.dev/k8s.io/client-go/tools/watch
[2] https://github.com/kiagnose/kiagnose/pull/84

Note: The early exit from the waiting does not include the ADDED event, as I am unsure if it will not be received as well (given we provide the resource version for the initial creation).

### DIFF
--- a/kiagnose/internal/checkup/checkup_test.go
+++ b/kiagnose/internal/checkup/checkup_test.go
@@ -148,10 +148,8 @@ func TestCheckupSetupShould(t *testing.T) {
 		nameGen := nameGeneratorStub{}
 		secondClusterRoleBindingName := nameGen.Name(expectedClusterRoles[1].Name)
 		testClient.injectClusterRoleBindingCreateError(secondClusterRoleBindingName, expectedErr)
-
-		testCheckup := checkup.New(
-			testClient,
-			&config.Config{Image: testImage, Timeout: testTimeout, ClusterRoles: expectedClusterRoles},
+		testClient.injectResourceVersionUpdateOnJobCreation()
+		testCheckup := checkup.New(testClient, &config.Config{Image: testImage, Timeout: testTimeout, ClusterRoles: expectedClusterRoles},
 			nameGen,
 		)
 
@@ -285,6 +283,7 @@ func TestCheckupRunShouldCreateAJob(t *testing.T) {
 	for _, testCase := range checkupRunTestCases {
 		t.Run(testCase.description, func(t *testing.T) {
 			testClient := newNormalizedFakeClientset()
+			testClient.injectResourceVersionUpdateOnJobCreation()
 			testClient.injectResourceVersionUpdateOnNamespaceCreation()
 			testClient.injectWatchWithNamespaceDeleteEvent()
 
@@ -312,6 +311,7 @@ func TestCheckupRunShouldCreateAJob(t *testing.T) {
 			actualJob, err := testClient.BatchV1().Jobs(checkupNamespaceName).Get(context.Background(), checkup.JobName, metav1.GetOptions{})
 			assert.NoError(t, err)
 
+			actualJob.ResourceVersion = ""
 			assert.Equal(t, actualJob, expectedJob)
 
 			assert.NoError(t, testCheckup.Teardown())
@@ -329,6 +329,7 @@ func TestCheckupRunShouldSucceed(t *testing.T) {
 		t.Run(testCase.description, func(t *testing.T) {
 			testClient := newNormalizedFakeClientset()
 			testClient.injectResourceVersionUpdateOnNamespaceCreation()
+			testClient.injectResourceVersionUpdateOnJobCreation()
 			testClient.injectWatchWithNamespaceDeleteEvent()
 
 			nameGen := nameGeneratorStub{}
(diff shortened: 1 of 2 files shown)


RamLavi:
@EdDev I can't fully understand the commit message. can you rephrase? Notice:
- typo: "It it".
- You are talking about the go-tooling but not linking it.
- You are referencing to another case "It it similar to the one..." - it is not understood unless one reviewed it that case, you need to either link to that "similar case" or make the commit justification understood standalone.

RamLavi (changes requested):
nit

kiagnose/internal/checkup/job/job.go
```diff
@@ -43,42 +45,49 @@ func Create(client kubernetes.Interface, job *batchv1.Job) (*batchv1.Job, error)
...
 
-	finishedJob, err := waitForJob(ctx, jobWatcher)
+	event, err := k8swatchtools.Until(ctx, job.ResourceVersion, w, func(event k8swatch.Event) (bool, error) {
+		j, ok := event.Object.(*batchv1.Job)
+		if !ok {
+			return false, nil
+		}
+
+		switch event.Type {
+		case k8swatch.Deleted:
```
ormergi, line 71: I think its odd that on Deleted event it will exit with `unexpected error` but for other types of event it return silently,
other events are not expected as well isnt it?
Also I think `ctx.Done` should be taken in account.

How about changing this switch to include:
case for `Added` & `Modified` events, case for `ctx.Done` and  return `unexpected event` error at `default`?

qinqon (dismissed):

kiagnose/internal/checkup/job/job.go
```diff
@@ -43,42 +45,49 @@ func Create(client kubernetes.Interface, job *batchv1.Job) (*batchv1.Job, error)
...
+
+	const JobNameLabel = "job-name"
+	jobLabel := k8slabels.Set{JobNameLabel: job.Name}
+	w := &cache.ListWatch{
+		WatchFunc: func(options metav1.ListOptions) (k8swatch.Interface, error) {
+			if options.LabelSelector != "" {
+				options.LabelSelector = "," + jobLabel.String()
+			} else {
+				options.LabelSelector = jobLabel.String()
+			}
```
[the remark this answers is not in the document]
EdDev: Although it is not a must, I find this safer for potential future changes.
it is true that the `Until` function is not passing in any LabelSelector, but this is an implementation detail you rely on based on reading its code. It does pass in other options.

I think we should not rely on these implementation details and assume nothing about if the label selectors are already there or not.

kiagnose/internal/checkup/job/job.go
EdDev: I only optimized for the parts I am sure about, I am not 100% sure about `ADDED`, because the RV we watch from is the one received from the creation operation. Does in include this initial `ADDED` event? I do not know.
I guess a following PR can try to add it, explaining why it is correct.

Regarding the `ctx.Done`, that is not required as the tooling is taking care of it. The whole point of using the `watch` tooling is to offload these mechanics to it.

qinqon (changes requested):
a proposal, is not a must

EdDev:
change: Passed to `Until` a function instead of defining an anonymous one inline (per review request).

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -362,35 +364,35 @@ func TestCheckupRunShouldFailWhen(t *testing.T) {
...
 		testClient := newNormalizedFakeClientset()
-		testClient.injectJobWatchError(expectedErr)
 		testClient.injectResourceVersionUpdateOnNamespaceCreation()
 		testClient.injectWatchWithNamespaceDeleteEvent()
 
 		testCheckup := checkup.New(testClient, &config.Config{Image: testImage, Timeout: testTimeout}, nameGeneratorStub{})
 
 		assert.NoError(t, testCheckup.Setup())
-		assert.ErrorContains(t, testCheckup.Run(), expectedErr)
+		assert.ErrorContains(t, testCheckup.Run(), "initial RV \"\" is not supported")
```
RamLavi, line 374: is there no type of error like this? if possible it is better to use their errors, because it is not guaranteed that the wording will stay the same in future bumps, thus braking this test.

RamLavi (changes requested):
I like this new tool! Added some comments.

EdDev:
@RamLavi , I interpreted silence as ok to take it as is. If you think that point about the test merits a helper, please suggest it in a PR and I will take it.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -312,6 +311,7 @@ func TestCheckupRunShouldCreateAJob(t *testing.T) {
 			actualJob, err := testClient.BatchV1().Jobs(checkupNamespaceName).Get(context.Background(), checkup.JobName, metav1.GetOptions{})
 			assert.NoError(t, err)
 
+			actualJob.ResourceVersion = ""
```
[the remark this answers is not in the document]
RamLavi: I will insist! but I'll put it i another PR :)

kiagnose/internal/checkup/checkup_test.go
RamLavi: if you don't have  a predefined error, then I would let it go. odds are slim they they will change it...

PR #102, opened 2022-05-31
vm-latency, e2e: Increase deploy kubevirt timeout

RamLavi:
Currently kubevirt deployment timeout is set to 2 minutes. This
sometimes fail when running locally.
This PR is aligning the timeout to 5m which is the same as kubevirt repo's own deploy
complete [wait](https://github.com/kubevirt/kubevirt/blob/main/hack/cluster-deploy.sh#L128
).


orelmisan (commented):
Thank you for the PR @RamLavi!
Please consider changing the commit subject and the PR title to something like `latency, e2e: Increase kubevirt deployment timeout`.

PR #106, opened 2022-06-01
latency, manifests: Delete SR-IOV NetworkAttachmentDefinition

orelmisan:
Currently, there is an example SR-IOV NetworkAttachmentDefinition
manifest file residing in the vm-latency-checkup's manifests directory.

The vm-latency checkup should only reference the
NetworkAttachmentDefinition by name, and should only care about
its CNI type.


ormergi (changes requested):
I dont understand why the fact that a NAD is referred to by its name to remove this manifest, please elaborate.

I prefer to keep this manifest so there will be an example for NAD configuration vmlatency checkup support.
Also #83 depends on it.

EdDev:
> I dont understand why the fact that a NAD is referred to by its name to remove this manifest, please elaborate.
> 
> I prefer to keep this manifest so there will be an example for NAD configuration vmlatency checkup support. Also #83 depends on it.

I requested it.
The NAD is not an example of anything related to Kiagnose or the latency checkup. It is something one uses in a test, nothing more.

As I recomended in the previous PR, if you find it useful, place it in some test folder.

I do not want any manifests that are not directly related to Kiagnose to be present, packaged and distributed by us.

EdDev (approved):
If someone wants this for a specific purpose, please introduce it in a PR and reason why we need to maintain it.

PR #112, opened 2022-06-06
echo, manifests: Remove example manifest

orelmisan:
Remove echo checkup example manifest.
Change the automation e2e script to use echo checkup's manifests from stdin.


EdDev (approved):
Thanks!

PR #123, opened 2022-06-22
actions: Push tags to quay.io

qinqon:
When repository is tagged with a semver it need to also tag the
generated containers at quay.io, this changes add a new trigger to push
containers and use the trigger ref to tag the container.

Signed-off-by: Enrique Llorente <ellorent@redhat.com>


orelmisan (approved):
Thank you!

PR #131, opened 2022-07-22
Define requirements for v1

phoracek:
This document aims to formalize the goals of v1 release. It should allow
us to focus on the core requirements, before we expand the feature set.

Signed-off-by: Petr Horáček <phoracek@redhat.com>

### DIFF
--- a/docs/way_towards_v1.md
+++ b/docs/way_towards_v1.md
@@ -1,0 +1,31 @@
+# Way towards v1
+
+This document aims to define acceptance criteria that should guide redesign of
+Kiagnose, driving it to v1.
+
+The goal of the v1 release is to provide a solid core of the framework and
+stable APIs. The goal is not to provide a feature rich tool.
+
+## Personas
+
+1. cluster-admin installing Kiagnose
+2. Vendor writting a checkup
+3. Cluster user running one or more checkups
+
+## User requirements
+
+1. Kiagnose is delivered through OLM and installed by the cluster administrator.
+2. Checkups need to run with the same or lesser rights as the user which
+   triggered them.
+3. As a project owner I would like to automate running several checkups.
+   Therefore I need a clear API to pass parameters to a checkup and collect its
+   output.
+4. As a Vendor I would like to have a clear API that my checkup must adhere to,
+   so it is easy to integrate to the framework.
+
+## Out of v1 scope
+
+1. Kiagnose is not required to cleanup of objects it did not create, e.g. objects
+   created by checkups.
+2. Kiagnose does not need to provide tooling allowing checkups to export
+   artifacts.


docs/way_towards_v1.md
```diff
@@ -0,0 +1,31 @@
...
+## Personas
+
+1. cluster-admin installing Kiagnose
+2. Vendor writting a checkup
+3. Cluster user running one or more checkups
+
+## User requirements
+
+1. Kiagnose is delivered through OLM and installed by the cluster administrator.
+2. Checkups need to run with the same or lesser rights as the user which
```
ormergi, line 18: I do not quite understand who is the user, do you mean the cluster user that creates the checkups or the serviceAccount the checkup runs with?

phoracek: The person who starts a checkup execution. e.g. project-admin who wants to run the latency checkup in their namespace.

AdiZav:
I'm not sure whether you'll regard this as a requirement or a nice to have, but -
When running two concurrent checkup jobs, the failing job will report Failure without giving explicit status. Only oc logs ... mentions that the job is failing due to the configmap being in use. Is it possible to report that in the job's status?
Or a CRD that will replace said jobs.

phoracek:
I added a clarification regarding privilege escalation.

docs/way_towards_v1.md
```diff
@@ -0,0 +1,35 @@
...
```
orelmisan, line 18: I am not aware of a reference in the K8s API to the user or ServiceAccount which created an object.
Thus the next design of Kiagnose may not be able to address this requirement.
There may be a need for a certain operational policy to enforce it.

phoracek: I don't think we need that. Is not checking that the service account is present in the namespace sufficient to satisfy the requirement?

orelmisan: Lets assume a user does not have the permission to create a `Foo` object.
Lets also assume that there is a checkup which creates a `Foo` object as part of its normal operation.

If this user is privileges enough to execute this checkup (by creating a CRD or by other means) - I am not aware of a way for Kiagnose to check that this user has the same privileges as the checkup or more.

So my preliminary thoughts were to not grant this user the required permissions to execute the checkup in the first place.

phoracek: > Lets assume a user does not have the permission to create a `Foo` object. Lets also assume that there is a checkup which creates a `Foo` object as part of its normal operation.

And when the user triggers a checkup, they must pass the name of the service account that should be used for it.

> 
> If this user is privileges enough to execute this checkup (by creating a CRD or by other means) - I am not aware of a way for Kiagnose to check that this user has the same privileges as the checkup or more.

This service account must be in its namespace. Meaning:
1) Kiagnose can easily confirm whether it indeed is available
2) The user could use this service account through a Pod, even without Kiagnose, so we are not exceeding privileges that the user already has.

If checkups are bound to the same RBAC rules as the user is in its namespace, there is no way the checkup can do more than the user. No

> 
> So my preliminary thoughts were to not grant this user the required permissions to execute the checkup in the first place.

If they cannot gain (even though privilege escallation) a right to say create a Deployment, a checkup running in their namespace should not neither, right?

Not sure how this is affecting the requirement 2.

orelmisan: >This service account must be in its (users, checkups) namespace. Meaning:
>1. Kiagnose can easily confirm whether it indeed is available
>2. The user could use this service account through a Pod, even without Kiagnose, so we are not exceeding privileges that the user already has.

1. True, but how does it help?
2. I agree that a user could hijack a token of a ServiceAccount, but the requirement is to verify that a user has at least the privileges of the checkup. 

> If checkups are bound to the same RBAC rules as the user is in its namespace, there is no way the checkup can do more than the user.

A human user and a ServiceAccount could have different privileges in the same namespace.

From the [K8s docs](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/):
```
When you (a human) access the cluster (for example, using kubectl), you are authenticated by the apiserver as a particular User Account (currently this is usually admin, unless your cluster administrator has customized your cluster). Processes in containers inside pods can also contact the apiserver. When they do, they are authenticated as a particular Service Account (for example, default).
```

phoracek: That's not a requirement. We don't have to verify their privileges (in sense that we have to read them and compare them), we just have to make sure they cannot escape the rules.

As you pointed out, hijacking of service account is already a part of Kubernetes: https://kubernetes.io/docs/reference/access-authn-authz/authorization/#privilege-escalation-via-pod-creation. We don't have to be holier than Kubernetes. We just must not make it worse.

Is relying on service accounts present in the namespace making it worse? If not then I believe that this requirement is satisfied. I have added a sub-point in the last commit speaking specifically about the privilege escalation.

phoracek: If I changed the line to "Checkups must not be able to gain higher rights than the user which triggered them. ", would it make more sense?

orelmisan: I understand what you are saying, but I still think the requirement's wording has to change.
As I've previously said, If a user has no permission to create `Foo` but the checkup has, it should be the responsibility of the admin to consider whether (or not) to grant this user a privilege to execute a checkup.
There is an implicit privilege escalation here.

orelmisan: https://www.youtube.com/watch?v=EJfYh-JVbJA

phoracek: I think that it should be the responsibility of the admin to consider whether (or not) to grant this user privileges **the checkup requires to run**.

That way checkup permissions are not a matter of concern for admin. The only thing the administrator cares about is which user can perform which operation in their namespace (or outside of it), and they express it through RBAC.

All the admin needs to do then is to grant a user a right to execute checkups (any checkup they may want to run). Depending on the permissions available to the user, a checkup may or may not be able to successfully finish.

Do you agree?

phoracek: I don't think we have to enforce anything, Kubernetes does that for us.

There is clearly a disconnect between us. Help me syncing by marking any lines bellow that are wrong:

Assumptions:

1. Any user able to run a Pod can gain privilege escalation through any of the service accounts in their namespace
2. Hence user can do anything them or any of the service accounts in their namespace was allowed to
3. User can create service accounts in their namespace
4. User can bind roles to service account, but these roles cannot exceed privileges of the user
5. User can run a Job with service account set to be one of those available in their namespace
6. Superprivileged kiagnose can create Jobs in any namespace in the cluster
7. These Jobs can have their serviceAccount set to any of the service accounts available in the Job's namespace
8. Admin can allow users (all/roles/specific) to CREATE Checkup CRs
9. User can create a Checkup CR referencing service account through its attribute
10. Kiagnose framework can monitor these CRs and then create Jobs in their respective namespaces, with requested service accounts

Given:
* There is a service account created in a Namespace, thus available to all users with access to the namespace
* A user is given the right to create Checkup CR in the namespace

When:
* The user creates the CR
* Kiagnose sees the CR and creates a Job in turn

Then:
* The content of the checkup starts under the requested service account
* The checkup can do any operation given to the service account
* If the checkup tries to do something out of the scope (e.g. delete another namespace), Kubernetes will prevent it from happening
* The checkup have the same amount of rights, or less, than any users with ability to create Pods in the namespace.

orelmisan: > Hence user can do anything them or any of the service accounts in their namespace was allowed to

This is a hack, and not the main expected way of doing things.
Lets put it aside for now.

> There is a service account created in a Namespace, thus available to all users with access to the namespace

An admin can decide that ServiceAccount listing is not permitted.


Lets assume a user can only do what he was allowed to do by its admin.

Everything else you wrote is correct.

Lets assume the ServiceAccount of the checkup is able to create and delete namespaces.
Lets assume an admin had deployed this ServiceAccount and bound it to the ClusterRole.

Now, a user, which does not have the permissions to create or delete namespace, triggers a checkup which can.
And this collides with the requirement:
>  Checkups need to run with the same or lesser rights as the user which triggered them.

orelmisan: This scenario still holds if you run the checkup as a stand-alone application.

phoracek: Excellent. But it would match a requirement "Checkups must not be able to gain higher rights than the user which triggered them could.", do you agree?

orelmisan: Checkups cannot gain more privileges than their ServiceAccount permits (unless the can somehow create and bind roles to themselves which will be poor configuration).
There is no association between what the user can do, and what the checkup could do after its Pod was created.
I suggest to drop this requirement because we cannot address it.

phoracek: Ok

orelmisan: Thank you for the discussion :)

phoracek:
@orelmisan I replaced requirement 2 with "2. Privileges of a checkup must be bound to a service account.". Let me know if you have any other concerns, or if you can think of a better working to make this a user requirement and not an implementation detail.

orelmisan:
> @orelmisan I replaced requirement 2 with "2. Privileges of a checkup must be bound to a service account.". Let me know if you have any other concerns, or if you can think of a better working to make this a user requirement and not an implementation detail.

I think its an implementation detail.

docs/way_towards_v1.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Personas
+
+1. cluster-admin installing Kiagnose
+2. Vendor writting a checkup
+3. Cluster user running one or more checkups
+
+## User requirements
+
+1. Kiagnose is delivered through OLM and installed by the cluster administrator.
```
orelmisan, line 17:
```suggestion
1. If the proposed solution will contain an operator, it is required to be delivered through OLM and installed by the cluster administrator.
```

phoracek:
> > @orelmisan I replaced requirement 2 with "2. Privileges of a checkup must be bound to a service account.". Let me know if you have any other concerns, or if you can think of a better working to make this a user requirement and not an implementation detail.
> 
> I think its an implementation detail.

If there is no restriction, nothing stops you from implementing it so every checkup gets a temporary super-privileged service account. I would like to express that the rights are bound to something. If you have a better suggestion than my original line, please share.

orelmisan:
> > > @orelmisan I replaced requirement 2 with "2. Privileges of a checkup must be bound to a service account.". Let me know if you have any other concerns, or if you can think of a better working to make this a user requirement and not an implementation detail.
> > 
> > 
> > I think its an implementation detail.
> 
> If there is no restriction, nothing stops you from implementing it so every checkup gets a temporary super-privileged service account. I would like to express that the rights are bound to something. If you have a better suggestion than my original line, please share.

What about "Every checkup shall execute with the minimal set of permissions it requires for its operation"?

phoracek:
That could mean we can have unprivileged user executing super-privileged setup, that's not desirable.

orelmisan:
> That could mean we can have unprivileged user executing super-privileged checkup, that's not desirable.

It can happen with any application, not just a checkup.
It is up to the admin to give permissions to trust-worthy people.

phoracek:
>  It is up to the admin to give permissions to trust-worthy people.

Exactly. Admin gives permissions to people and members of namespaces through RBAC. Kiagnose must not bypass this by introducing more permissions, not even temporarily.

By enforcing this, admins can be fearless allowing users to run any arbitrary checkups they find on the internet, since they would be bound to the same rules as everything else. If we don't have this embedded in the solution, admins would need to audit every checkup somebody may want to run, that's unrealistic.

phoracek:
> I'm not sure whether you'll regard this as a requirement or a nice to have, but - When running two concurrent checkup jobs, the failing job will report Failure without giving explicit status. Only oc logs ... mentions that the job is failing due to the configmap being in use. Is it possible to report that in the job's status? Or a CRD that will replace said jobs.

With v1 there should be a custom CR representing each checkup execution. To every run would have its own place to report the status in. Logs would be still only available through the underlying Pod. Not sure I fully follow the request though, are you asking for concurrent checkup executions>

docs/way_towards_v1.md
```diff
@@ -0,0 +1,50 @@
...
+
+1. Kiagnose is not required to cleanup of objects it did not create, e.g. objects
+   created by checkups.
+2. Kiagnose does not need to provide tooling allowing checkups to export
+   artifacts.
+
+## Project-admin user stories
+
+ * As a cluster admin,
+   I don't want to be bothered by developers wanting to run a test of their application.
```
orelmisan, line 43: Could you please move this part to a section of its own?
I think more details should be added like:
I am not willing to install CRDs or custom control plane components for checkups controlled by project admins.

orelmisan: *for individual checkups

docs/way_towards_v1.md
```diff
@@ -0,0 +1,51 @@
...
+
+## User requirements
+
+1. Nothing but `kubectl` should be required on the client-side to interact with
+   checkups.
+2. Kiagnose must not create a service account or bind any roles. It must rely on
+   existing Kubernetes autorization mechanisms, and on present RBAC
+   configuration.
+3. Checkup must be initiated in the namespace it was requested from.
+4. If there is a need for a long-running controller, it is delivered through OLM
```
EdDev, line 23: Can we remove it now?

docs/way_towards_v1.md
```diff
@@ -0,0 +1,51 @@
...
+   and can be installed by the cluster administrator.
+5. A project-admin can run a checkup without any prior work done by
+   cluster-admin (with the exception described in the requirement 4).
```
EdDev, line 26: Can we remove the exception?

docs/way_towards_v1.md
EdDev: I am not sure if we should explicitly say what a cluster-admin should not be bothered with, as that opens up the option of listing many other things. But this is not that critical to me.

docs/way_towards_v1.md
```diff
@@ -0,0 +1,51 @@
...
+1. Kiagnose is not required to cleanup of objects it did not create, e.g. objects
+   created by checkups.
+2. Kiagnose does not need to provide tooling allowing checkups to export
+   artifacts.
+
+## Project-admin user stories
+
+ * As a cluster admin,
+   I don't want to be bothered by developers wanting to run a test of their application.
+ * As a developer (project admin),
```
EdDev, line 45: A "developer" is not a persona here. Can you change it to a project user?

docs/way_towards_v1.md
```diff
@@ -0,0 +1,51 @@
...
+Kiagnose, driving it to v1.
+
+The goal of the v1 release is to provide a solid core of the framework and
+stable APIs. The goal is not to provide a feature rich tool.
+
+## Personas
+
+1. cluster-admin installing Kiagnose
+2. Vendor writting a checkup
+3. Cluster user running one or more checkups
```
EdDev, line 13: Should this be a project user?

docs/way_towards_v1.md
```diff
@@ -0,0 +1,51 @@
...
+
+This document aims to define acceptance criteria that should guide redesign of
+Kiagnose, driving it to v1.
+
+The goal of the v1 release is to provide a solid core of the framework and
+stable APIs. The goal is not to provide a feature rich tool.
+
+## Personas
+
+1. cluster-admin installing Kiagnose
```
EdDev, line 11: Do we need this now? At the moment, we do not need the cluster-admin intervention.

EdDev (changes requested):
Now that we released 0.2.1, it is time to revisit this high level requirement.

docs/way_towards_v1.md
phoracek: Done

docs/way_towards_v1.md
phoracek: Done

docs/way_towards_v1.md
phoracek: Done

docs/way_towards_v1.md
phoracek: Removed

EdDev (commented):
Thank you @phoracek .

Before taking this in, could you please squash the whole thing?

EdDev (commented):
One more thing: Although we have git to give us dates, it will be nice to add a date on this document.

docs/way_towards_v1.md
```diff
@@ -0,0 +1,53 @@
...
+   artifacts.
+
+## Project-admin user stories
+
+ * As a cluster admin,
+   I don't want to be bothered by project admins wanting to run a test of their
+   application, that could be executed with their current privileges.
+ * As a cluster admin,
+   I'm only willing to deal with native Kubernetes resources,
+   I don't want to be bothered with any Kiagnose-specific resources.
```
orelmisan, line 46: These user stories talk about the cluster-admin.
Please consider moving them to a section of their own.

docs/way_towards_v1.md
```diff
@@ -0,0 +1,53 @@
...
+2. Kiagnose must not create a service account or bind any roles. It must rely on
+   existing Kubernetes autorization mechanisms, and on present RBAC
+   configuration.
+3. Checkup must be initiated in the namespace it was requested from.
+4. A project-admin can run a checkup without any prior work done by
+   cluster-admin.
+5. As a project owner I would like to automate running several checkups.
+   Therefore I need a clear API to pass parameters to a checkup and collect its
+   output. Note: To operate over heterogenous set of checkups and expose them as
+   a single resource, they need to share the same API group and kind.
```
orelmisan, line 28: Please consider moving this paragraph under `## Project-admin user stories`.

docs/way_towards_v1.md
```diff
@@ -0,0 +1,53 @@
...
+6. As a Vendor I would like to have a clear API that my checkup must adhere to,
+   so it is easy to integrate to the framework.
```
orelmisan, line 30: Please consider moving this paragraph under a `vendor user stories` section.

docs/way_towards_v1.md
```diff
@@ -0,0 +1,53 @@
...
+## User requirements
+
+1. Nothing but `kubectl` should be required on the client-side to interact with
+   checkups.
+2. Kiagnose must not create a service account or bind any roles. It must rely on
+   existing Kubernetes autorization mechanisms, and on present RBAC
+   configuration.
+3. Checkup must be initiated in the namespace it was requested from.
+4. A project-admin can run a checkup without any prior work done by
+   cluster-admin.
```
orelmisan, line 24: Please consider moving this paragraph under `## Project-admin user stories`.

EdDev: Currently all are under the same general "user" and it does not affect readability.
I think it is fine as is (although I do not really mind if it is moved).

docs/way_towards_v1.md
EdDev: same

docs/way_towards_v1.md
EdDev: same

docs/way_towards_v1.md
EdDev: I think the "Project-admin" on line 39 should be dropped, making the user as part of the individual story.

docs/way_towards_v1.md
EdDev: And these are requirements, not user stories.

orelmisan: This specific sentence is indeed a requirement.
Please ignore this comment.

RamLavi (dismissed):
/lgtm

phoracek:
Squashed. Let me know if there is anything left to resolve, the way I understand it, it's all answered or not relevant anymore.

orelmisan (approved):
Thank you @phoracek.

EdDev (approved):

PR #139, opened 2022-08-01
gotestsuite: Introduce a test suite structure for tests

EdDev:
The test suite is providing a structure to write tests.
One is able to add fixtures (currently for setup only) and ref
them from the test itself in case it is needed.

Some fixtures may be added and instructed to run before each test,
without the need to ref them explicitly from the tests.

The checkup teardown tests are adjusted to use the new suite.

### DIFF
--- a/kiagnose/internal/checkup/checkup_test.go
+++ b/kiagnose/internal/checkup/checkup_test.go
@@ -43,6 +43,8 @@ import (
 	"k8s.io/client-go/kubernetes/fake"
 	clienttesting "k8s.io/client-go/testing"
 
+	"github.com/kiagnose/kiagnose/kiagnose/internal/gotestsuite"
+
 	"github.com/kiagnose/kiagnose/kiagnose/internal/checkup"
 	"github.com/kiagnose/kiagnose/kiagnose/internal/config"
 )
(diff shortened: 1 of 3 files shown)


ormergi:
Looks like you building new testing framework for GO :wink: 

I do not quite understand how this improves the teardown test suite, each test setup in this test suite consists of different setup steps, how does setting the setup logic using the new code improve it?

EdDev:
> Looks like you building new testing framework for GO wink

I am creating a small tooling to support declarative fixtures, similar to the one pytest introduces.

> I do not quite understand how this improves the teardown test suite, each test setup in this test suite consists of different setup steps, how does setting the setup logic using the new code improve it?

Each test declares what setup/teardown fixture it requires, and the tooling is responsible to run them.
In this specific case, there was only a small portion of relevant fixtures, so the improvement is not that noticeable.
But I think it works nicely even in this case, without forcing anyone to use it and with the advantage of a reducing duplication in a simple manner. It also allows extensibility with not too much effort as well, introduced here as well.

I was considering to create a dedicated repo for it, but have not reached that yet.

EdDev:
I need to open a separate repo and check if it may interest others.
For now, there is no point to keep it open here.

PR #140, opened 2022-08-01
checkup, tests: Split TestCheckupTeardownShould()

orelmisan:
Currently, `TestCheckupTeardownShould()` has several subtests.
It is long and difficult to change.

Split each subtest of `TestCheckupTeardownShould()` to a new test.

### DIFF
--- a/kiagnose/internal/checkup/checkup_test.go
+++ b/kiagnose/internal/checkup/checkup_test.go
@@ -184,31 +184,7 @@ func TestCheckupTeardownShouldFail(t *testing.T) {
 	testTeardownShouldFailWhenNamespaceDeletionTimeoutExpires(t)
 	testTeardownShouldFailWhenClusterRoleBindingsDeletionFails(t)
 	testTeardownShouldFailWhenClusterRoleBindingsDeletionTimeoutExpires(t)
-
-	t.Run("fail when failed to delete both Namespace and ClusterRoleBindings", func(t *testing.T) {
-		testClient := newNormalizedFakeClientset()
-		testCheckup := checkup.New(
-			testClient,
-			testCheckupName,
-			&config.Config{Image: testImage, Timeout: testTimeout, ClusterRoles: newTestClusterRoles()},
-			nameGeneratorStub{},
-		)
-
-		testClient.injectResourceVersionUpdateOnNamespaceCreation()
-
-		assert.NoError(t, testCheckup.Setup())
-
-		const (
-			deleteNamespaceError          = "failed to delete Namespace"
-			deleteClusterRoleBindingError = "failed to delete ClusterRoleBindings"
-		)
-		testClient.injectDeleteErrorForResource(namespaceResource, deleteNamespaceError)
-		testClient.injectDeleteErrorForResource(clusterRoleBindingResource, deleteClusterRoleBindingError)
-
-		err := testCheckup.Teardown()
-		assert.ErrorContains(t, err, deleteNamespaceError)
-		assert.ErrorContains(t, err, deleteClusterRoleBindingError)
-	})
+	testTeardownShouldFailWhenNamespaceAndClusterRoleBindingsDeletionFails(t)
 }
 
 func testTeardownShouldFailWhenNamespaceDeletionFails(t *testing.T) {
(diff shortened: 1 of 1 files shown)


RamLavi (changes requested):
looks like the commits have the same title..
could you change it so to something more informative?

orelmisan:
> looks like the commits have the same title.. could you change it so to something more informative?

Thank you for the review @RamLavi.
The scenarios' names are quite long and don't fit nicely in the subject line.
Do you have any suggestions?

RamLavi:
> > looks like the commits have the same title.. could you change it so to something more informative?
> 
> Thank you for the review @RamLavi. The scenarios' names are quite long and don't fit nicely in the subject line. Do you have any suggestions?

As this change is just moving around code to function with no added logic, I would squash them.

orelmisan:
> > > looks like the commits have the same title.. could you change it so to something more informative?
> > 
> > 
> > Thank you for the review @RamLavi. The scenarios' names are quite long and don't fit nicely in the subject line. Do you have any suggestions?
> 
> As this change is just moving around code to function with no added logic, I would squash them.

IMO if this PR had only one commit, it would be harder for a reviewer to verify that code had just moved without changing.

RamLavi:
> > > > looks like the commits have the same title.. could you change it so to something more informative?
> > > 
> > > 
> > > Thank you for the review @RamLavi. The scenarios' names are quite long and don't fit nicely in the subject line. Do you have any suggestions?
> > 
> > 
> > As this change is just moving around code to function with no added logic, I would squash them.
> 
> IMO if this PR had only one commit, it would be harder for a reviewer to verify that code had just moved without changing.

The two better options I see are either squash or keep it split with unique commit titles.. both are good by me
I would avoid keeping the commit title the same, as it's not informative. 
But I would not insist.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/3d0b24756fa865cae2b76a4f24ef3790db4e20dc..758a6cbe9b2bfbb8596b14b62cb742f96f1e4feb): Commits subjects.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/758a6cbe9b2bfbb8596b14b62cb742f96f1e4feb..6168c895a70d48faf4111dd1fda7e72f73e81eac): Commits subjects.

RamLavi (approved):
Thanks for considering my input :) 
/lgtm

ormergi (dismissed):

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -179,106 +179,104 @@ func TestCheckTeardownShouldSucceed(t *testing.T) {
...
-func TestCheckupTeardownShould(t *testing.T) {
-	t.Run("fail when failed to delete ClusterRoleBinding", func(t *testing.T) {
-		testClient := newNormalizedFakeClientset()
-		testCheckup := checkup.New(
-			testClient,
-			testCheckupName,
-			&config.Config{Image: testImage, Timeout: testTimeout, ClusterRoles: newTestClusterRoles()},
-			nameGeneratorStub{},
-		)
+func TestTeardownShouldFailWhenDeleteClusterRoleBindingsFails(t *testing.T) {
```
EdDev, line 182: You can do this and keep a nice structure and still have it flatten in terms of nesting:

```
func TestCheckupTeardownShould(t *testing.T) {
    testTeardownShouldFailWhenDeleteClusterRoleBindingsFails(t)
    testTeardownShouldFailWhenDeleteClusterRoleBindingsTimeoutExpires(t)
    ...
}
```

orelmisan: Thanks.
Done.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -281,6 +268,21 @@ func TestCheckupTeardownShould(t *testing.T) {
...
 
+func testTeardownShouldFailWhenNamespaceDeletionFails(t *testing.T) {
+	testClient := newNormalizedFakeClientset()
+	testCheckup := checkup.New(testClient, testCheckupName, &config.Config{Image: testImage, Timeout: testTimeout}, nameGeneratorStub{})
+
+	testClient.injectResourceVersionUpdateOnNamespaceCreation()
+
+	assert.NoError(t, testCheckup.Setup())
+
+	const expectedErr = "failed to delete ClusterRoleBinding"
```
orelmisan: @ormergi I think we have a bug in this line.
I think it should be:
```suggestion
	const expectedErr = "failed to delete Namespace"
```

EdDev (approved):

RamLavi (approved):
/lgtm

kiagnose/internal/checkup/checkup_test.go
ormergi: At line 280 we configure the tests client with the error it will return on namespace delete,
Line 282 calls Teardown() which sends a delete namespace using that client.
Note that there is no actual bug, the error content doesn't have meaning in this test it could be set to "foo".
But yes the current message is wrong.

orelmisan: Would you mind to send a PR to fix it?

PR #141, opened 2022-08-01
checkup: Move teardown logic to a dedicated method

orelmisan:
The teardown sequence when using an ephemeral namespace is different from when using a target namespace.

Move the existing teardown logic of ephemeral namespace to a dedicated method.

### DIFF
--- a/kiagnose/internal/checkup/checkup.go
+++ b/kiagnose/internal/checkup/checkup.go
@@ -274,6 +274,10 @@ func (c *Checkup) SetTeardownTimeout(duration time.Duration) {
 }
 
 func (c *Checkup) Teardown() error {
+	return c.teardownWithEphemeralNamespace()
+}
+
+func (c *Checkup) teardownWithEphemeralNamespace() error {
 	const errPrefix = "teardown"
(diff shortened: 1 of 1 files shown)


orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/bc760034fd75f15c6352bc5da569a616af683b2e..2e66a4a54151563a6f35960766c92fee967a68d5): Commit subject.

ormergi (commented):
Hi Orel thanks for the PR.

It's not clear to me if this change is necessary or other changes should be done,
Are you doing to introduce the logic for handling non-ephemeral namespace on this PR?

orelmisan:
> Hi Orel thanks for the PR.
> 
> It's not clear to me if this change is necessary or other changes should be done, Are you doing to introduce the logic for handling non-ephemeral namespace on this PR?

This change is part of the preparation to support teardown when using a target namespace.
The sequence of events when performing a teardown with an ephemeral namespace is different from the sequence when using a target namespace.

orelmisan:
This change will be included as part of PR #142.

PR #142, opened 2022-08-01
checkup: Introduce target Namespace

orelmisan:
Currently, checkups are executed in ephemeral Namespace objects.

This change enables checkup execution in a target Namespace - the Namespace in which the user-supplied ConfigMap object is at.

Create and delete an ephemeral namespace only if the user-supplied ConfigMap object is at the "kiagnose" namespace.
Else, setup all objects in the target Namespace.

This PR implements the design document published in PR #126.

Followup PRs will:
1. Split checkup_test.go to test cases regarding ephemeral and target namespace scenarios.
2. Add failure test cases for the target namespace feature.

~~This PR depends on PR #140.~~

### DIFF
--- a/kiagnose/internal/checkup/checkup.go
+++ b/kiagnose/internal/checkup/checkup.go
@@ -324,6 +324,10 @@ func (c *Checkup) teardownWithTargetNamespace() error {
 		errs = append(errs, err)
 	}
 
+	if err := serviceaccount.Delete(c.client, c.serviceAccount.Namespace, c.serviceAccount.Name); err != nil {
+		errs = append(errs, err)
+	}
+
 	if len(errs) > 0 {
 		const errPrefix = "teardown"
 		return fmt.Errorf("%s: %v", errPrefix, concentrateErrors(errs))
--- a/kiagnose/internal/checkup/checkup.go
+++ b/kiagnose/internal/checkup/checkup.go
@@ -332,6 +332,10 @@ func (c *Checkup) teardownWithTargetNamespace() error {
 		errs = append(errs, err)
 	}
 
+	if err := configmap.Delete(c.client, c.resultConfigMap.Namespace, c.resultConfigMap.Name); err != nil {
+		errs = append(errs, err)
+	}
+
 	if len(errs) > 0 {
 		const errPrefix = "teardown"
 		return fmt.Errorf("%s: %v", errPrefix, concentrateErrors(errs))
--- a/kiagnose/internal/checkup/checkup.go
+++ b/kiagnose/internal/checkup/checkup.go
@@ -328,6 +328,10 @@ func (c *Checkup) teardownWithTargetNamespace() error {
 		errs = append(errs, err)
 	}
 
+	if err := rbac.DeleteRoles(c.client, c.roles); err != nil {
+		errs = append(errs, err)
+	}
+
 	if len(errs) > 0 {
 		const errPrefix = "teardown"
 		return fmt.Errorf("%s: %v", errPrefix, concentrateErrors(errs))
--- a/kiagnose/internal/checkup/checkup.go
+++ b/kiagnose/internal/checkup/checkup.go
@@ -310,9 +310,19 @@ func (c *Checkup) teardownWithEphemeralNamespace() error {
 }
 
 func (c *Checkup) teardownWithTargetNamespace() error {
+	var errs []error
+
 	if err := job.DeleteAndWait(c.client, c.job, c.teardownTimeout); err != nil {
+		errs = append(errs, err)
+	}
+
+	if err := rbac.DeleteClusterRoleBindings(c.client, c.clusterRoleBindings, c.teardownTimeout); err != nil {
+		errs = append(errs, err)
+	}
+
+	if len(errs) > 0 {
 		const errPrefix = "teardown"
-		return fmt.Errorf("%s: %v", errPrefix, err)
+		return fmt.Errorf("%s: %v", errPrefix, concentrateErrors(errs))
 	}
 
 	return nil
--- a/kiagnose/internal/checkup/checkup.go
+++ b/kiagnose/internal/checkup/checkup.go
@@ -320,6 +320,10 @@ func (c *Checkup) teardownWithTargetNamespace() error {
 		errs = append(errs, err)
 	}
 
+	if err := rbac.DeleteRoleBindings(c.client, c.roleBindings); err != nil {
+		errs = append(errs, err)
+	}
+
 	if len(errs) > 0 {
 		const errPrefix = "teardown"
 		return fmt.Errorf("%s: %v", errPrefix, concentrateErrors(errs))
--- a/kiagnose/internal/checkup/checkup.go
+++ b/kiagnose/internal/checkup/checkup.go
@@ -44,7 +44,7 @@ import (
 type Checkup struct {
 	client              kubernetes.Interface
 	teardownTimeout     time.Duration
-	namespace           *corev1.Namespace
+	ephemeralNamespace  *corev1.Namespace
 	serviceAccount      *corev1.ServiceAccount
 	resultConfigMap     *corev1.ConfigMap
 	roles               []*rbacv1.Role
@@ -55,6 +55,7 @@ type Checkup struct {
 }
 
 const (
+	KiagnoseNamespace        = "kiagnose"
 	EphemeralNamespacePrefix = "kiagnose-checkup"
 
 	ResultsConfigMapNameEnvVarName      = "RESULT_CONFIGMAP_NAME"
@@ -65,8 +66,14 @@ type namer interface {
 	Name(string) string
 }
 
-func New(c kubernetes.Interface, name string, checkupConfig *config.Config, namer namer) *Checkup {
-	nsName := namer.Name(EphemeralNamespacePrefix)
+func New(c kubernetes.Interface, targetNsName, name string, checkupConfig *config.Config, namer namer) *Checkup {
+	var ephemeralNamespace *corev1.Namespace
+	nsName := targetNsName
+
+	if targetNsName == KiagnoseNamespace {
+		nsName = namer.Name(EphemeralNamespacePrefix)
+		ephemeralNamespace = NewNamespace(nsName)
+	}
 
 	resultsConfigMapName := NameResultsConfigMap(name)
 	resultsConfigMapWriterRoleName := NameResultsConfigMapWriterRole(name)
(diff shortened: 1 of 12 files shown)


kiagnose/internal/checkup/checkup.go
```diff
@@ -292,6 +309,10 @@ func (c *Checkup) Teardown() error {
 	return nil
 }
 
+func (c *Checkup) teardownWithTargetNamespace() error {
+	return nil
+}
```
EdDev, line 345: I am not clear why we need this.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -179,6 +245,38 @@ func TestCheckTeardownShouldSucceed(t *testing.T) {
...
+
+	targetNs := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: testTargetNs,
+		},
+	}
+
+	_, err := testClient.CoreV1().Namespaces().Create(context.Background(), targetNs, metav1.CreateOptions{})
+	assert.NoError(t, err)
+	assertNamespaceExists(t, testClient.Clientset, testTargetNs)
```
EdDev, line 261: I do not think you need to validate that the `Create` succeeds beyond asserting its returned error.

kiagnose/internal/checkup/checkup_test.go
```diff
@@ -179,6 +245,38 @@ func TestCheckTeardownShouldSucceed(t *testing.T) {
...
+		testTargetNs,
+		testCheckupName,
+		&config.Config{Image: testImage, Timeout: testTimeout},
+		nameGen,
+	)
+
+	assert.NoError(t, testCheckup.Setup())
+
+	expectedEphemeralNamespaceName := nameGen.Name(checkup.EphemeralNamespacePrefix)
+	assertNamespaceDoesntExists(t, testClient.Clientset, expectedEphemeralNamespaceName)
```
EdDev: How is the ephemeral namespace relevant here? I was under the impression this test checks the explicit target-ns.

EdDev (changes requested):
Review of the 1st commit.

kiagnose/internal/checkup/checkup.go
```diff
@@ -90,7 +97,7 @@ func New(c kubernetes.Interface, name string, checkupConfig *config.Config, name
 	return &Checkup{
 		client:              c,
 		teardownTimeout:     defaultTeardownTimeout,
-		namespace:           NewNamespace(nsName),
+		namespace:           ephemeralNamespace,
```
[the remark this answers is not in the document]
orelmisan: Agreed.

kiagnose/internal/checkup/checkup.go
```diff
@@ -218,14 +225,16 @@ func (c *Checkup) Setup() error {
 	const errPrefix = "setup"
 	var err error
 
-	if c.namespace, err = namespace.Create(c.client, c.namespace); err != nil {
-		return fmt.Errorf("%s: %v", errPrefix, err)
-	}
-	defer func() {
-		if err != nil {
-			_ = namespace.DeleteAndWait(c.client, c.namespace, c.teardownTimeout)
+	if c.namespace != nil {
```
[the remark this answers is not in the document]
orelmisan: Thank you for the suggestion.
For now I rather keep thinks simpler, but we can do it in a follow-up PR if you think it has benefits.

kiagnose/internal/checkup/checkup.go
orelmisan: So Kiagnose will not delete the target namespace.
In the next commits it holds the teardown logic, which is different from the current teardown logic.

kiagnose/internal/checkup/checkup_test.go
orelmisan: Ok.

kiagnose/internal/checkup/checkup_test.go
orelmisan: I wanted to check that the ephemeral namespace is not created by mistake.

kiagnose/internal/checkup/checkup.go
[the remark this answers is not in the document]
orelmisan: Done.

kiagnose/internal/checkup/checkup_test.go
orelmisan: Done.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/3373c7eb0a35752234bc9cd51e928550d63aaa6f..afddb3640770147736b8c8afa41509babd825902): Renamed `Checkup.namespace` to `Checkup.ephemeralNamespace`.
Removed two assertions that a namespace exists after its creation in unit tests.

orelmisan:
Rebase.

RamLavi (changes requested):
added some comments,
But more importantly, did we explore the option to use OwnerReference?

orelmisan:
> added some comments, But more importantly, did we explore the option to use OwnerReference?

Most of this PR is handling the tear down process when a checkup executes in a target namespace.
As the commit progress, it can be seen that a certain order of tear down should be in place (in order to strip the privileges from the checkup):
1. Delete the checkup Job and its underlying Pod.
2. Delete the ClusterRoleBindings.
3. Delete the RoleBinding to the ConfigMapWriter Role.
4. Delete the ServiceAccount.
5. Delete the ConfigMapWriter Role.

AFAIU, when using OwnerReference we don't have control over the order in which objects are deleted.

kiagnose/internal/checkup/checkup.go
```diff
@@ -333,6 +333,10 @@ func (c *Checkup) teardownWithTargetNamespace() error {
 		errs = append(errs, err)
 	}
 
+	if err := configmap.Delete(c.client, c.resultConfigMap.Namespace, c.resultConfigMap.Name); err != nil {
```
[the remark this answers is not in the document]
orelmisan: Prior to reaching this place in the code, the results were fetched, the Job was deleted and all of the RBAC related objects were removed.

kiagnose/internal/checkup/checkup.go
EdDev: I just saw it is used in the next commit to add more content to it.

orelmisan:
Added documentation regarding the feature in the main README file, and in the checkups' README files.

kiagnose/internal/checkup/job/job.go
```diff
@@ -76,6 +76,43 @@ func WaitForJobToFinish(client kubernetes.Interface, job *batchv1.Job, timeout t
 	return updatedJob, nil
 }
 
+func DeleteAndWait(client kubernetes.Interface, job *batchv1.Job, timeout time.Duration) error {
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	deletePolicy := metav1.DeletePropagationForeground
```
EdDev, line 83: I must admit, I am not sure if this is considered safe (or safer from the default background option).
In any case, I would suggest to say a few words about this in the commit, explaining your reasoning you choose it.

EdDev (commented):
Looks very nice!

orelmisan:
> > vm latency checkup, e2e: Run checkup in a target namespace
> 
> IIUC this means that we no longer check the ephemeral NS usage, right? If this is the case, please mention it in the commit message.

Done.

kiagnose/internal/checkup/checkup.go
```diff
@@ -310,6 +310,17 @@ func (c *Checkup) teardownWithEphemeralNamespace() error {
 }
 
 func (c *Checkup) teardownWithTargetNamespace() error {
+	const errPrefix = "teardown"
```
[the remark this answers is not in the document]
orelmisan: Done.

kiagnose/internal/checkup/checkup.go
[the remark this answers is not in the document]
orelmisan: Done.

kiagnose/internal/checkup/checkup.go
```diff
@@ -310,6 +310,17 @@ func (c *Checkup) teardownWithEphemeralNamespace() error {
...
+	var errs []error
+
+	if err := job.DeleteAndWait(c.client, c.job, c.teardownTimeout); err != nil {
+		errs = append(errs, err)
```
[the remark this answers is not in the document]
orelmisan: Moved the usage of `var errs []error` to the first commit that needs more than one error.

kiagnose/internal/checkup/checkup.go
orelmisan: Moved it to the first commit which is doing a teardown action.

kiagnose/internal/checkup/checkup.go
```diff
@@ -274,14 +283,22 @@ func (c *Checkup) SetTeardownTimeout(duration time.Duration) {
 }
 
 func (c *Checkup) Teardown() error {
+	if c.ephemeralNamespace != nil {
```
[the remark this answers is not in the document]
orelmisan: Done.

kiagnose/internal/checkup/job/job.go
orelmisan: Added reasoning to the commit message.
Can you please elaborate why do you think it is not considered safe?

orelmisan:
Changes [1](https://github.com/kiagnose/kiagnose/compare/7bec0bd191fc1a3407da46314b787ed8ac33a946..f84b9d171cd87ca24738467ceb4bb60d0812aa06) and [2](https://github.com/kiagnose/kiagnose/compare/f84b9d171cd87ca24738467ceb4bb60d0812aa06..69447cf693152844c5b05462725c4423d7492a31): Answered review comments by @RamLavi and @EdDev.

RamLavi:
> > added some comments, But more importantly, did we explore the option to use OwnerReference?
> 
> Most of this PR is handling the tear down process when a checkup executes in a target namespace. As the commit progress, it can be seen that a certain order of tear down should be in place (in order to strip the privileges from the checkup):
> 
> 1. Delete the checkup Job and its underlying Pod.
> 2. Delete the ClusterRoleBindings.
> 3. Delete the RoleBinding to the ConfigMapWriter Role.
> 4. Delete the ServiceAccount.
> 5. Delete the ConfigMapWriter Role.
> 6. Delete the results ConfigMap.
> 
> AFAIU, when using OwnerReference we don't have control over the order in which objects are deleted.

I would try nonetheless just to see, it would make the teardown responsibilities out of our hands. 
Even if it does needs ordering, perhaps if we ever let go of the RBAC roles responsibilities then we won't need to worry about that.

RamLavi:
> I would try nonetheless just to see, it would make the teardown responsibilities out of our hands.

we talked about it offline, I understand why we can't use it now. 
I would reconsider doing it once/if we ever renounce our rback responsibilities.

RamLavi (approved):
/lgtm

kiagnose/internal/checkup/job/job.go
EdDev: It may be unsafe because it requires the dependent objects to be deleted first, which may actually fail and therefore stuck this operation, or even cause it to fail.

If something else blocks the deletion of the dependent, it is a deadlock.

Said all that, I am not sure what is the best approach in general. It may be good enough here.
The explanation you added to the commit seems convincing.
For the long run, some thought about how to delete things and the risks of each approach may need exploration.

EdDev (approved):
Thank you!

PR #148, opened 2022-08-15
CI: Fix echo checkup e2e tests to use locally built images

orelmisan:
Currently, Kiagnose e2e test is a success scenario with the echo checkup.
Move this e2e test to be under the "echo.checks" workflow in order to be aligned with the VM latency checkup workflow.

Furthermore, the echo e2e test is using kiagnose and echo-checkup images pulled from quay and not locally built images
from current source code.

Move the echo e2e test code from `automation/e2e.sh` to `checkups/echo/automation/e2e.sh`.
Use locally built images of kiagnose and echo-checkup.
Align the e2e job with the e2e job of vm latency checkup.
Use target namespace instead of ephemeral namespace in the e2e test.

### DIFF
--- a/checkups/echo/automation/e2e.sh
+++ b/checkups/echo/automation/e2e.sh
@@ -28,8 +28,8 @@ KUBECTL=${KUBECTL:-$PWD/kubectl}
 KIND=${KIND:-$PWD/kind}
 CLUSTER_NAME=${CLUSTER_NAME:-kind}
 
-FRAMEWORK_IMAGE="quay.io/kiagnose/kiagnose:main"
-CHECKUP_IMAGE="quay.io/kiagnose/echo-checkup:main"
+FRAMEWORK_IMAGE="quay.io/kiagnose/kiagnose:devel"
+CHECKUP_IMAGE="quay.io/kiagnose/echo-checkup:devel"
 
 KIAGNOSE_NAMESPACE=kiagnose
 KIAGNOSE_JOB=echo-checkup
--- a/checkups/echo/automation/e2e.sh
+++ b/checkups/echo/automation/e2e.sh
@@ -1,0 +1,137 @@
+#!/usr/bin/env bash
+#
+# This file is part of the kiagnose project
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Copyright 2022 Red Hat, Inc.
+#
+
+set -e
+
+ARGCOUNT=$#
+
+SCRIPT_PATH=$(dirname $(realpath -s $0))
+
+KUBECTL=${KUBECTL:-$PWD/kubectl}
+
+KIND=${KIND:-$PWD/kind}
+CLUSTER_NAME=${CLUSTER_NAME:-kind}
+
+FRAMEWORK_IMAGE="quay.io/kiagnose/kiagnose:main"
+CHECKUP_IMAGE="quay.io/kiagnose/echo-checkup:main"
+
+KIAGNOSE_NAMESPACE=kiagnose
+KIAGNOSE_JOB=echo-checkup
+ECHO_CONFIGMAP=echo-checkup
+
+TARGET_NAMESPACE="echo-checkup-e2e-test"
+
+options=$(getopt --options "" \
+    --long deploy-checkup,run-tests,clean-run,help\
+    -- "${@}")
+eval set -- "$options"
+while true; do
+    case "$1" in
+    --deploy-checkup)
+        OPT_DEPLOY_CHECKUP=1
+        ;;
+    --run-tests)
+        OPT_RUN_TEST=1
+        ;;
+    --clean-run)
+        OPT_CLEAN_RUN=1
+        ;;
+    --help)
+        set +x
+        echo -n "$0 [--deploy-checkup] [--run-tests] [--clean-run]"
+        exit
+        ;;
+    --)
+        shift
+        break
+        ;;
+    esac
+    shift
+done
+
+if [ "${ARGCOUNT}" -eq "0" ] ; then
+    OPT_DEPLOY_CHECKUP=1
+    OPT_RUN_TEST=1
+fi
+
+if [ -n "${OPT_DEPLOY_CHECKUP}" ]; then
+    echo
+    echo "Deploy echo checkup..."
+    echo
+
+    ${KIND} load docker-image "${CHECKUP_IMAGE}" --name "${CLUSTER_NAME}"
+fi
+
+if [ -n "${OPT_RUN_TEST}" ]; then
+    ${KUBECTL} create namespace ${TARGET_NAMESPACE}
+    echo
+    echo "Deploy ConfigMap with input data: "
+    echo
+    cat <<EOF | ${KUBECTL} apply -f -
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ${ECHO_CONFIGMAP}
+  namespace: ${TARGET_NAMESPACE}
+data:
+  spec.image: ${CHECKUP_IMAGE}
+  spec.timeout: 1m
+  spec.param.message: "Hi!"
+EOF
+
+    echo
+    echo "Deploy and run kiagnose job: "
+    echo
+    cat <<EOF | ${KUBECTL} apply -f -
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: ${KIAGNOSE_JOB}
+  namespace: ${KIAGNOSE_NAMESPACE}
+spec:
+  backoffLimit: 0
+  template:
+    spec:
+      serviceAccount: kiagnose
+      restartPolicy: Never
+      containers:
+        - name: framework
+          image: ${FRAMEWORK_IMAGE}
+          env:
+            - name: CONFIGMAP_NAMESPACE
+              value: ${TARGET_NAMESPACE}
+            - name: CONFIGMAP_NAME
+              value: ${ECHO_CONFIGMAP}
+EOF
+
+    ${KUBECTL} wait --for=condition=complete --timeout=10m job.batch/${KIAGNOSE_JOB} -n ${KIAGNOSE_NAMESPACE}
+
+     echo
+     echo "Result:"
+     echo
+     ${KUBECTL} get configmap ${ECHO_CONFIGMAP} -n ${TARGET_NAMESPACE} -o yaml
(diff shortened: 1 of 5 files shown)


RamLavi (commented):
can you split the "Move the echo e2e test code from `automation/e2e.sh` to
`checkups/echo/automation/e2e.sh`." part to a new commit?

orelmisan:
> can you split the "Move the echo e2e test code from `automation/e2e.sh` to `checkups/echo/automation/e2e.sh`." part to a new commit?

This move has implication on other files, so I don't think I can do it more neatly with another commit.

orelmisan:
Change: Added a commit to make the change to `.github/workflows/checkup-echo.check.yaml` more gradual.

RamLavi (commented):
I think we should be able to recognize release commits (i.e. tagged `vx.y.z`) so that we DO check these commits with pushed quay images.
Not necessarily in this PR though. WTYT @orelmisan ?

orelmisan:
> Thanks for splitting up the commits!
> 
> I think we should be able to recognize release commits (i.e. tagged `vx.y.z`) so that we DO check these commits with pushed quay images. Not necessarily in this PR though. WTYT @orelmisan ?

Sounds like a good idea to me!
I think our CI and build system needs some love and care.
This change was made mostly to enable the change introduced by PR #149.

orelmisan:
> This change was made mostly to enable the change introduced by PR https://github.com/kiagnose/kiagnose/pull/149.

And to enable the PR that will follow, which will introduce a breaking API change.

RamLavi:
> > This change was made mostly to enable the change introduced by PR #149.
> 
> And to enable the PR that will follow, which will introduce a breaking API change.

Let's do it (= use pushed quay images on release commits)on a follow up PR then.
I think it's OK not to insert it as long as we do it before the next release.

RamLavi (dismissed):
/lgtm

checkups/echo/automation/e2e.sh
```diff
@@ -0,0 +1,137 @@
...
+CLUSTER_NAME=${CLUSTER_NAME:-kind}
+
+FRAMEWORK_IMAGE="quay.io/kiagnose/kiagnose:main"
+CHECKUP_IMAGE="quay.io/kiagnose/echo-checkup:main"
+
+KIAGNOSE_NAMESPACE=kiagnose
+KIAGNOSE_JOB=echo-checkup
+ECHO_CONFIGMAP=echo-checkup
+
+TARGET_NAMESPACE="target-ns"
```
ormergi, line 38: Please change the ns name to `echo-checkup-e2e-test`.

checkups/echo/automation/e2e.sh
```diff
@@ -0,0 +1,137 @@
...
+    esac
+    shift
+done
+
+if [ "${ARGCOUNT}" -eq "0" ] ; then
+    OPT_DEPLOY_CHECKUP=1
+    OPT_RUN_TEST=1
+fi
+
+if [ -n "${OPT_DEPLOY_CHECKUP}" ]; then
```
ormergi, line 73: The name of the corresponding flag `--deploy-checkup` is confusing it makes me think we create objects on the cluster.

I think the right place for it is at `checkups/echo/automation/make.sh` at a new target maybe `--push-checkup-image`.

checkups/echo/automation/e2e.sh
```diff
@@ -0,0 +1,137 @@
...
+  name: ${ECHO_CONFIGMAP}
+  namespace: ${TARGET_NAMESPACE}
+data:
+  spec.image: ${CHECKUP_IMAGE}
+  spec.timeout: 1m
+  spec.param.message: "Hi!"
+EOF
+
+    echo
+    echo "Deploy and run kiagnose job: "
```
ormergi, line 100: nit: run is redundent.

checkups/echo/automation/e2e.sh
```diff
@@ -0,0 +1,137 @@
...
+              value: ${ECHO_CONFIGMAP}
+EOF
+
+    ${KUBECTL} wait --for=condition=complete --timeout=10m job.batch/${KIAGNOSE_JOB} -n ${KIAGNOSE_NAMESPACE}
+
+     echo
+     echo "Result:"
+     echo
+     ${KUBECTL} get configmap ${ECHO_CONFIGMAP} -n ${TARGET_NAMESPACE} -o yaml
+     ${KUBECTL} get configmap ${ECHO_CONFIGMAP} -n ${TARGET_NAMESPACE} -o yaml | grep -q "status.result.echo: Hi!"
```
ormergi, line 131: Is it possible to add an assert to make sure the checkup ran?
Maybe checking that `status.result.echo `equals to the expected message?

ormergi (commented):
Thanks for the PR, good catch!

Regarding "CI: Move echo e2e tests" commit it would be much easier to review if you would split the part of moving code around and changing it to different commits.
**If its not a lot of work** please do it so we can be sure to not miss anything.

checkups/echo/automation/e2e.sh
orelmisan: Done.

checkups/echo/automation/e2e.sh
orelmisan: `checkups/echo/automation/e2e.sh` was created from a copy and paste operation from `checkups/kubevirt-vm-latency/automation/e2e.sh` minus the kubevirt / CNAO / Multus stuff.

If you wish to change the `--deploy-checkup` option, please open a PR against this one and change the following files:
- `checkups/echo/automation/e2e.sh`
- `.github/workflows/checkup-echo.check.yaml` 
- `checkups/kubevirt-vm-latency/automation/e2e.sh`
- `.github/workflows/checkup-kubevirt-vm-latency.check.yaml`

checkups/echo/automation/e2e.sh
orelmisan: `checkups/echo/automation/e2e.sh` was created from a copy and paste operation from `checkups/kubevirt-vm-latency/automation/e2e.sh` minus the kubevirt / CNAO / Multus stuff.

If you wish to change this, please open a PR against this one and change the following files:
- `checkups/echo/automation/e2e.sh`
- `checkups/kubevirt-vm-latency/automation/e2e.sh`

checkups/echo/automation/e2e.sh
orelmisan: > Is it possible to add an assert to make sure the checkup ran?

Can you please elaborate?

orelmisan: If its not a small change, I suggest to do it in a follow up PR.
Two large PRs depends on this one.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/a75b265e7145b4988f556e8d781fc4fa46c33dbc..aa8b5e7cbe05c7b93025ad60908e4994e4e0f7f0): Answered @ormergi's [comment](https://github.com/kiagnose/kiagnose/pull/148#discussion_r947999837).

checkups/echo/automation/e2e.sh
ormergi: The fact that you followed other script example doesn't mean you should not change it,
if something is wrong at the source we should not follow it.

If you are not going to move it to `checkups/echo/automation/make.sh` could you at least rename it to `--push-checkup-image`?

checkups/echo/automation/e2e.sh
ormergi: Same here, note that it is just an echo message...

checkups/echo/automation/e2e.sh
ormergi: I meant to make the script fail if the checkup fails (the message won't show at the status), here is what we have at the latency checkup https://github.com/kiagnose/kiagnose/blob/main/checkups/kubevirt-vm-latency/automation/e2e.sh#L228-L232.

checkups/echo/automation/e2e.sh
orelmisan: > The fact that you followed other script example doesn't mean you should not change it,
if something is wrong at the source we should not follow it.

IMO the two `checkups/*/automation/e2e.sh` scripts should have the same common options.
If one of them exists, I think the other should follow in order to be aligned with the first.
After they are aligned, changes should be made do both of them simultaneously.

checkups/echo/automation/e2e.sh
orelmisan: At line 20 there is a `set -e`.directive.
If this command (line 131) fails:
```bash
${KUBECTL} get configmap ${ECHO_CONFIGMAP} -n ${TARGET_NAMESPACE} -o yaml | grep -q "status.result.echo: Hi!"
```
the entire test fails.

EdDev (commented):
I'm good with this change.
There is a small disagreement on the 2nd commit, and I would prefer it resolved before continuing.

As a side note, the 2nd commit is suffering from doing too many things. Next time, it would be preferred to do each part in a separate commit. Doing so would also strengthen the point that the changes are limited to what the author intended for. Said that, I would not invest in to further to split it now, it is not worth the investment IMO.

checkups/echo/automation/e2e.sh
orelmisan: @ormergi I'm in favor of:
> Leave it as is, and follow up with a naming PR fix.

Or

checkups/echo/automation/e2e.sh
ormergi: > Once the job manifest is deployed, it is automatically ran. Isn't this the case?

Yes, this is why I think there is no need to mention "run".

checkups/echo/automation/e2e.sh
ormergi: From my perspective, there is no difference.

checkups/echo/automation/e2e.sh
ormergi: How about adding a message that the test is failed?

ormergi:
It looks good to me overall.
Make sure to not combine moving and changing code in the same PR it makes it hard to follow.

I am not in favor of the pricipale that changing something in a file must be in its own PR if the thing we want to change exists in other files, for example the mentioned script flag name, or the definition of "user-supplied Config Map" in the design docs.
I think it is better to change one file at PR and align the rest of the files on the same PR or on a follow-up than keep going and create another file with the same wrong pattern.

checkups/echo/automation/e2e.sh
orelmisan: Do you mind if we'll do it in a follow-up PR?

orelmisan: This is out of the scope of this PR.

EdDev:
> I am not in favor of the principle that changing something in a file must be in its own PR if the thing we want to change exists in other files, for example, the mentioned script flag name, or the definition of "user-supplied Config Map" in the design docs.
> I think it is better to change one file at PR and align the rest of the files on the same PR or on a follow-up than keep going and create another file with the same wrong pattern.

In general, the author has the privilege to choose the scope of the change proposed.
If it is not wrong, options for improving some parts can be left as optional.
I am in favor of fixing problems seen while working, but I do not think this can (or should) be strongly enforced.

In the end, there is a balance that should be kept. I think in this change the balance is fine, and it is good that some points have been raised on issues with the current scripts.

Thank you both for the work on this.

EdDev (approved):
Thank you!

PR #152, opened 2022-09-01
e2e infra: Support multi nodes cluster

ormergi:
This change is required to support testing network connectivity
across cluster nodes over linux-bridge.

Currently KinD does not support secondary networks for cluster nodes.

Each worker node is connected through veth pair to establish connectivity.
Next each node veth is connected to its own bridge that the cluster workloads will consume through NetworkAttachmentDefinition.

Notes:
- This change introduces using koko [[1]](https://github.com/redhat-nfvpe/koko) utility, it enables connecting between containers network namespaces in a convenient way.
- Connecting between container network namespaces requires high privileges.
- There is an emerging project that enables spinning up KinD cluster with secondary networks [[2]](https://github.com/aojea/kind-networking-plugins/tree/main/baremetal)
It seems to be in PoC stage, once it will productive we can use it.

[1] https://github.com/redhat-nfvpe/koko
[2] https://github.com/aojea/kind-networking-plugins/tree/main/baremetal

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ bin/
 # Others
 kind
 kubectl
+# koko - containers network connector untility
+koko
--- a/automation/e2e.sh
+++ b/automation/e2e.sh
@@ -28,10 +28,15 @@ KIND_VERSION=${KIND_VERSION:-v0.12.0}
 KIND=${KIND:-$PWD/kind}
 CLUSTER_NAME=${CLUSTER_NAME:-kind}
 
+KOKO_VERSION=${KOKO_VERSION:-0.83}
+KOKO=${KOKO:-$PWD/koko}
+BRIDGE_NAME=${BRIDGE_NAME:-br10}
+VETH_NAME=${VETH_NAME:-link_${BRIDGE_NAME}}
+
 FRAMEWORK_IMAGE="quay.io/kiagnose/kiagnose:devel"
 
 options=$(getopt --options "" \
-    --long install-kind,install-kubectl,create-cluster,delete-cluster,deploy-kiagnose,help\
+    --long install-kind,install-kubectl,create-cluster,create-multi-node-cluster,delete-cluster,deploy-kiagnose,help\
     -- "${@}")
 eval set -- "$options"
 while true; do
@@ -45,6 +50,9 @@ while true; do
     --create-cluster)
         OPT_CREATE_CLUSTER=1
         ;;
+    --create-multi-node-cluster)
+        OPT_CREATE_MULTI_NODE_CLUSTER=1
+        ;;
     --delete-cluster)
         OPT_DELETE_CLUSTER=1
         ;;
@@ -53,7 +61,7 @@ while true; do
         ;;
     --help)
         set +x
-        echo "$0 [--install-kind] [--install-kubectl] [--create-cluster] [--delete-cluster] [--deploy-kiagnose]"
+        echo "$0 [--install-kind] [--install-kubectl] [--create-cluster] [--create-multi-node-cluster] [--delete-cluster] [--deploy-kiagnose]"
         exit
         ;;
     --)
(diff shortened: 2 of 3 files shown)


automation/e2e.sh
```diff
@@ -100,6 +108,45 @@ if [ -n "${OPT_CREATE_CLUSTER}" ]; then
...
+        ${KUBECTL} get nodes -o wide
+    else
+        echo "Cluster '${CLUSTER_NAME}' already exists!"
+    fi
+
+    if [ ! -f "${KOKO}" ]; then
+        curl -Lo "${KOKO}" https://github.com/redhat-nfvpe/koko/releases/download/v${KOKO_VERSION}/koko_${KOKO_VERSION}_linux_amd64
+        chmod +x "${KOKO}"
+        echo "koko installed successfully at ${KOKO}"
+    fi
```
orelmisan, line 133: Please consider splitting `koko` installation to another target, similar to the way we handle the installation of `kind` and `kubectl`.

.github/workflows/checkup-kubevirt-vm-latency.check.yaml
```diff
@@ -90,7 +90,7 @@ jobs:
         working-directory: ./checkups/kubevirt-vm-latency
         run: ./automation/make.sh --build-checkup --build-checkup-image
       - name: Start cluster
-        run: ./automation/make.sh --e2e -- --install-kind --install-kubectl --create-cluster
+        run: ./automation/make.sh --e2e -- --install-kind --install-kubectl --create-multi-node-cluster;
```
orelmisan, line 93: Do we still need the `--create-cluster` option (and its logic)?

nit:
```suggestion
        run: ./automation/make.sh --e2e -- --install-kind --install-kubectl --create-multi-node-cluster
```

orelmisan (commented):

automation/e2e.sh
ormergi: Unlike `kubectl` and `kind`, `koko` is used only for this "target" this is why I placed it here. make sense?

.github/workflows/checkup-kubevirt-vm-latency.check.yaml
ormergi: There may be cases where a single node cluster is better fit in terms of CI time, such the echo checkup.

orelmisan: Thanks for the answer.

orelmisan:
Now that the second commit had moved to PR #155, the code in this PR is not in use and is not tested by the CI.
IMO it should return to this PR.

ormergi:
I created this PR to not distract the discussion of #155 changes.
Having the second commit changes w/o #155 may cause CI to fail (in case both VMs are scheduled on different nodes).

ormergi:
This PR changes are tested at #155.

orelmisan:
> Having the second commit changes w/o #155 may cause CI to fail (in case both VMs are scheduled on different nodes).

Can you please elaborate why should the CI fail in a multi-node cluster?

ormergi:
> Can you please elaborate why would the CI fail when using a multi-node cluster?

It wont make it fail, sorry for the noise.

orelmisan:
I suggest to bring back the commit that uses it for the VM latency checkup, so this code will be executed by the CI until PR #155 will be merged.
Its your call, please tell me how do you want to proceed.

ormergi:
I prefer to continue with this PR as it, my intention was to add support for multi node setup and use it on a follow up PR.

orelmisan (approved):
Thank you @ormergi

.gitignore
```diff
@@ -18,3 +18,5 @@ bin/
 # Others
 kind
 kubectl
+# koko - continaers network connector untility
+koko
```
orelmisan, line 22: Can you please add a newline after `koko`?

ormergi: Done

orelmisan (approved):

.gitignore
```diff
@@ -18,3 +18,5 @@ bin/
 # Others
 kind
 kubectl
+# koko - continaers network connector untility
```
EdDev, line 21: typo in the text here (`continaers`).
I am not sure why you need to describe it here (we do not do it for `kubectl` for example), but I am fine with whatever @orelmisan prefers.

orelmisan: I've reviewed it on my phone so I haven't noticed this typo, thanks @EdDev.
This tool is way less known then `kind` or `kubectl` so the comment is appropriate IMO.

ormergi: Yep, I added a description because it's not common as other tools we use.

orelmisan (approved):

EdDev (approved):
Thanks!

Raising

PR #154, opened 2022-09-01
vmlatency: Optimize waiting for VMs readiness

ormergi:
This change optimizes the checkup setup flow by considering the VMs ready when `Ready` condition is set.
The "Ready" condition will be added to a VM status before `AgentConnected` condition, roughly 10 seconds on Alpine VM.

Moreover, there is no need to wait for the "AgentConnected" condition at all as `checkup.Run()` waits for the VMI IP addresses to reflects on the VMI status (which is reported by the guest agent).


orelmisan (commented):

AFAIU from the Kubevirt [code](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-controller/watch/vmi.go#L796), the `VirtualMachineInstanceReady` condition represents the fact the the VMI's underlying pod is ready (and not the VM itself).

Could you please elaborate how this change improves the code?

BTW, do you know if this field is documented? I haven't found documentation about it.

ormergi:
> AFAIU from the Kubevirt [code](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-controller/watch/vmi.go#L796), the `VirtualMachineInstanceReady` condition represents the fact the the VMI's underlying pod is ready (and not the VM itself).
> 
> Could you please elaborate how this change improves the code?

There is no need to wait for the AgentConnected condition (QEMU guest-agent to start and report) at the checkup setup phase
waiting for Ready condition is sufficient.
checkup Run waits for the VMs  status to report their IP addresses which means the guest-agent is up.

It improves the time it takes for the checkup setup to finish.


> BTW, do you know if this field is documented? I haven't found documentation about it.

I didn't find it either.

orelmisan:
I understand why its only necessary to wait for `VirtualMachineInstanceReady` after creating the source VMI, before creating the target VMI.

What I don't understand is why `Setup()` does not wait for the two VMIs to be completely ready (having an IP address or some other condition) before exiting the method.

This way, the `Run()` method should not have to wait for it.

ormergi:
Rebased

ormergi:
> What I don't understand is why `Setup()` does not wait for the two VMIs to be completely ready (having an IP address or some other condition) before exiting the method.
> 
> This way, the `Run()` method should not have to wait for it.

Setup() and Run() were implemented with the following mindset:
Setup() is responsible for spinning up VMs, it waits for the VMs to have the `Ready` condition to ensure their spec is valid (kind of a sanity check).
Run() preforms the latency measurement which requires the target VM IP address, waiting for the IP address to appear is necessary since `Ready` condition doesn't indicate that the VM status is updated with the IP address.

I will push new changes to change it to what you suggest.

ormergi:
@orelmisan please take another look.

/cc @RamLavi @EdDev

orelmisan:
@ormergi could you please rebase?

ormergi:
[Rebased](https://github.com/kiagnose/kiagnose/compare/11046b1282b7f8d442c3e4f757996051048be2f9..f4f8dfdd309f0f629bc6bd632b89babad1bf0468)

orelmisan (commented):
Thank you @ormergi.
The code looks good to me.

I want to suggest a minor wording change to the commit message and PR description:

> Currently, at the checkup setup phase, a VM is considered ready when the "Ready" condition is present in the VM status.
`Run()` waits for the VM IP addresses to be present in the VM status, which indicated that the VM has booted (as the guest-agent reported the IP addresses).


> Instead of waiting for different states of the VMs at different times, wait for the VM IP address to report as part of `Setup()`.
`Run()` will only console to the VM and start the latency measurement w/o waiting for any state of the VM.

orelmisan:
nit: The commit subject and PR title are different.

ormergi:
> I want to suggest a minor wording change to the commit message and PR description:

It would be easier if you would mention only the sentence or part of it you suggest to change :slightly_smiling_face: 

> nit: The commit subject and PR title are different.

Note that in general, thy don't have to match (even when there is a single commit).

RamLavi (approved):
/lgtm

orelmisan (approved):
Thank you @ormergi.

EdDev (approved):
Thanks!

Raising.

PR #155, opened 2022-09-01
kubevirt-vm-latency: Ensure VMs are not scheduled on the same node by default

ormergi:
Currently, when both the source and the target node are not defined by the user, the source and target VMs may end up scheduled on the same node.
Though this may be useful for some use cases it's not the desired default behavior.

With this PR changes, VMs will create with `latency-check-vm` label which indicates they belong to the latency checkup.
As long as both "source_node" and "target_node" are not specified in the checkup config, both VMs won't end up running on the same node (using a [Pod Anti Affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#types-of-inter-pod-affinity-and-anti-affinity) rule and the `latency-check-vm` label as the `labelSelector`).

In order to override this behavior and force both VMs to run on the same node, specify both `source_node` and `target_node` in the checkup config (each VM will be created with a corresponding [Node Affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity) rule).

Specifying only one of `source_node` and `source_node` is no longer supported.

~~This PR depends on #152 (first commit), it enables e2e testing of this PR changes.~~

### DIFF
--- a/.github/workflows/checkup-kubevirt-vm-latency.check.yaml
+++ b/.github/workflows/checkup-kubevirt-vm-latency.check.yaml
@@ -88,7 +88,7 @@ jobs:
         working-directory: ./checkups/kubevirt-vm-latency
         run: ./automation/make.sh --build-checkup --build-checkup-image
       - name: Start cluster
-        run: ./automation/make.sh --e2e -- --install-kind --install-kubectl --create-cluster
+        run: ./automation/make.sh --e2e -- --install-kind --install-kubectl --create-multi-node-cluster
       - name: Deploy kiagnose
         run: ./automation/make.sh --e2e -- --deploy-kiagnose
       - name: Deploy kubevirt, CNAO and the NetworkAttachementDefinition
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
@@ -25,6 +25,8 @@ import (
 	"strings"
 	"time"
 
+	k8scorev1 "k8s.io/api/core/v1"
+
 	kvcorev1 "kubevirt.io/api/core/v1"
 
 	netattdefv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
@@ -66,6 +68,12 @@ func (c *checkup) Preflight() error {
 	return nil
 }
 
+const (
+	SourceVmiName       = "latency-check-source"
+	TargetVmiName       = "latency-check-target"
+	LabelLatencyCheckVM = "latency-check-vm"
+)
+
 func (c *checkup) Setup(ctx context.Context) error {
 	const (
 		errMessagePrefix = "setup"
@@ -73,10 +81,8 @@ func (c *checkup) Setup(ctx context.Context) error {
 		defaultSetupTimeout = time.Minute * 10
 
 		networkName   = "net0"
-		sourceVmiName = "latency-check-source"
 		sourceVmiMac  = "02:00:00:01:00:01"
 		sourceVmiCidr = "192.168.100.10/24"
-		targetVmiName = "latency-check-target"
 		targetVmiMac  = "02:00:00:02:00:02"
 		targetVmiCidr = "192.168.100.20/24"
 	)
@@ -88,24 +94,21 @@ func (c *checkup) Setup(ctx context.Context) error {
 		return fmt.Errorf("%s: %v", errMessagePrefix, err)
 	}
 
-	sourceVmi := newLatencyCheckVmi(sourceVmiName, c.params.SourceNodeName, networkName, netAttachDef, sourceVmiMac, sourceVmiCidr)
-	targetVmi := newLatencyCheckVmi(targetVmiName, c.params.TargetNodeName, networkName, netAttachDef, targetVmiMac, targetVmiCidr)
+	sourceVmi := newLatencyCheckVmi(SourceVmiName, c.params.SourceNodeName, networkName, netAttachDef, sourceVmiMac, sourceVmiCidr)
+	targetVmi := newLatencyCheckVmi(TargetVmiName, c.params.TargetNodeName, networkName, netAttachDef, targetVmiMac, targetVmiCidr)
 
 	if err = vmi.Start(c.client, c.namespace, sourceVmi); err != nil {
 		return fmt.Errorf("%s: %v", errMessagePrefix, err)
 	}
-
 	if err = vmi.Start(c.client, c.namespace, targetVmi); err != nil {
 		return fmt.Errorf("%s: %v", errMessagePrefix, err)
 	}
 
 	waitCtx, cancel := context.WithTimeout(ctx, defaultSetupTimeout)
 	defer cancel()
-
 	if c.targetVM, err = vmi.WaitUntilReady(waitCtx, c.client, c.namespace, targetVmi.Name); err != nil {
 		return fmt.Errorf("%s: %v", errMessagePrefix, err)
 	}
-
 	if c.sourceVM, err = vmi.WaitUntilReady(waitCtx, c.client, c.namespace, sourceVmi.Name); err != nil {
 		return fmt.Errorf("%s: %v", errMessagePrefix, err)
 	}
@@ -134,8 +137,17 @@ func newLatencyCheckVmi(
 		vmiInterface = vmi.NewInterface(networkName, vmi.WithMacAddress(macAddress), vmi.WithBridgeBinding())
 	}
 
+	vmLabel := vmi.Label{Key: LabelLatencyCheckVM, Value: ""}
+	var affinity *k8scorev1.Affinity
+	if nodeName != "" {
+		affinity = &k8scorev1.Affinity{NodeAffinity: vmi.NewNodeAffinity(nodeName)}
+	} else {
+		affinity = &k8scorev1.Affinity{PodAntiAffinity: vmi.NewPodAntiAffinity(vmLabel)}
+	}
+
 	return vmi.NewAlpine(name,
-		vmi.WithNodeSelector(nodeName),
+		vmi.WithLabels(vmLabel),
+		vmi.WithAffinity(affinity),
 		vmi.WithMultusNetwork(networkName, netAttachDef.Namespace+"/"+netAttachDef.Name),
 		vmi.WithInterface(vmiInterface),
 		vmi.WithCloudInitNoCloudNetworkData(networkData),
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
@@ -37,6 +37,7 @@ import (
 
 	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/vmlatency/internal/checkup"
 	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/vmlatency/internal/config"
+	"github.com/kiagnose/kiagnose/checkups/kubevirt-vm-latency/vmlatency/internal/vmi"
 )
 
 const (
@@ -154,6 +155,74 @@ func TestCheckupSetupShouldCreateVMsWith(t *testing.T) {
 			}
 		})
 	}
+}
+
+func TestCheckupSetupShouldCreateVMsWithPodAntiAffinity(t *testing.T) {
+	t.Run("when source and target nodes names are not specified", func(t *testing.T) {
+		testClient := newTestClient()
+		testClient.returnNetAttachDef = newTestNetAttachDef("")
+		testCheckup := checkup.New(testClient, testNamespace, config.CheckupParameters{}, &checkerStub{})
+
+		assert.NoError(t, testCheckup.Preflight())
+		assert.NoError(t, testCheckup.Setup(context.Background()))
+		assertVmiPodAntiAffinityExist(t, testClient, checkup.SourceVmiName)
+		assertVmiPodAntiAffinityExist(t, testClient, checkup.TargetVmiName)
+		assertVmiNodeAffinityNotExist(t, testClient, checkup.SourceVmiName)
+		assertVmiNodeAffinityNotExist(t, testClient, checkup.TargetVmiName)
+	})
+}
+
+func TestCheckupSetupShouldCreateVMsWithNodeAffinity(t *testing.T) {
+	t.Run("when both source and target node names are specified", func(t *testing.T) {
+		const (
+			testSourceNode = "source.example.com"
+			testTargetNode = "target.example.com"
+		)
+		testClient := newTestClient()
+		testClient.returnNetAttachDef = newTestNetAttachDef("blah")
+		testCheckupParams := config.CheckupParameters{SourceNodeName: testSourceNode, TargetNodeName: testTargetNode}
+		testCheckup := checkup.New(testClient, testNamespace, testCheckupParams, &checkerStub{})
+
+		assert.NoError(t, testCheckup.Preflight())
+		assert.NoError(t, testCheckup.Setup(context.Background()))
+		assertVmiNodeAffinityExist(t, testClient, checkup.SourceVmiName, testSourceNode)
+		assertVmiNodeAffinityExist(t, testClient, checkup.TargetVmiName, testTargetNode)
+		assertVmiPodAntiAffinityNotExist(t, testClient, checkup.SourceVmiName)
+		assertVmiPodAntiAffinityNotExist(t, testClient, checkup.TargetVmiName)
+	})
+}
+
+func assertVmiPodAntiAffinityExist(t *testing.T, testClient *clientStub, vmiName string) {
+	actualVmi, err := testClient.GetVirtualMachineInstance(testNamespace, vmiName)
+	assert.NoError(t, err)
+	assert.NotNil(t, actualVmi.Spec.Affinity.PodAntiAffinity)
+	expectedPodAntiAffinity := vmi.NewPodAntiAffinity(vmi.Label{Key: checkup.LabelLatencyCheckVM, Value: ""})
+	assert.Equal(t, expectedPodAntiAffinity, actualVmi.Spec.Affinity.PodAntiAffinity)
+}
+
+func assertVmiPodAntiAffinityNotExist(t *testing.T, testClient *clientStub, name string) {
+	actualVmi, err := testClient.GetVirtualMachineInstance(testNamespace, name)
+	assert.NoError(t, err)
+
+	assert.Nil(t, actualVmi.Spec.Affinity.PodAntiAffinity)
+}
+
+func assertVmiNodeAffinityExist(t *testing.T, testClient *clientStub, vmiName, nodeName string) {
+	actualVmi, err := testClient.GetVirtualMachineInstance(testNamespace, vmiName)
+	assert.NoError(t, err)
+
+	actualVmiNodeAffinity := actualVmi.Spec.Affinity.NodeAffinity
+	assert.NotNil(t, actualVmiNodeAffinity)
+
+	expectedNodeAffinity := vmi.NewNodeAffinity(nodeName)
+	assert.Equal(t, expectedNodeAffinity, actualVmiNodeAffinity)
+}
+
+func assertVmiNodeAffinityNotExist(t *testing.T, testClient *clientStub, name string) {
+	actualVmi, err := testClient.GetVirtualMachineInstance(testNamespace, name)
+	assert.NoError(t, err)
+
+	assert.Nil(t, actualVmi.Spec.Affinity.NodeAffinity)
 }
 
 func newTestNetAttachDef(cniPluginName string) *netattdefv1.NetworkAttachmentDefinition {
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
@@ -49,12 +49,9 @@ const (
 func TestCheckupSetupShouldFailWhen(t *testing.T) {
 	t.Run("NetworkAttachmentDefinition does not exist", func(t *testing.T) {
 		expectedError := errors.New("get netAttachDef test error")
-		testCheckup := checkup.New(
-			&clientStub{failGetNetAttachDef: expectedError},
-			testNamespace,
-			newTestsCheckupParameters(),
-			&checkerStub{},
-		)
+		testClient := newTestClient()
+		testClient.failGetNetAttachDef = expectedError
+		testCheckup := checkup.New(testClient, testNamespace, newTestsCheckupParameters(), &checkerStub{})
 
 		assert.NoError(t, testCheckup.Preflight())
 		assert.ErrorContains(t, testCheckup.Setup(context.Background()), expectedError.Error())
@@ -62,13 +59,10 @@ func TestCheckupSetupShouldFailWhen(t *testing.T) {
 
 	t.Run("failed to create a VM", func(t *testing.T) {
 		expectedError := errors.New("vmi create test error")
-		testClient := &clientStub{failCreateVmi: expectedError, returnNetAttachDef: &netattdefv1.NetworkAttachmentDefinition{}}
-		testCheckup := checkup.New(
-			testClient,
-			testNamespace,
-			newTestsCheckupParameters(),
-			&checkerStub{},
-		)
+		testClient := newTestClient()
+		testClient.failCreateVmi = expectedError
+		testClient.returnNetAttachDef = &netattdefv1.NetworkAttachmentDefinition{}
+		testCheckup := checkup.New(testClient, testNamespace, newTestsCheckupParameters(), &checkerStub{})
 
 		assert.NoError(t, testCheckup.Preflight())
 		assert.ErrorContains(t, testCheckup.Setup(context.Background()), expectedError.Error())
@@ -76,12 +70,10 @@ func TestCheckupSetupShouldFailWhen(t *testing.T) {
 
 	t.Run("VMs were not ready before timeout expiration", func(t *testing.T) {
 		expectedError := errors.New("timed out")
-		testCheckup := checkup.New(
-			&clientStub{failGetVmi: expectedError, returnNetAttachDef: &netattdefv1.NetworkAttachmentDefinition{}},
-			testNamespace,
-			newTestsCheckupParameters(),
-			&checkerStub{},
-		)
+		testClient := newTestClient()
+		testClient.failGetVmi = expectedError
+		testClient.returnNetAttachDef = &netattdefv1.NetworkAttachmentDefinition{}
+		testCheckup := checkup.New(testClient, testNamespace, newTestsCheckupParameters(), &checkerStub{})
 
 		testCtx, cancel := context.WithTimeout(context.Background(), testTimeout)
 		defer cancel()
@@ -93,13 +85,10 @@ func TestCheckupSetupShouldFailWhen(t *testing.T) {
 
 func TestCheckupTeardownShouldFailWhen(t *testing.T) {
 	t.Run("failed to delete a VM", func(t *testing.T) {
-		testClient := &clientStub{returnNetAttachDef: &netattdefv1.NetworkAttachmentDefinition{}}
-		testCheckup := checkup.New(
-			testClient,
-			testNamespace,
-			newTestsCheckupParameters(),
-			&checkerStub{},
-		)
+		testClient := newTestClient()
+		testClient.returnNetAttachDef = &netattdefv1.NetworkAttachmentDefinition{}
+		testCheckup := checkup.New(testClient, testNamespace, newTestsCheckupParameters(), &checkerStub{})
+
 		testCtx, cancel := context.WithTimeout(context.Background(), testTimeout)
 		defer cancel()
 
@@ -113,13 +102,9 @@ func TestCheckupTeardownShouldFailWhen(t *testing.T) {
 	})
 
 	t.Run("VMs were not disposed before timeout expiration", func(t *testing.T) {
-		testClient := &clientStub{returnNetAttachDef: &netattdefv1.NetworkAttachmentDefinition{}}
-		testCheckup := checkup.New(
-			testClient,
-			testNamespace,
-			newTestsCheckupParameters(),
-			&checkerStub{},
-		)
+		testClient := newTestClient()
+		testClient.returnNetAttachDef = &netattdefv1.NetworkAttachmentDefinition{}
+		testCheckup := checkup.New(testClient, testNamespace, newTestsCheckupParameters(), &checkerStub{})
 
 		assert.NoError(t, testCheckup.Preflight())
 		assert.NoError(t, testCheckup.Setup(context.Background()))
@@ -156,15 +141,9 @@ func TestCheckupSetupShouldCreateVMsWith(t *testing.T) {
 	}
 	for _, testCase := range testsCases {
 		t.Run(testCase.description, func(t *testing.T) {
-			testClient := &clientStub{
-				returnNetAttachDef: testCase.netAttachDef,
-			}
-			testCheckup := checkup.New(
-				testClient,
-				testNamespace,
-				newTestsCheckupParameters(),
-				&checkerStub{},
-			)
+			testClient := newTestClient()
+			testClient.returnNetAttachDef = testCase.netAttachDef
+			testCheckup := checkup.New(testClient, testNamespace, newTestsCheckupParameters(), &checkerStub{})
 
 			assert.NoError(t, testCheckup.Preflight())
 			assert.NoError(t, testCheckup.Setup(context.Background()))
@@ -196,6 +175,10 @@ func newTestsCheckupParameters() config.CheckupParameters {
 	}
 }
 
+func newTestClient() *clientStub {
+	return &clientStub{createdVmis: map[string]*kvcorev1.VirtualMachineInstance{}}
+}
+
 type clientStub struct {
 	createdVmis map[string]*kvcorev1.VirtualMachineInstance
 
@@ -214,10 +197,6 @@ func (c *clientStub) GetVirtualMachineInstance(_, name string) (*kvcorev1.Virtua
 func (c *clientStub) CreateVirtualMachineInstance(_ string, v *kvcorev1.VirtualMachineInstance) (*kvcorev1.VirtualMachineInstance, error) {
 	if c.failCreateVmi != nil {
 		return nil, c.failCreateVmi
-	}
-
-	if c.createdVmis == nil {
-		c.createdVmis = map[string]*kvcorev1.VirtualMachineInstance{}
 	}
 
 	v.Status.Conditions = append(v.Status.Conditions, kvcorev1.VirtualMachineInstanceCondition{
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
@@ -104,9 +104,6 @@ func TestCheckupTeardownShouldFailWhen(t *testing.T) {
 		defer cancel()
 
 		assert.NoError(t, testCheckup.Preflight())
-
-		testClient.returnVmi = newTestReadyVmi()
-
 		assert.NoError(t, testCheckup.Setup(testCtx))
 
 		expectedErr := errors.New("delete vmi test error")
@@ -125,9 +122,6 @@ func TestCheckupTeardownShouldFailWhen(t *testing.T) {
 		)
 
 		assert.NoError(t, testCheckup.Preflight())
-
-		testClient.returnVmi = newTestReadyVmi()
-
 		assert.NoError(t, testCheckup.Setup(context.Background()))
 
 		testCtx, cancel := context.WithTimeout(context.Background(), testTimeout)
@@ -164,7 +158,6 @@ func TestCheckupSetupShouldCreateVMsWith(t *testing.T) {
 		t.Run(testCase.description, func(t *testing.T) {
 			testClient := &clientStub{
 				returnNetAttachDef: testCase.netAttachDef,
-				returnVmi:          newTestReadyVmi(),
 			}
 			testCheckup := checkup.New(
 				testClient,
@@ -182,17 +175,6 @@ func TestCheckupSetupShouldCreateVMsWith(t *testing.T) {
 			}
 		})
 	}
-}
-
-func newTestReadyVmi() *kvcorev1.VirtualMachineInstance {
-	return &kvcorev1.VirtualMachineInstance{Status: kvcorev1.VirtualMachineInstanceStatus{
-		Conditions: []kvcorev1.VirtualMachineInstanceCondition{
-			{
-				Type:   kvcorev1.VirtualMachineInstanceAgentConnected,
-				Status: k8scorev1.ConditionTrue,
-			},
-		},
-	}}
 }
 
 func newTestNetAttachDef(cniPluginName string) *netattdefv1.NetworkAttachmentDefinition {
@@ -215,8 +197,8 @@ func newTestsCheckupParameters() config.CheckupParameters {
 }
 
 type clientStub struct {
-	returnVmi          *kvcorev1.VirtualMachineInstance
-	createdVmis        []*kvcorev1.VirtualMachineInstance
+	createdVmis map[string]*kvcorev1.VirtualMachineInstance
+
 	returnNetAttachDef *netattdefv1.NetworkAttachmentDefinition
 
 	failGetNetAttachDef error
@@ -225,13 +207,27 @@ type clientStub struct {
 	failDeleteVmi       error
 }
 
-func (c *clientStub) GetVirtualMachineInstance(_, _ string) (*kvcorev1.VirtualMachineInstance, error) {
-	return c.returnVmi, c.failGetVmi
+func (c *clientStub) GetVirtualMachineInstance(_, name string) (*kvcorev1.VirtualMachineInstance, error) {
+	return c.createdVmis[name], c.failGetVmi
 }
 
 func (c *clientStub) CreateVirtualMachineInstance(_ string, v *kvcorev1.VirtualMachineInstance) (*kvcorev1.VirtualMachineInstance, error) {
-	c.createdVmis = append(c.createdVmis, v)
-	return v, c.failCreateVmi
+	if c.failCreateVmi != nil {
+		return nil, c.failCreateVmi
+	}
+
+	if c.createdVmis == nil {
+		c.createdVmis = map[string]*kvcorev1.VirtualMachineInstance{}
+	}
+
+	v.Status.Conditions = append(v.Status.Conditions, kvcorev1.VirtualMachineInstanceCondition{
+		Type:   kvcorev1.VirtualMachineInstanceAgentConnected,
+		Status: k8scorev1.ConditionTrue,
+	})
+
+	c.createdVmis[v.Name] = v
+
+	return v, nil
 }
 
 func (c *clientStub) DeleteVirtualMachineInstance(_, _ string) error {
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/config/config_test.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/config/config_test.go
@@ -42,6 +42,8 @@ func TestCreateConfigFromEnvShould(t *testing.T) {
 		testNetAttachDefName              = "blue-net"
 		testDesiredMaxLatencyMilliseconds = 100
 		testSampleDurationSeconds         = 60
+		testSourceNodeName                = "worker1"
+		testTargetNodeName                = "worker2"
 	)
 
 	testCases := []configCreateTestCases{
@@ -80,6 +82,30 @@ func TestCreateConfigFromEnvShould(t *testing.T) {
 					NetworkAttachmentDefinitionName:      testNetAttachDefName,
 					NetworkAttachmentDefinitionNamespace: testNamespace,
 					SampleDurationSeconds:                testSampleDurationSeconds,
+				},
+				ResultsConfigMapName:      testResultConfigMapName,
+				ResultsConfigMapNamespace: testNamespace,
+			},
+		},
+		{
+			"set source and target nodes when both are specified",
+			map[string]string{
+				config.ResultsConfigMapNameEnvVarName:      testResultConfigMapName,
+				config.ResultsConfigMapNamespaceEnvVarName: testNamespace,
+				config.NetworkNameEnvVarName:               testNetAttachDefName,
+				config.NetworkNamespaceEnvVarName:          testNamespace,
+				config.SampleDurationSecondsEnvVarName:     fmt.Sprintf("%d", testSampleDurationSeconds),
+				config.SourceNodeNameEnvVarName:            testSourceNodeName,
+				config.TargetNodeNameEnvVarName:            testTargetNodeName,
+			},
+			config.Config{
+				CheckupParameters: config.CheckupParameters{
+					DesiredMaxLatencyMilliseconds:        config.DefaultDesiredMaxLatencyMilliseconds,
+					NetworkAttachmentDefinitionName:      testNetAttachDefName,
+					NetworkAttachmentDefinitionNamespace: testNamespace,
+					SampleDurationSeconds:                testSampleDurationSeconds,
+					SourceNodeName:                       testSourceNodeName,
+					TargetNodeName:                       testTargetNodeName,
 				},
 				ResultsConfigMapName:      testResultConfigMapName,
 				ResultsConfigMapNamespace: testNamespace,
@@ -188,6 +214,63 @@ func TestCreateConfigFromEnvShouldFailWhen(t *testing.T) {
 				config.ResultsConfigMapNamespaceEnvVarName: "default",
 				config.NetworkNameEnvVarName:               "blue-net",
 				config.NetworkNamespaceEnvVarName:          "",
+			},
+		},
+	}
+	for _, testCase := range testCases {
+		t.Run(testCase.description, func(t *testing.T) {
+			_, err := config.New(testCase.env)
+			assert.Equal(t, err, testCase.expectedError)
+		})
+	}
+}
+
+func TestCreateConfigFromEnvShouldFailWhenNodeNames(t *testing.T) {
+	testCases := []configCreateFallingTestCases{
+		{
+			"source node name is set but target node name isn't",
+			config.ErrTargetNodeNameMissing,
+			map[string]string{
+				config.ResultsConfigMapNameEnvVarName:      "results",
+				config.ResultsConfigMapNamespaceEnvVarName: "default",
+				config.NetworkNameEnvVarName:               "blue-net",
+				config.NetworkNamespaceEnvVarName:          "default",
+				config.SourceNodeNameEnvVarName:            "worker1",
+			},
+		},
+		{
+			"target node name is set but source node name isn't",
+			config.ErrSourceNodeNameMissing,
+			map[string]string{
+				config.ResultsConfigMapNameEnvVarName:      "results",
+				config.ResultsConfigMapNamespaceEnvVarName: "default",
+				config.NetworkNameEnvVarName:               "blue-net",
+				config.NetworkNamespaceEnvVarName:          "default",
+				config.TargetNodeNameEnvVarName:            "worker2",
+			},
+		},
+		{
+			"source node name is empty",
+			config.ErrInvalidSourceNodeName,
+			map[string]string{
+				config.ResultsConfigMapNameEnvVarName:      "results",
+				config.ResultsConfigMapNamespaceEnvVarName: "default",
+				config.NetworkNameEnvVarName:               "blue-net",
+				config.NetworkNamespaceEnvVarName:          "default",
+				config.SourceNodeNameEnvVarName:            "",
+				config.TargetNodeNameEnvVarName:            "worker2",
+			},
+		},
+		{
+			"target node name is empty",
+			config.ErrInvalidTargetNodeName,
+			map[string]string{
+				config.ResultsConfigMapNameEnvVarName:      "results",
+				config.ResultsConfigMapNamespaceEnvVarName: "default",
+				config.NetworkNameEnvVarName:               "blue-net",
+				config.NetworkNamespaceEnvVarName:          "default",
+				config.SourceNodeNameEnvVarName:            "worker1",
+				config.TargetNodeNameEnvVarName:            "",
 			},
 		},
 	}
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/launcher/launcher_test.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/launcher/launcher_test.go
@@ -244,7 +244,8 @@ func (c *fakeClient) GetVirtualMachineInstance(namespace, name string) (*kvcorev
 }
 
 // CreateVirtualMachineInstance adds the given VMI to the VMI tracker,
-// and adds VirtualMachineInstanceReady condition to the VMI status.
+// adds VirtualMachineInstanceAgentConnected condition to the VMI status and
+// the node name according to node affinity rule with 'kubernetes.io/hostname' label selector.
 func (c *fakeClient) CreateVirtualMachineInstance(
 	namespace string,
 	vmi *kvcorev1.VirtualMachineInstance) (*kvcorev1.VirtualMachineInstance, error) {
@@ -254,7 +255,15 @@ func (c *fakeClient) CreateVirtualMachineInstance(
 			Status: k8scorev1.ConditionTrue,
 		},
 	)
-	vmi.Status.NodeName = vmi.Spec.NodeSelector[k8scorev1.LabelHostname]
+
+	if vmi.Spec.Affinity != nil && vmi.Spec.Affinity.NodeAffinity != nil {
+		term := vmi.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms[0]
+		req := term.MatchExpressions[0]
+		if req.Key == k8scorev1.LabelHostname {
+			vmi.Status.NodeName = req.Values[0]
+		}
+	}
+
 	c.vmiTracker[vmiKey(namespace, vmi.Name)] = vmi
 	return vmi, nil
 }
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/vmi/affinity.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/vmi/affinity.go
@@ -1,0 +1,76 @@
+/*
+ * This file is part of the kiagnose project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Copyright 2022 Red Hat, Inc.
+ *
+ */
+
+package vmi
+
+import (
+	k8scorev1 "k8s.io/api/core/v1"
+	k8smetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	kvcorev1 "kubevirt.io/api/core/v1"
+)
+
+// WithAffinity adds the given affinity.
+func WithAffinity(affinity *k8scorev1.Affinity) Option {
+	return func(vmi *kvcorev1.VirtualMachineInstance) {
+		if affinity != nil {
+			vmi.Spec.Affinity = affinity
+		}
+	}
+}
+
+// NewNodeAffinity returns new node affinity with node selector of the given node name.
+// Adding it to a VMI will make sure it will schedule on the given node name.
+func NewNodeAffinity(nodeName string) *k8scorev1.NodeAffinity {
+	req := k8scorev1.NodeSelectorRequirement{
+		Key:      k8scorev1.LabelHostname,
+		Operator: k8scorev1.NodeSelectorOpIn,
+		Values:   []string{nodeName},
+	}
+	term := []k8scorev1.NodeSelectorTerm{
+		{
+			MatchExpressions: []k8scorev1.NodeSelectorRequirement{req},
+		},
+	}
+	return &k8scorev1.NodeAffinity{
+		RequiredDuringSchedulingIgnoredDuringExecution: &k8scorev1.NodeSelector{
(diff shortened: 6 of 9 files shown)


orelmisan:
@ormergi Does this PR replaces PR #153?

ormergi:
> @ormergi Does this PR replaces PR #153?

@orelmisan yes

orelmisan:
> This PR depends on https://github.com/kiagnose/kiagnose/pull/152 (first commit), it enables e2e testing of this PR changes.

PR #152 has two commits. I guess it was split after this PR was opened, is that assumption correct?

orelmisan:
I now see that the last commit of this PR is the second commit of PR #152.
Can you please rebase this PR on PR #152?

ormergi:
> I now see that the last commit of this PR is the second commit of PR #152. Can you please rebase this PR on PR #152?

It is a mistake #152 should have had only one commit (w/o [vmlatency, github jobs: Use multi node cluster](https://github.com/kiagnose/kiagnose/pull/152/commits/8685b4f108422601b804c81d54c0e4dada7fca64)).
I didn't rebase this PR on top of #152 so it won't be blocked by it.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
```diff
@@ -225,13 +208,30 @@ type clientStub struct {
...
 func (c *clientStub) CreateVirtualMachineInstance(_ string, v *kvcorev1.VirtualMachineInstance) (*kvcorev1.VirtualMachineInstance, error) {
-	c.createdVmis = append(c.createdVmis, v)
-	return v, c.failCreateVmi
+	if c.failCreateVmi != nil {
+		return nil, c.failCreateVmi
+	}
+
+	if c.createdVmis == nil {
+		c.createdVmis = map[string]*kvcorev1.VirtualMachineInstance{}
+	}
```
orelmisan, line 270: Please consider creating a factory function for this type that will contain (among others) maps initialization:
```go
func newClientStub(...) *ClientStub {
    return &ClientStub {
        createdVmis: map[string]*kvcorev1.VirtualMachineInstance{},
        ...
    }
}
```

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
```diff
@@ -215,8 +197,9 @@ func newTestsCheckupParameters() config.CheckupParameters {
 }
 
 type clientStub struct {
-	returnVmi          *kvcorev1.VirtualMachineInstance
-	createdVmis        []*kvcorev1.VirtualMachineInstance
+	createdVmis    map[string]*kvcorev1.VirtualMachineInstance
+	vmiNodeNameMap map[string]string
```
orelmisan, line 249: Where is this map initialized?

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -25,6 +25,8 @@ import (
 	"strings"
 	"time"
 
+	k8scorev1 "k8s.io/api/core/v1"
```
orelmisan, line 28:
```suggestion
	corev1 "k8s.io/api/core/v1"
```

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -88,34 +91,67 @@ func (c *checkup) Setup(ctx context.Context) error {
...
+	return nil
+}
 
-	if c.targetVM, err = vmi.WaitUntilReady(waitCtx, c.client, c.namespace, targetVmi.Name); err != nil {
-		return fmt.Errorf("%s: %v", errMessagePrefix, err)
+func newSourceLatencyCheckVmi(
+	name string,
+	sourceNodeName string,
+	networkName string, netAttachDef *netattdefv1.NetworkAttachmentDefinition,
+	macAddress, cidr string) *kvcorev1.VirtualMachineInstance {
```
orelmisan, line 116: This function signature is quite long. Please consider using a struct instead of a long arguments list.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -88,34 +91,67 @@ func (c *checkup) Setup(ctx context.Context) error {
...
-	if c.sourceVM, err = vmi.WaitUntilReady(waitCtx, c.client, c.namespace, sourceVmi.Name); err != nil {
-		return fmt.Errorf("%s: %v", errMessagePrefix, err)
+	return newLatencyCheckVmi(name, sourceNodeSelectorReq, networkName, netAttachDef, macAddress, cidr)
+}
+
+func newTargetLatencyCheckVmi(
+	name string,
+	targetNodeName, sourceNodeName string,
+	networkName string, netAttachDef *netattdefv1.NetworkAttachmentDefinition,
+	macAddress, cidr string) *kvcorev1.VirtualMachineInstance {
```
orelmisan, line 133: Same.

.github/workflows/checkup-kubevirt-vm-latency.check.yaml
```diff
@@ -90,7 +90,7 @@ jobs:
         working-directory: ./checkups/kubevirt-vm-latency
         run: ./automation/make.sh --build-checkup --build-checkup-image
       - name: Start cluster
-        run: ./automation/make.sh --e2e -- --install-kind --install-kubectl --create-cluster
+        run: ./automation/make.sh --e2e -- --install-kind --install-kubectl --create-multi-node-cluster;
```
orelmisan, line 93:
```suggestion
        run: ./automation/make.sh --e2e -- --install-kind --install-kubectl --create-multi-node-cluster
```

orelmisan (commented):

orelmisan (commented):
Please consider splitting the [3rd commit](https://github.com/kiagnose/kiagnose/pull/155/commits/dc8f63a9394d45f8f26a6108e20704260fa81fc1) to the changes made in `StartAndWaitUntilReady()`, and the affinity/anti-affinity logic.
IMO it could help the reviewers.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
ormergi: If you don't mind I prefer to do it on a follow-up PR.

.github/workflows/checkup-kubevirt-vm-latency.check.yaml
ormergi: Note that adding `;` at the end of the line is not a must.

orelmisan: I've suggested to remove the semi-colon `;`.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
ormergi: I didn't define such function because it looked a bit odd in tests IMO.
It required changing the returned instance (of` clientStub{}`) in order to inject predefined objects/errors to the client, for example:
```
testClient := newClientStub() 
testClient.returnNetAttachDef = testNet
...
```
I also thought about defining newClientStub() to receive parameters, but it also looks odd because most parameters are not required for all tests, it will end up with:
```
testClient := newClientStub(nil, nil, nil, nil, testNet, nil)
```
I don't mind doing what you suggest, let me know if you still want it.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
ormergi: At CreateVirtualMachineInstance.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
orelmisan: Is there a common initialization to most tests?

ormergi:
Changes: [[1]](https://github.com/kiagnose/kiagnose/compare/e93b796bf8b2045913d2012532a98ca8eb05c0d6..93c5ecfd5c3f85ce40a3c5d8953dcd55d8e1d967)  resolve comment https://github.com/kiagnose/kiagnose/pull/155#discussion_r962332028

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
ormergi: Done

ormergi:
Changes: [[1]](https://github.com/kiagnose/kiagnose/compare/93c5ecfd5c3f85ce40a3c5d8953dcd55d8e1d967..66ff5abe2806c557958b71ec2acb5e2705c6a7c2) remove redundent smeicolon at vmlatnecy e2e github job.

.github/workflows/checkup-kubevirt-vm-latency.check.yaml
ormergi: Done

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/66ff5abe2806c557958b71ec2acb5e2705c6a7c2..6f972778f7cd3c53fd6f330e0ec44c3f7e77dbcc):
- Add pod anti-affinity to the target VM to ensure it won't be scheduled on the same node as the source VM (unlike the previous impl which waited for the source VM to be ready before creating the target VM)
@orelmisan @EdDev

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -114,8 +128,7 @@ func (c *checkup) Setup(ctx context.Context) error {
 }
 
 func newLatencyCheckVmi(
-	name,
-	nodeName,
+	name string, label map[string]string,
```
RamLavi, line 131: shouldn't `nodeName` be also a parameter?
I mean, the user can still manually choose to use the same node. no?

ormergi: It is handled here https://github.com/kiagnose/kiagnose/pull/155/commits/69980a6ff5c2082f84f7f7952aa8afc35acba57f#diff-d60aa090110dbe4421acf2db6a61ce9c88ae82264bd0e66ddc999917378b3c22R97-R109

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/spec.go
```diff
@@ -156,6 +155,29 @@ func newContainerDiskVolume(name, image string) kvcorev1.Volume {
...
+// WithNodeAffinity adds node affinity rule of match-expression kind with the given requirement.
+func WithNodeAffinity(nodeSelectorRequirement *k8scorev1.NodeSelectorRequirement) Option {
+	return func(vmi *kvcorev1.VirtualMachineInstance) {
+		if nodeSelectorRequirement == nil {
+			return
+		}
+
+		term := []k8scorev1.NodeSelectorTerm{
+			{
+				MatchExpressions: []k8scorev1.NodeSelectorRequirement{*nodeSelectorRequirement},
```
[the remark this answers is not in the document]
ormergi: Done

orelmisan:
PR description:
> Currently, the source and target node may end up scheduled on
the same node.

Suggestion:
Currently, if both the source and the target node are not defined by the user, they may be scheduled on the same node.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
orelmisan: In an offline discussion from yesterday, you've told me that it is initialized in another place.
Where was it?

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
```diff
@@ -49,39 +49,31 @@ const (
...
 	t.Run("NetworkAttachmentDefinition does not exist", func(t *testing.T) {
 		expectedError := errors.New("get netAttachDef test error")
-		testCheckup := checkup.New(
-			&clientStub{failGetNetAttachDef: expectedError},
-			testNamespace,
-			newTestsCheckupParameters(),
-			&checkerStub{},
-		)
+		testClient := newTestClient()
+		testClient.failGetNetAttachDef = expectedError
```
orelmisan, line 54: Please consider adding a method to `clientStub`:
```go
func (cs *clientStub) injectGetNADError(expectedError err) {
   cs.failGetNetAttachDef = expectedError
}
```

And then the call in this line could be:
```go
testClient.injectGetNADError(expectedError)
```
IMO it would improve the readability of the test, and be similar to the style used in the framework's `checkup_test.go`.

Same for similar cases bellow.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
RamLavi: yes, but if names are specified, do we still want to set the labels?

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -66,17 +66,23 @@ func (c *checkup) Preflight() error {
 	return nil
 }
 
+const (
+	LabelLatencyCheckEnd  = "latency-check/end"
```
orelmisan, line 70:
```suggestion
	LatencyCheckupLabelPrefix  = "latency-checkup/end"
```

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -66,17 +66,23 @@ func (c *checkup) Preflight() error {
...
+	LatencyCheckSourceEnd = "source"
```
orelmisan, line 71:
```suggestion
	LatencyCheckupSourceLabelValue = "source"
```

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -88,24 +94,32 @@ func (c *checkup) Setup(ctx context.Context) error {
 		return fmt.Errorf("%s: %v", errMessagePrefix, err)
 	}
 
-	sourceVmi := newLatencyCheckVmi(sourceVmiName, c.params.SourceNodeName, networkName, netAttachDef, sourceVmiMac, sourceVmiCidr)
-	targetVmi := newLatencyCheckVmi(targetVmiName, c.params.TargetNodeName, networkName, netAttachDef, targetVmiMac, targetVmiCidr)
+	sourceVmiLabel := map[string]string{LabelLatencyCheckEnd: LatencyCheckSourceEnd}
```
orelmisan, line 97: Please consider hiding this detail in a function:
```go
func sourceVMILabels() map[string]string {
    return map[string]string{LabelLatencyCheckEnd: LatencyCheckSourceEnd}
}
```

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -88,24 +94,32 @@ func (c *checkup) Setup(ctx context.Context) error {
...
+	sourceVmi := newLatencyCheckVmi(SourceVmiName, sourceVmiLabel, networkName, netAttachDef, sourceVmiMac, sourceVmiCidr)
+	if c.params.SourceNodeName != "" {
+		vmi.AddNodeAffinityRule(sourceVmi, c.params.SourceNodeName)
+	}
+
+	targetVmiLabel := map[string]string{LabelLatencyCheckEnd: LatencyCheckTargetEnd}
```
orelmisan, line 103: Same.

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/spec.go
```diff
@@ -59,16 +58,16 @@ func withTerminationGracePeriodSecond(duration int64) Option {
...
 }
 
-// WithNodeSelector ensures that the VMI gets scheduled on the specified node.
-func WithNodeSelector(nodeName string) Option {
+// WithLabels adds the given labels.
+func WithLabels(labels map[string]string) Option {
 	return func(vmi *kvcorev1.VirtualMachineInstance) {
-		if nodeName == "" {
-			return
+		if len(vmi.ObjectMeta.Labels) < 1 {
```
orelmisan, line 64: Will this snippet work?

```suggestion
		if vmi.ObjectMeta.Labels == nil {
```

orelmisan (commented):
Thank you for the changes @ormergi, it looks great.
What will happen in case `targetNode` is specified but `sourceNode` doesn't?

WDYT about disabling the option to define only one of them?
This way we would have only two options:
1. They are both defined.
2. They are both undefined.

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/spec.go
ormergi: It won't cover the case where `vmi.ObjectMeta.Labels` is initialized but empty, this is why I think its better to use `len` for such checks.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
ormergi: I followed k8s label naming `LabelXYZ` I think its a better fit.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
ormergi: `createdVmis` is now initialised at `newTestClient()`.
`vmiNodeNameMap` is no longer needed, I will remove it.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
ormergi: In general I agree to make self-explanatory funcs, in this case, I think it won't add much more than the details there are.
In any case, refactoring this test suite is not the scope of this PR, we can do it in a follow-up if needed.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
ormergi: It was removed because it was no longer needed.

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/spec.go
ormergi: Done

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
ormergi: Done

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/5b1dc43b6f4fe57df0f726080738fff3f9f51f8e..439fa2c6c5fdb8dc629e66a9a61492d8f350aaef): resolve https://github.com/kiagnose/kiagnose/pull/155#discussion_r963658105 and https://github.com/kiagnose/kiagnose/pull/155#discussion_r963659414

ormergi:
@orelmisan I pushed new changes, please take another look.

> What will happen in case targetNode is specified but sourceNode doesn't?

The target VM will be created at the specified target node (it will have node affinity added to its spec).

> WDYT about disabling the option to define only one of them?
> This way we would have only two options:
> 1. They are both defined.
> 2. They are both undefined.

At this point I think the granularity we have today has value, it may come in handy in the scenario one would like to check the latency between a specific node to another arbitrary node in the cluster.

orelmisan:
Thank you for the answer @ormergi

orelmisan:
> vmlatency, checkup: Ensure VMs get scheduled on diffrent nodes

`diffrent` => `different`

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/spec.go
```diff
@@ -59,16 +58,20 @@ func withTerminationGracePeriodSecond(duration int64) Option {
 	}
 }
 
-// WithNodeSelector ensures that the VMI gets scheduled on the specified node.
-func WithNodeSelector(nodeName string) Option {
+type Label struct {
+	Key, Value string
+}
```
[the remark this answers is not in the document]
ormergi: When creating the Affinity object it is much easier to consume `Label` instead of `map[string]string`, it also enables exposing only SourceVmiLabel() for the tests instead of exposing also the latency-check-end the label key and value.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -88,24 +101,31 @@ func (c *checkup) Setup(ctx context.Context) error {
...
+	sourceVmi := newLatencyCheckVmi(SourceVmiName, sourceVmiLabel, networkName, netAttachDef, sourceVmiMac, sourceVmiCidr)
+	if c.params.SourceNodeName != "" {
+		vmi.AddNodeAffinityRule(sourceVmi, c.params.SourceNodeName)
+	}
+
+	targetVmi := newLatencyCheckVmi(TargetVmiName, targetVmiLabel(), networkName, netAttachDef, targetVmiMac, targetVmiCidr)
+	if c.params.TargetNodeName != "" {
+		vmi.AddNodeAffinityRule(targetVmi, c.params.TargetNodeName)
+	} else {
+		vmi.AddPodAntiAffinityRule(targetVmi, sourceVmiLabel)
```
RamLavi, line 114: is the affinity rules on both vmis required in order for the antiaffinity to work? is not - then could it be on another commit?

RamLavi, line 114: what happens if TargetNodeName==SourceNodeName set by the user? will the anti affinity fail to schedule the vmis?

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -66,17 +66,30 @@ func (c *checkup) Preflight() error {
...
 
+const (
+	SourceVmiName         = "latency-check-source"
+	TargetVmiName         = "latency-check-target"
+	labelLatencyCheckEnd  = "latency-check/end"
+	latencyCheckSourceEnd = "source"
+	latencyCheckTargetEnd = "target"
+)
+
+func SourceVmiLabel() vmi.Label {
```
RamLavi, line 77: SourceVmiLabel and targetVmiLabel are rather general names, but we insert very specific labels to it. Can you change the name to be more specific, or pass the data as params to a real general function?

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -88,24 +101,31 @@ func (c *checkup) Setup(ctx context.Context) error {
...
 
-	sourceVmi := newLatencyCheckVmi(sourceVmiName, c.params.SourceNodeName, networkName, netAttachDef, sourceVmiMac, sourceVmiCidr)
-	targetVmi := newLatencyCheckVmi(targetVmiName, c.params.TargetNodeName, networkName, netAttachDef, targetVmiMac, targetVmiCidr)
+	sourceVmiLabel := SourceVmiLabel()
+	sourceVmi := newLatencyCheckVmi(SourceVmiName, sourceVmiLabel, networkName, netAttachDef, sourceVmiMac, sourceVmiCidr)
+	if c.params.SourceNodeName != "" {
+		vmi.AddNodeAffinityRule(sourceVmi, c.params.SourceNodeName)
+	}
+
+	targetVmi := newLatencyCheckVmi(TargetVmiName, targetVmiLabel(), networkName, netAttachDef, targetVmiMac, targetVmiCidr)
```
RamLavi, line 110: is it just me or foes `targetVmiLabel` have no  use? if so, why add it?

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
RamLavi, line 114: I'm curious as to what happens if both source and node VMIs don't specify Node name.
Won't there be a sort of race? 
For example, if target VMI gets scheduled first to NODE1, then source VMI gets scheduled to NODE1 as well (it has not anti affinity right?) - then would the target VMI get reassigned? 

if this is the case, we can solve it by adding the AddPodAntiAffinityRule to the sroucevmi as well..

RamLavi (commented):
Added questions

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
ormergi: I added it for completeness I think its odd if there a label that indicates the VM porpuse only on one of them.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -66,17 +66,30 @@ func (c *checkup) Preflight() error {
 	return nil
 }
 
+const (
+	SourceVmiName         = "latency-check-source"
+	TargetVmiName         = "latency-check-target"
+	labelLatencyCheckEnd  = "latency-check/end"
```
RamLavi, line 72: why do we use the term "End" here? it's confusing to me.. I would try to rephrase if possibel

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
ormergi: There won't be a race, the scheduler will consider that there is a pod with `PodAntiAffinity` rule and schedule the other one at a different node.

Following our discussion offline, we concluded in the scenario where only the target Node is specified in the checkup config both VMs might schedule on the same node.
I will change the code to add `PodAntiAffinity` for the source VM when needed.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
ormergi: I used "end" as in networking when you want to describe two endpoints of a network connection.

I thought about using `SourceVmiName` and `TargetVmiName` as the label keys, but it looked odd,
what do you think, do you have other suggestions?

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/12b34368a1ca63f972794f3a30543a12a111a30e..aeaab4a80b27f6986644664e964c44bb2c4f1c03): Add PodAntiAffinity to both VMs when both or one of `sourceNode` and `targetNode` are not specified in the checkup config (when both are specified PodAntiAffinity is not added).

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/aeaab4a80b27f6986644664e964c44bb2c4f1c03..65b21ee264e287106793db95df0b404f585e6f14): Simplify the checkup behavior regarding "source_node" and "target_node" parameters handling.
Source and target node names are now considered only when both are specified.
@orelmisan @RamLavi please take another look.

orelmisan (commented):
Thanks for the changes @ormergi.
Could you please consider splitting the config validation change (in the [third](https://github.com/kiagnose/kiagnose/pull/155/commits/11108e790b942215f642f21d92e42f575e7ac41f) commit) to a commit of its own?
IMO it stands on its own.

There is a typo in the third commit's message:
> With this change specifying only one of "source_node" and "source_node"
is no longer supported.

`source_node` appears twice.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
RamLavi: AFAIK, we checked and agreed that there is no need for multiple anti affinities - only one is needed.

ormergi:
> Thanks for the changes @ormergi. Could you please consider splitting the config validation change (in the [third](https://github.com/kiagnose/kiagnose/pull/155/commits/11108e790b942215f642f21d92e42f575e7ac41f) commit) to a commit of its own? IMO it stands on its own.
> 

Done

> There is a typo in the third commit's message:
> 
> > With this change specifying only one of "source_node" and "source_node"
> > is no longer supported.
> 
> `source_node` appears twice.

Fixed.

orelmisan:
I suggest to swith the order of the [third](https://github.com/kiagnose/kiagnose/pull/155/commits/1f243630e1caf0bb2fc3a4a22ea2f3905e33dfca) and [forth](https://github.com/kiagnose/kiagnose/pull/155/commits/61e27933240e351676012179a5dc01561431000c) commits.

The VM scheduling commit builds on the assumption that the config validation commit is in place.
I also suggest to move the changes to the README file to the config validation commit.

After the change in the config validation, `checkup_test.go` could include only the cases were both or neither source and target nodes are specified.
IIUC, there are still tests when one of them is missing.

ormergi:
> I suggest to switch the order of the [third](https://github.com/kiagnose/kiagnose/pull/155/commits/1f243630e1caf0bb2fc3a4a22ea2f3905e33dfca) and [forth](https://github.com/kiagnose/kiagnose/pull/155/commits/61e27933240e351676012179a5dc01561431000c) commits.
> 
> The VM scheduling commit builds on the assumption that the config validation commit is in place. I also suggest to move the changes to the README file to the config validation commit. 

`checkup` pkg changes does not depend on `config` pkg changes, it could work w/o them, the only change is the UX where the checkup will return an error iff only one of the names is specified.
Regarding the order as you said before both can stand on their own, I think the order doesn't really matter in this case.

> After the change in the config validation, `checkup_test.go` could include only the cases were both or neither source and target nodes are specified. IIUC, there are still tests where one of them is missing.

I will remove the redandent tests.

ormergi:
Changes [[1]](https://github.com/kiagnose/kiagnose/compare/19781861f0bb2667115ef5894725fbbcd92bdc31..10fd46578adafc47a2dbd90aaac6ba680eee2485) [[2]](https://github.com/kiagnose/kiagnose/compare/10fd46578adafc47a2dbd90aaac6ba680eee2485..df44ae22fb86610bca06d3a438a4e8e082e94e29): move readme changes to `config` pkg changes commit , fix typo in the commit message.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
```diff
@@ -225,13 +207,26 @@ type clientStub struct {
...
+
+	if c.createdVmis == nil {
+		c.createdVmis = map[string]*kvcorev1.VirtualMachineInstance{}
+	}
+	c.createdVmis[v.Name] = v
+
+	v.Status.Conditions = append(v.Status.Conditions, kvcorev1.VirtualMachineInstanceCondition{
+		Type:   kvcorev1.VirtualMachineInstanceAgentConnected,
+		Status: k8scorev1.ConditionTrue,
+	})
```
orelmisan, line 284: nit: Please consider making all the changes to the VMI before storing it in the `createdVmis` map.
I find it a bit confusing to see a change to an object that was already stored.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
```diff
@@ -156,6 +157,82 @@ func TestCheckupSetupShouldCreateVMsWith(t *testing.T) {
...
+		assertTargetVmiPodAntiAffinityExist(t, testClient)
+		assertVmiNodeAffinityNotExist(t, testClient, checkup.SourceVmiName)
+		assertVmiNodeAffinityNotExist(t, testClient, checkup.TargetVmiName)
+	})
+}
+
+func TestCheckupSetupShouldCreateVMsWithNodeAffinity(t *testing.T) {
+	t.Run("when both source and target node names are specified", func(t *testing.T) {
+		const testSourceNode = "sourceNode.example.com"
+		const testTargetNode = "targetNode.example.com"
```
orelmisan, line 178: nit:

```suggestion
		const (
		    testSourceNode = "sourceNode.example.com"
		    testTargetNode = "targetNode.example.com"
		)
```

orelmisan (commented):
[This](https://github.com/kiagnose/kiagnose/pull/155#issuecomment-1238055608) comment about the PR description is also applicable to [this](https://github.com/kiagnose/kiagnose/pull/155/commits/d952986e02c19201a03b733ae8ad70330d9bdd8b) commit's message.

[This](https://github.com/kiagnose/kiagnose/pull/155/commits/07150e5ab84631033e8dfac89fece9529639f3ec) commit's message IMO is lacking reasoning.
I suggest to write something about logic and tests simplification by eliminating two out of four permutations of source and destination nodes names.

I still suggest to put the limit of input permutations before the changes to the VMI setup.

ormergi:
Changes [[1]](https://github.com/kiagnose/kiagnose/compare/df44ae22fb86610bca06d3a438a4e8e082e94e29..55224bec5c22f6c56bdea28476ecacfb3ff74538) [[2]](https://github.com/kiagnose/kiagnose/compare/55224bec5c22f6c56bdea28476ecacfb3ff74538..31e0f0c67308fca69f52438c9f0e56fbe5ee33d6): resolve https://github.com/kiagnose/kiagnose/pull/155#pullrequestreview-1104231309 review comments, @orelmisan please take another look.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
ormergi: Done

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
ormergi: Done

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
orelmisan: It was fixed in the [second](https://github.com/kiagnose/kiagnose/pull/155/commits/7eaa3df5b0d69838a767b4c97c8366f0a46b3c10) commit instead of the [first](https://github.com/kiagnose/kiagnose/pull/155/commits/ab141e4b5ad471309670fde87aabba3f2a2fda14) (where it was introduced).

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -88,24 +101,30 @@ func (c *checkup) Setup(ctx context.Context) error {
 		return fmt.Errorf("%s: %v", errMessagePrefix, err)
 	}
 
-	sourceVmi := newLatencyCheckVmi(sourceVmiName, c.params.SourceNodeName, networkName, netAttachDef, sourceVmiMac, sourceVmiCidr)
-	targetVmi := newLatencyCheckVmi(targetVmiName, c.params.TargetNodeName, networkName, netAttachDef, targetVmiMac, targetVmiCidr)
+	sourceVmiLabel := LatencyCheckSourceLabel()
+	targetVmiLabel := LatencyCheckTargetLabel()
```
orelmisan, line 105: I think we can use only one label for both VMs:
`vm-latency-checkup/<checkup-name>` (or something along those lines)

It will make the solution more general, and we won't have to care which VM has which role (source or target).
Furthermore, it will help us to be able to run multiple instances of this checkup simultaneously (currently its not possible #164).

orelmisan (commented):
Thank you for the changes @ormergi.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/31e0f0c67308fca69f52438c9f0e56fbe5ee33d6..63397ed227929ac66f128448ae4ead47c93eceed): resolve https://github.com/kiagnose/kiagnose/pull/155#pullrequestreview-1105257330 comments, @orelmisan please take another look.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
ormergi: https://github.com/kiagnose/kiagnose/pull/165

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
```diff
@@ -156,6 +157,84 @@ func TestCheckupSetupShouldCreateVMsWith(t *testing.T) {
...
+}
+
+func assertTargetVmiPodAntiAffinityExist(t *testing.T, testClient *clientStub) {
+	actualVmi, err := testClient.GetVirtualMachineInstance(testNamespace, checkup.TargetVmiName)
+	assert.NoError(t, err)
+
+	assert.NotNil(t, actualVmi.Spec.Affinity.PodAntiAffinity)
+	expectedPodAntiAffinity := vmi.NewPodAntiAffinity(vmi.Label{Key: checkup.LabelLatencyCheckVM, Value: ""})
+	assert.Equal(t, expectedPodAntiAffinity, actualVmi.Spec.Affinity.PodAntiAffinity)
+}
```
orelmisan, line 211: Is it necessary to have this separation now when we are using a single label?
```suggestion
func assertVmiPodAntiAffinityExist(t *testing.T, testClient *clientStub, vmiName string) {
	actualVmi, err := testClient.GetVirtualMachineInstance(testNamespace, vmiName)
	assert.NoError(t, err)

	assert.NotNil(t, actualVmi.Spec.Affinity.PodAntiAffinity)
	expectedPodAntiAffinity := vmi.NewPodAntiAffinity(vmi.Label{Key: checkup.LabelLatencyCheckVM, Value: ""})
	assert.Equal(t, expectedPodAntiAffinity, actualVmi.Spec.Affinity.PodAntiAffinity)
}
```

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/spec.go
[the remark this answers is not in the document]
orelmisan: Is the `Label` type still needed?

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -88,24 +92,29 @@ func (c *checkup) Setup(ctx context.Context) error {
...
-	targetVmi := newLatencyCheckVmi(targetVmiName, c.params.TargetNodeName, networkName, netAttachDef, targetVmiMac, targetVmiCidr)
+	latencyCheckVMLabel := vmi.Label{Key: LabelLatencyCheckVM, Value: ""}
+	sourceVmi := newLatencyCheckVmi(SourceVmiName, latencyCheckVMLabel, networkName, netAttachDef, sourceVmiMac, sourceVmiCidr)
+	targetVmi := newLatencyCheckVmi(TargetVmiName, latencyCheckVMLabel, networkName, netAttachDef, targetVmiMac, targetVmiCidr)
+	if c.params.SourceNodeName != "" && c.params.TargetNodeName != "" {
+		vmi.AddNodeAffinityRule(sourceVmi, c.params.SourceNodeName)
+		vmi.AddNodeAffinityRule(targetVmi, c.params.TargetNodeName)
+	} else {
+		vmi.AddPodAntiAffinityRule(sourceVmi, latencyCheckVMLabel)
+		vmi.AddPodAntiAffinityRule(targetVmi, latencyCheckVMLabel)
```
RamLavi, line 103: the discussion got erased so I'll ask again here - why do we set 2 pod anti affinities if we saw that 1 is enough?

ormergi: I think we discussed it offline, It does no harm and is required for #165.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
ormergi: Done

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/spec.go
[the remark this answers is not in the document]
ormergi: Yes, for the same reasons I mentioned.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/98fd9ec44a55561cf42750701c680b030f6fb033..69c7c62b7b0b2db0c15367babb2f23c7a45a0782):  resolve https://github.com/kiagnose/kiagnose/pull/155#discussion_r970348497
@orelmisan @RamLavi please take another look.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
RamLavi: I don't remember such a discussion - but I think that if it's not needed then we shouldn't add it. 
Moreover, from a brief review of #165 - I fail to see how it is required there - adding a randomizes label is great but we still don't need 2 AntiAffinity rules - one will suffice cause it affects as 2 way repellent.

orelmisan: I suggest to keep it as-is for now and consider removing one in a future PR.
BTW, the discussion was at the office.

orelmisan:
The PR description is not updated after the transition to a single label.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
EdDev: It is irrelevant to ref offline discussions without elaborating the details.
We should formally reason everything.

Please try and always give a good reason, otherwise it means there is no need for it to exist.
It is sometime hard to express it, I admit.

The reasoning for adding the rules symmetrical goes something like this:
It is an implementation detail that given N pods, only N-1 need to have the rule set.
Moreover, if it is set consistently (and symmetrically), it is simpler to understand, over having only one and then needing to reverse engineer why it works.
I am pretty sure this would not have been raised if we would have used 3 VMI/s and for me, consistency is making things simpler (I do not need to think about other cases, e.g. who was created first, how it works, etc...).

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -88,24 +92,29 @@ func (c *checkup) Setup(ctx context.Context) error {
 		return fmt.Errorf("%s: %v", errMessagePrefix, err)
 	}
 
-	sourceVmi := newLatencyCheckVmi(sourceVmiName, c.params.SourceNodeName, networkName, netAttachDef, sourceVmiMac, sourceVmiCidr)
-	targetVmi := newLatencyCheckVmi(targetVmiName, c.params.TargetNodeName, networkName, netAttachDef, targetVmiMac, targetVmiCidr)
+	latencyCheckVMLabel := vmi.Label{Key: LabelLatencyCheckVM, Value: ""}
```
orelmisan, line 95: Could you please add a note to the commit message that the label's value will be set to a significant value in a follow-up PR?

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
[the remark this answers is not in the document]
EdDev: Ignore this one.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
RamLavi: To bring my offline discussion with @ormergi  to the table:
it is enough to add 1 anti affinity rule for 2 pods, it doesn't matter which was ready first, so I think keeping it simple is best here.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
ormergi: I prefer to not mention #165 changes in that commit message, also the label changes.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
EdDev: I would prefer you respond to the reasoning I provided, as I did feedback to this reasoning you just raised (1 is enough and works, making it "simpler").

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
orelmisan: I don't suggest to give details about the future change, just to note that the empty string will not stay for long.

orelmisan:
PR title "...defualt" should be "default".

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -66,17 +66,21 @@ func (c *checkup) Preflight() error {
 	return nil
 }
 
+const (
+	SourceVmiName       = "latency-check-source"
```
EdDev, line 70: nit: `SourceVMIName`

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/spec.go
```diff
@@ -24,7 +24,6 @@ import (
 	k8scorev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	k8smetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
```
EdDev, line 27: Why this was changed? I find separation between domains is useful.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
ormergi: It's not mandatory to set a label value, it can be empty.

orelmisan: I agree that it can be empty.
But this will not be the final value, right?
We are planning to change it.

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/spec.go
```diff
@@ -88,6 +91,57 @@ func withRng() Option {
 	}
 }
 
+// AddNodeAffinityRule adds node affinity rule which makes the given VMI to schedule at the given node name.
+func AddNodeAffinityRule(vmi *kvcorev1.VirtualMachineInstance, nodeName string) {
```
EdDev, line 95: Most of these helpers are using a different pattern from what existed so far.
Is there a reason for this?

To be clear, the pattern used so far is a builder one.

ormergi:
> PR title "...defualt" should be "default".

Done

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup_test.go
ormergi: Done

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
EdDev: It is something that repeats all over.
It could be fixed for all the source code in one PR as a follow up.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
ormergi: The label was changed to "latency-check-vm" with no value.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
ormergi: This function was removed it is no longer needed since the label has only one permutation, the label name is exposed instead of a func.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
EdDev: If that is the case, you could require to add that commit in this PR.
If it is not in this PR, it means that it may or may not be added, depending if the other PR is convincing or not.

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/spec.go
```diff
@@ -88,6 +91,57 @@ func withRng() Option {
...
+	vmi.Spec.Affinity = &k8scorev1.Affinity{NodeAffinity: NewNodeAffinity(nodeName)}
+}
+
+// NewNodeAffinity returns new node affinity with node selector of the given node name.
+// Adding it to a VMI will make sure it will schedule on the given node name.
+func NewNodeAffinity(nodeName string) *k8scorev1.NodeAffinity {
```
EdDev, line 101: I would also move all the new affinity family helpers to a separate file (e.g. `affinity`)

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
ormergi: The code is changed so there is only one label for both VMs "latency-check-vm".

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/spec.go
ormergi: It was removed by mistake, I will add it.

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/spec.go
ormergi: My intent was to keep the logic that sets the VMs affinity visible at checkup `Setup()`, instead of hiding it in another function or over-complicate the func that creates the VMs for the checkup.

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/spec.go
ormergi: Done

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
ormergi: The fact that it's going to change doesn't mean it should be added to the commit message, it doesn't have value for git history.
It's similar to not mentioning any other changes one would plan to do on every commit.

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/spec.go
EdDev: The cost on the other hand is that the flow structure of how and when things happen got more complicated.
Also, these helpers are mutating the inputted argument, which is IMO a bad behavior.

Creating the VMI once and not mutating it later is a nice rule to keep.
I.e. collect the configuration options, then build the spec once, create the object on the API server, etc...

I am not going to insist on this now, but lets try to have structure so we will not enter too many flows where things get changed in different ways.

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/affinity.go
```diff
@@ -0,0 +1,59 @@
+package vmi
```
EdDev, line 1: Missing header.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
RamLavi: I think my case is hanging over "seen" behavior, and not documented one. as much as I looked, I didn't find a place where it says that anti-affinity is co-dependent, i.e. repels in a two-way fashion. 
@EdDev unless you find otherwise, I cannot insist on using N-1 (i.e. 1) anti-affinity rule.

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/affinity.go
ormergi: Done

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
ormergi: Practically speaking, this checkup is not likely to change in a way it will spin up more than two VMs, this is a latency measurement between two ends.
If for some reason it will have more than two VMs, it is likely that it will require more changes.

I agree it will look more consistent for the next reviewer, but having the affinity rule on both ends is not really needed,
it may confuse other reviewers to think that it's mandatory.
Keep in mind that this checkup should serve as a reference for other checkup authors.

orelmisan (commented):
Thank you for the changes @ormergi.

In the [third](https://github.com/kiagnose/kiagnose/pull/155/commits/497776b05803802d58876b9f6f1b2fdc97aedea6) commit's message:

> Currently, it's possible to specify one or both of 'source_node'
and 'target_node' parameters.

Its also possible to specify neither of them.

In the [forth](https://github.com/kiagnose/kiagnose/pull/155/commits/57eab56b0f378e0bdb9962f92d5e8650ef6c9ba7) commit message:
> The VMs will create with "latency-check-vm" label which indicates

"will create" -> "will be created"

> ... both VMs will be crated with 

"crated" -> "created"

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/spec.go
ormergi: @EdDev in this particular case, would you "hide" the logic that decides which affinity to add inside `newLatnecyChecupVM()`?

orelmisan:
Now after PR #157 was merged, I think its worth adding a little check to the VM latency's e2e tests - to check that the source and target nodes are different.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
EdDev: > This PR introduces another potential issue we need to handle if we'll want to address issue #164, so I think its worth documenting.

I do not understand.
The existing situation is that the other PR is handling the limitation. If it does, what is the purpose to document it in a commit as a "TODO"? Feels an overkill to me.
You you want to require it to exist in the first place, just request to do it in this PR and not in a separate one.
If it is not something that is worth it, then it is not worth also to comment about it.

But even if a comment is added, then the comment should come for the limitation it introduces, however, the limitation is actually masks by other limitation which are stronger, so such a comment will be misleading or confusing.

ormergi:
Changes: 
[[1]](https://github.com/kiagnose/kiagnose/compare/ce38182e123d9c2a90ea05168a0678c1719c6c36..278bee9651bc1450107e72849f7b31db911628e6) @orelmisan  resolve https://github.com/kiagnose/kiagnose/pull/155#pullrequestreview-1107411592 
[[2]](https://github.com/kiagnose/kiagnose/compare/278bee9651bc1450107e72849f7b31db911628e6..f87a55d977e7ab5a5a384ed4d84d3b668cb81f8b) @EdDev changes the way VMs affinity is added to be alined with the current pattern.

checkups/kubevirt-vm-latency/vmlatency/internal/config/config_test.go
```diff
@@ -199,6 +225,63 @@ func TestCreateConfigFromEnvShouldFailWhen(t *testing.T) {
...
+				config.ResultsConfigMapNameEnvVarName:      "results",
+				config.ResultsConfigMapNamespaceEnvVarName: "default",
+				config.NetworkNameEnvVarName:               "blue-net",
+				config.NetworkNamespaceEnvVarName:          "default",
+				config.SourceNodeNameEnvVarName:            "",
+				config.TargetNodeNameEnvVarName:            "worker2",
+			},
+		},
+		{
+			"source node name is empty",
```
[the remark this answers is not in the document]
ormergi: Done

checkups/kubevirt-vm-latency/vmlatency/internal/config/config_test.go
```diff
@@ -85,6 +87,30 @@ func TestCreateConfigFromEnvShould(t *testing.T) {
 				ResultsConfigMapNamespace: testNamespace,
 			},
 		},
+		{
+			"set source and target nodes when both specified",
```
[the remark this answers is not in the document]
ormergi: Done

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
orelmisan: Thank you for the clarification @EdDev.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
ormergi: I understand where this comes from, note that it does not always improve the code readability when the variable name includes one acronym after another, for example, `SourceSRIOVVMI` vs. `SourceSriovVm` (there is no such example in the code base)

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/spec.go
EdDev: The structure of the setup goes like this:
- Fetch the NAD
- Build the VMI spec from various data inputs.
- Start the VMI.
- Wait for it to come up.

In this structure, the affinity fits the VMI spec build. So I guess the answer would be yes, it fits that function. But adding more and more arguments may not be the best solution.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
ormergi: I also meant to the anti-affinity, we are on the same page.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -88,24 +94,30 @@ func (c *checkup) Setup(ctx context.Context) error {
...
-	sourceVmi := newLatencyCheckVmi(sourceVmiName, c.params.SourceNodeName, networkName, netAttachDef, sourceVmiMac, sourceVmiCidr)
-	targetVmi := newLatencyCheckVmi(targetVmiName, c.params.TargetNodeName, networkName, netAttachDef, targetVmiMac, targetVmiCidr)
+	vmLabel := vmi.Label{Key: LabelLatencyCheckVM, Value: ""}
+	var sourceVmiAffinity, targetVmiAffinity *k8scorev1.Affinity
+	if c.params.SourceNodeName != "" && c.params.TargetNodeName != "" {
+		sourceVmiAffinity = &k8scorev1.Affinity{NodeAffinity: vmi.NewNodeAffinity(c.params.SourceNodeName)}
+		targetVmiAffinity = &k8scorev1.Affinity{NodeAffinity: vmi.NewNodeAffinity(c.params.TargetNodeName)}
+	} else {
+		sourceVmiAffinity = &k8scorev1.Affinity{PodAntiAffinity: vmi.NewPodAntiAffinity(vmLabel)}
+		targetVmiAffinity = &k8scorev1.Affinity{PodAntiAffinity: vmi.NewPodAntiAffinity(vmLabel)}
```
EdDev, line 104: Expressing that the affinity is the same means you only need one `VMIAffinity` variable.

ormergi:
[Rebased](https://github.com/kiagnose/kiagnose/compare/f87a55d977e7ab5a5a384ed4d84d3b668cb81f8b..b3ea806fd2d5e09e0b49ff2e0a1b321537da9eeb)

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
ormergi: But its not the same when source and target nodes are specified and NodeAffinity is required.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/b3ea806fd2d5e09e0b49ff2e0a1b321537da9eeb..dd75aedb43cfaad80db2fd42045417f774234862): fix `checkup` pkg unit test after rebasing.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/dd75aedb43cfaad80db2fd42045417f774234862..6996f253552c9a9aad01329520af6e2204625c9f): simplify and move the code that adds affinity to `newLatencyCheckVmi()`

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
ormergi: This code is moved to inside `newLatencyCheckVmi()`

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
EdDev: There are always exceptions, we just try to avoid them as much as possible.
And it comes from a formal recommendation, see [here](https://github.com/golang/go/wiki/CodeReviewComments#initialisms).

checkups/kubevirt-vm-latency/vmlatency/internal/launcher/launcher_test.go
```diff
@@ -254,7 +255,15 @@ func (c *fakeClient) CreateVirtualMachineInstance(
...
 		},
 	)
-	vmi.Status.NodeName = vmi.Spec.NodeSelector[k8scorev1.LabelHostname]
+
+	if vmi.Spec.Affinity != nil && vmi.Spec.Affinity.NodeAffinity != nil {
+		term := vmi.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms[0]
+		req := term.MatchExpressions[0]
+		if req.Key == k8scorev1.LabelHostname {
+			vmi.Status.NodeName = req.Values[0]
+		}
```
orelmisan, line 264: nit:
```suggestion
		if req := term.MatchExpressions[0]; req.Key == k8scorev1.LabelHostname {
			vmi.Status.NodeName = req.Values[0]
		}
```

orelmisan (commented):
Thank you for the changes @ormergi.

[Forth](https://github.com/kiagnose/kiagnose/pull/155/commits/d2bc1b35c117288fefc0cc9063ee72f25251e9cd) commit message:
> Currently, when both the source and the target node are not defined

"node" -> "nodes"

> With this change specifying only one of "source_node" and "target_node"
is no longer supported.

IMO this sentence should move to the [third](https://github.com/kiagnose/kiagnose/pull/155/commits/7cba72302f7299e338d3d775527008123f368711) commit message.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
orelmisan: IMO we could fix it across the project in a follow-up PR.

RamLavi (approved):
Looks great! I like moving the Affinity assigning to the vmi creation.
Regarding the Anti-Affinity for 2 VMIs - I think we might hit a hurdle in the future - but we discussed it greatly and I must admit I did not find a proper documentation in my favor. Moreover I am in the minority party - so I remove my objection..
/lgtm

orelmisan (approved):
Thank you @ormergi!
It looks great :)

checkups/kubevirt-vm-latency/vmlatency/internal/launcher/launcher_test.go
ormergi: I didn't follow this pattern in this case because it would look odd to scope `req` and not also `term`, but scoping `term` makes it less readable.

orelmisan: Thank you for your answer.

EdDev (approved):
Thank you, looks very good!

Raising

PR #156, opened 2022-09-05
vm-latency, reporter: Enhance results tests

RamLavi:
This PR enhances the reporter unit tests to also check conversion of values set on the results configmap

### DIFF
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
@@ -21,7 +21,10 @@ package reporter_test
 
 import (
 	"errors"
+	"fmt"
+	"strconv"
 	"testing"
+	"time"
 
 	assert "github.com/stretchr/testify/require"
 
@@ -51,10 +54,73 @@ func TestReportShouldFailWhen(t *testing.T) {
 	})
 }
 
-type configMapClientStub struct {
-	failUpdateConfigMap error
+func TestReportShouldSuccessfullyConvertResultValues(t *testing.T) {
+	const (
+		someFailureReason      = "some reason"
+		someOtherFailureReason = "some other reason"
+	)
+	testClient := &configMapClientStub{}
+	testReporter := reporter.New(testClient, testNamespace, testConfigMapName)
+
+	t.Run("on checkup successful completion", func(t *testing.T) {
+		checkupStatus := status.Status{
+			FailureReason: []string{},
+			Results: status.Results{
+				MinLatency:          1 * time.Minute,
+				AvgLatency:          2 * time.Minute,
+				MeasurementDuration: 3 * time.Minute,
+				MaxLatency:          4 * time.Minute,
+			},
+		}
+		assert.NoError(t, testReporter.Report(checkupStatus))
+
+		expectedReportData := map[string]string{
+			"status.result.minLatencyNanoSec":      fmt.Sprint(checkupStatus.MinLatency.Nanoseconds()),
+			"status.result.maxLatencyNanoSec":      fmt.Sprint(checkupStatus.MaxLatency.Nanoseconds()),
+			"status.result.avgLatencyNanoSec":      fmt.Sprint(checkupStatus.AvgLatency.Nanoseconds()),
+			"status.result.measurementDurationSec": fmt.Sprint(checkupStatus.MeasurementDuration.Seconds()),
+			"status.succeeded":                     strconv.FormatBool(true),
+			"status.failureReason":                 "",
+		}
+
+		assert.Equal(t, expectedReportData, testClient.configMapData)
+	})
+
+	t.Run("on checkup failure", func(t *testing.T) {
+		checkupStatus := status.Status{
+			FailureReason: []string{someFailureReason},
+		}
+		assert.NoError(t, testReporter.Report(checkupStatus))
+
+		expectedReportData := map[string]string{
+			"status.succeeded":     strconv.FormatBool(false),
+			"status.failureReason": someFailureReason,
+		}
+
+		assert.Equal(t, expectedReportData, testClient.configMapData)
+	})
+
+	t.Run("on checkup multiple failures", func(t *testing.T) {
+		checkupStatus := status.Status{
+			FailureReason: []string{someFailureReason, someOtherFailureReason},
+		}
+		assert.NoError(t, testReporter.Report(checkupStatus))
+
+		expectedReportData := map[string]string{
+			"status.succeeded":     strconv.FormatBool(false),
+			"status.failureReason": someFailureReason + ", " + someOtherFailureReason,
+		}
+
+		assert.Equal(t, expectedReportData, testClient.configMapData)
+	})
 }
 
-func (c configMapClientStub) UpdateConfigMap(_, _ string, _ map[string]string) error {
+type configMapClientStub struct {
+	failUpdateConfigMap error
+	configMapData       map[string]string
+}
+
+func (c *configMapClientStub) UpdateConfigMap(_, _ string, data map[string]string) error {
+	c.configMapData = data
 	return c.failUpdateConfigMap
 }


checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -46,7 +55,61 @@ func TestReportShouldRunSuccessfullyWhen(t *testing.T) {
...
+
+		expectedReportData := map[string]string{
+			reporter.SucceededKey:     "true",
+			reporter.FailureReasonKey: "",
+		}
+
+		assert.Equal(t,
+			expectedReportData,
+			testClient.configMapData,
+		)
```
orelmisan, line 67: AFAIU, `Reporter.Report()` is only called [once](https://github.com/kiagnose/kiagnose/blob/3dac964846dc62170cf1cf725c79c162e5e35fad/checkups/kubevirt-vm-latency/vmlatency/internal/launcher/launcher.go#L59).
I think this test case cannot happen with the production code.
@ormergi please correct me if I'm wrong.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -46,7 +55,61 @@ func TestReportShouldRunSuccessfullyWhen(t *testing.T) {
...
 	})
+
+	testCases := []successTestCase{
+		{
+			description:   "on checkup successful completion",
+			succeeded:     true,
+			failureReason: nil,
+		},
+		{
+			description:   "on checkup failed completion",
```
orelmisan, line 77:
```suggestion
			description:   "on checkup failure",
```

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -46,7 +55,61 @@ func TestReportShouldRunSuccessfullyWhen(t *testing.T) {
...
+			succeeded:     false,
+			failureReason: []string{"some reason"},
+		},
+	}
+	for _, testCase := range testCases {
+		t.Run(testCase.description, func(t *testing.T) {
+			setup()
+			testReporter := reporter.New(testClient, testNamespace, testConfigMapName)
```
orelmisan, line 85: Please consider moving `testReporter` initialization to `setup()`

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -66,9 +129,16 @@ func TestReportShouldFailWhen(t *testing.T) {
...
 type configMapClientStub struct {
+	configMapData       map[string]string
 	failUpdateConfigMap error
 }
 
-func (c configMapClientStub) UpdateConfigMap(_, _ string, _ map[string]string) error {
+func (c *configMapClientStub) UpdateConfigMap(_, _ string, data map[string]string) error {
+	if len(c.configMapData) > 0 {
+		c.configMapData = map[string]string{}
+	}
```
orelmisan, line 139: Please consider adding a factory function for `configMapClientStub`:
```go
func newConfigMapClientStub(failUpdateConfigMap error) *configMapClientStub {
    return &configMapClientStub{
      configMapData: map[string]string{},
      failUpdateConfigMap: failUpdateConfigMap,
  }
}
```

orelmisan (commented):
Thank you for the PR @RamLavi.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: hmm.. this is an already existing test, I just expanded it, so I would be surprised if it's a non relevant case.
@ormergi ?

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: since the returned value of `reporter.New()` is not public, then I can't.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: Done

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
ormergi: Looks like Report() is called one time only in the test, I think its fine.

@RamLavi is there a difference between this test and then line [72](https://github.com/kiagnose/kiagnose/pull/156/commits/3e65a22b405f132678c1ed9aeb541002ea5a6215#diff-2d21b1780f545bd92df143499f288130aaf54c3bde3f23c77cb40c1530d3cee2R72-R74)?

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -46,7 +55,61 @@ func TestReportShouldRunSuccessfullyWhen(t *testing.T) {
...
+			assert.NoError(t, testReporter.Report(checkupStatus))
+
+			expectedReportData := map[string]string{
+				reporter.ResultMinLatencyKey:          fmt.Sprint(checkupStatus.MinLatency.Nanoseconds()),
+				reporter.ResultMaxLatencyKey:          fmt.Sprint(checkupStatus.MaxLatency.Nanoseconds()),
+				reporter.ResultAvgLatencyKey:          fmt.Sprint(checkupStatus.AvgLatency.Nanoseconds()),
+				reporter.ResultMeasurementDurationKey: fmt.Sprint(checkupStatus.MeasurementDuration.Seconds()),
+				reporter.SucceededKey:                 fmt.Sprintf("%t", testCase.succeeded),
+				reporter.FailureReasonKey:             strings.Join(testCase.failureReason, ","),
+			}
```
ormergi, line 105: Please consider exposing `reporter.formatStatus()` instead each of the status keys,
they are considered as an implementation detail that may change over time and break the tests.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -35,17 +35,31 @@ const (
 )
 
 func TestReportShouldRunSuccessfullyWhen(t *testing.T) {
+	var testClient *configMapClientStub
+
+	setup := func() {
+		testClient = &configMapClientStub{}
+	}
```
ormergi, line 51: I suggest to not set gloabl var for the tests client + setup() because the next dev the write a test would need to remember it there.
I would initialize the client at each test according to what is needed instead.
WDYT?

ormergi (commented):
Thanks for the PR, please see my inline comments.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: yes, in `status is initialized` Report is sent with empty status `status.Status{}`
and in the new one it is sent with non empty `status.Status`

orelmisan: > Looks like Report() is called one time only in the test, I think its fine.

Yes, but on the production code it will not have these expected values.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: I don't think it's a good idea
if there will be a bug in formatStatus we will not see it in the unit test anymore.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
orelmisan: It took me by surprise TBH.
IMO it should be exported, I'll open a PR to export it.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: I think using the same testClient variable is a better approach than introducing it each test, but it's not a must in this case.. 
Will remove this commit and fix accordingly

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
orelmisan: PR #159.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
ormergi: If reporter.Report() is changed, these tests need to be changed accordingly, in case a bug is introduced it will also manifest in the test and it also won't be visible, isn't it?

Note that this test suite `TestReportShouldRunSuccessfullyWhen` is not about the correctness of the conversion from `Status{}` to `map[string]string`, it is merely a sanity check.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
ormergi: #159 was opened in order to export `Reporter{}` to enable using it as a global variable in tests where it can be avoided, I think this is wrong.

orelmisan: @ormergi can you please elaborate why do you think it is wrong?

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
orelmisan: I share @RamLavi's opinion the test should be against an independent set of expected results.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
ormergi: > @ormergi can you please elaborate why do you think it is wrong?

I think setting tests objects as globals is wrong, it requires more cognitive effort for the reader to figure our where these objects came from and how they initialize. 
In this test suite it may seem neglectable, but as there are more tests its will be noticeable, FWIW I experienced it myself in other projects.
I also think that exporting Reporter{} just to use it as global in tests where it can be avoided (by setting it in each test) is wrong.

orelmisan: > I also think that exporting Reporter{} just to use it as global in tests where it can be avoided (by setting it in each test) is wrong.

Can you please write it in PR #159?

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
orelmisan: @ormergi @RamLavi can we remove this test case?

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: removed the setup after offline talk with Or

ormergi (commented):
Thanks for the changes.

Regarding testing the conversion from `Status` to `map[string]string` a while back we had a discussion about whether it it's right to have such test, If I remember correctly, we concluded that this test is not effective because it will require a change every time the code it tests changes. 
Also because there is no actual logic to test most of it is variable assignments.

Even though such test would be considered not effective, it gives an indication that the checkup results ConfigMap data is according to the checkup Status as expected, without the need to check it manually with e2e test.

I think that it's okay to have it on a separate test suite until we start using a CustomResource for the checkup object.

Note: using a CustomResource instead of the checkup results ConfigMap which would include a Status object we could assert in tests and require no conversions.

RamLavi:
> Thanks for the changes.
> 
> Regarding testing the conversion from `Status` to `map[string]string` a while back we had a discussion about whether it it's right to have such test, If I remember correctly, we concluded that this test is not effective because it will require a change every time the code it tests changes. Also because there is no actual logic to test most of it is variable assignments.

maybe it was right then, but I think that right now we are not checking the values of the results, and we should. 

> Even though such test would be considered not effective, it gives an indication that the checkup results ConfigMap data is according to the checkup Status as expected, without the need to check it manually with e2e test.
> 
> I think that it's okay to have it on a separate test suite until we start using a CustomResource for the checkup object.

There are 2 ways we can test the values:
1. (Your suggestion) add a unit test in `launcher_test.go` where we have access to the configmap and check there. 
pros: we don't need to expose the const variables.
cons: no logic was added in launcher regarding the configmap, so there is no reason to check it there.
2. (my suggestion) add unit tests in `reporter_test.go`
pros: we're checking as closest to the actual code
cons: we have to expose the variables consts/the entire function (I prefer the former).

in this case I think exposing some const variables is the better approach.

> Note: using a CustomResource instead of the checkup results ConfigMap which would include a Status object we could assert in tests and require no conversions.

no argument here - and when we do so - the tests should be in `reporter_test.go` IMO.

ormergi:
@RamLavi 
> maybe it was right then, but I think that right now we are not checking the values of the results, and we should.
> 

I think it still stands, what's on the stack is having not ideal test vs. coverage and reducing the time it would take to debug it.

> in this case I think exposing some const variables is the better approach.

Maybe I didn't explain myself correctly, I was saying I am fine with having a tests that check the conversion at the reporter test, just put them in their own func to express that it checks the conversion.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: removed it , so yes

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: removed the golbal

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -55,6 +57,12 @@ type configMapClientStub struct {
 	failUpdateConfigMap error
 }
 
+func newConfigMapClientStub(failUpdateConfigMap error) *configMapClientStub {
```
ormergi, line 115: Is it still necessary?

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -53,16 +56,70 @@ func TestReportShouldFailWhen(t *testing.T) {
...
+	}
+
+	testCases := []successTestCase{
+		{
+			description:   "on checkup successful completion",
+			succeeded:     true,
+			failureReason: nil,
+		},
+		{
+			description:   "on checkup failure",
```
ormergi, line 73: Currently, when the checkup fails, the status Results{} is not populated, according to this test one might think that it is.
I suggest keeping each scenario in its own t.Run(), so the failure scenario will not assert status Results{}, WDYT?

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: it wasn't necessary from the beginning :) 
but I think using a factory function is a good call here, as it helps readability...

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: but the results are updated at the start of t.Run(), so they are indeed updated in this case.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
ormergi: IMO there is not much of improvement to readability in this case :sweat_smile: 
See how its going to be used:
```
c := newConfigMapClientStub(expectedErr)
vs.
c := &configMapClientStub{}
c := &configMapClientStub{failUpdateConfigMap: expectedErr}
```
Having said all that, feel free to leave it as is.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
ormergi: Yes, but there is no such scenario, if the checkup fails at` Run()` it will eventually report only `succeeded` and `failureReason`.
this tests show otherwise, that results are reported whether the checkup failed or not, which IMO may confuse the reader thinking this is how the production code works.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: I will, thanks :)

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: ok I'll remove it

ormergi (approved):
Thanks for the changes!

RamLavi:
> IMO we should add a test case of a failed setup/execution/whatever scenario that updates `succeeded` and `failureReason` to `false` and `some reason` respectively.

@orelmisan please refer @ormergi's comment on this:
```
Yes, but there is no such scenario, if the checkup fails at Run() it will eventually report only succeeded and failureReason.
this tests show otherwise, that results are reported whether the checkup failed or not, which IMO may confuse the reader thinking this is how the production code works.
```
Since upon failure there wont be conversion data on the configmap, I tend to agree with Or, we should not document otherwise.

orelmisan:
> Since upon failure there wont be conversion data on the configmap, I tend to agree with Or, we should not document otherwise.

`succeeded` and `failureReason` should be filled in case of a failure, so why can't we test it?

RamLavi:
> > Since upon failure there wont be conversion data on the configmap, I tend to agree with Or, we should not document otherwise.
> 
> `succeeded` and `failureReason` should be filled in case of a failure, so why can't we test it?

[Done](https://github.com/kiagnose/kiagnose/compare/2f650ceb260d497cb2cbb792f05079b2ff7fc06f..544b760105e02b5069432c3d0fd4c733e2549c2e)

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -53,16 +55,68 @@ func TestReportShouldFailWhen(t *testing.T) {
...
+			},
+		}
+		assert.NoError(t, testReporter.Report(checkupStatus))
+
+		expectedReportData := map[string]string{
+			"status.result.minLatencyNanoSec":      fmt.Sprint(checkupStatus.MinLatency.Nanoseconds()),
+			"status.result.maxLatencyNanoSec":      fmt.Sprint(checkupStatus.MaxLatency.Nanoseconds()),
+			"status.result.avgLatencyNanoSec":      fmt.Sprint(checkupStatus.AvgLatency.Nanoseconds()),
+			"status.result.measurementDurationSec": fmt.Sprint(checkupStatus.MeasurementDuration.Seconds()),
+			"status.succeeded":                     fmt.Sprintf("%t", true),
```
orelmisan, line 78: nit:
```suggestion
			"status.succeeded":                     strconv.FormatBool(true),
```

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -53,16 +55,68 @@ func TestReportShouldFailWhen(t *testing.T) {
...
+		assert.Equal(t,
+			expectedReportData,
+			testClient.configMapData,
+		)
+	})
+	t.Run("on checkup failure", func(t *testing.T) {
+		testClient := newConfigMapClientStub(nil)
+		testReporter := reporter.New(testClient, testNamespace, testConfigMapName)
+		checkupStatus := status.Status{
+			FailureReason: []string{"some reason"},
```
orelmisan, line 91: Please consider testing more than one failure reason, so we will know that the conversion is performed correctly, for example:
```go
FailureReason: []string{"some reason", "some other reason"},
```

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -53,16 +55,68 @@ func TestReportShouldFailWhen(t *testing.T) {
...
+		}
+		assert.NoError(t, testReporter.Report(checkupStatus))
+
+		expectedReportData := map[string]string{
+			"status.succeeded":     fmt.Sprintf("%t", false),
```
orelmisan, line 96: nit:
```suggestion
			"status.succeeded":     strconv.FormatBool(false),
```

orelmisan (commented):
Thank you for the changes @RamLavi.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
orelmisan: I think it should come as an additional test. WDYT?

RamLavi: why not

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: Done

RamLavi:
@orelmisan fixes [here](https://github.com/kiagnose/kiagnose/compare/544b760105e02b5069432c3d0fd4c733e2549c2e..540dab940319771c995daa656abc696b544a5fa8)

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -45,16 +49,93 @@ func TestReportShouldRunSuccessfullyWhen(t *testing.T) {
...
+		}
+		assert.NoError(t, testReporter.Report(checkupStatus))
+
+		expectedReportData := map[string]string{
+			"status.result.minLatencyNanoSec":      fmt.Sprint(checkupStatus.MinLatency.Nanoseconds()),
+			"status.result.maxLatencyNanoSec":      fmt.Sprint(checkupStatus.MaxLatency.Nanoseconds()),
+			"status.result.avgLatencyNanoSec":      fmt.Sprint(checkupStatus.AvgLatency.Nanoseconds()),
+			"status.result.measurementDurationSec": fmt.Sprint(checkupStatus.MeasurementDuration.Seconds()),
+			"status.succeeded":                     strconv.FormatBool(true),
+			"status.failureReason":                 "",
```
orelmisan, line 80: Please move the constants from `reporter.formatStatus()` to the top of the file `reporter.go`, export them and use them here.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -45,16 +49,93 @@ func TestReportShouldRunSuccessfullyWhen(t *testing.T) {
...
+		testClient := newConfigMapClientStub(nil)
+		testReporter := reporter.New(testClient, testNamespace, testConfigMapName)
+		checkupStatus := status.Status{
+			FailureReason: []string{"some reason"},
+		}
+		assert.NoError(t, testReporter.Report(checkupStatus))
+
+		expectedReportData := map[string]string{
+			"status.succeeded":     strconv.FormatBool(false),
+			"status.failureReason": "some reason",
```
orelmisan, line 98: Same

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -45,16 +49,93 @@ func TestReportShouldRunSuccessfullyWhen(t *testing.T) {
...
+		testClient := newConfigMapClientStub(nil)
+		testReporter := reporter.New(testClient, testNamespace, testConfigMapName)
+		checkupStatus := status.Status{
+			FailureReason: []string{"some reason", "some other reason"},
+		}
+		assert.NoError(t, testReporter.Report(checkupStatus))
+
+		expectedReportData := map[string]string{
+			"status.succeeded":     strconv.FormatBool(false),
+			"status.failureReason": "some reason, some other reason",
```
orelmisan, line 116: Same.

orelmisan (changes requested):

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
orelmisan: This change will make this file consistent with the framework's [reporter tests](https://github.com/kiagnose/kiagnose/blob/main/kiagnose/internal/reporter/reporter_test.go).

RamLavi: I don't see a value in exporting these consts only for test in this particular case. 
this test should document what we expect to see. the way we format it, it not the issue tested.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: same answer

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: same answer

orelmisan (approved):

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
EdDev: I tend to agree it adds nothing useful.
If you take into account it is a test, can enter a single line and avoids the need to pass an explicit `nil` as input, not having this is a plus IMO.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -53,16 +56,86 @@ func TestReportShouldFailWhen(t *testing.T) {
...
+			"status.result.avgLatencyNanoSec":      fmt.Sprint(checkupStatus.AvgLatency.Nanoseconds()),
+			"status.result.measurementDurationSec": fmt.Sprint(checkupStatus.MeasurementDuration.Seconds()),
+			"status.succeeded":                     strconv.FormatBool(true),
+			"status.failureReason":                 "",
+		}
+
+		assert.Equal(t,
+			expectedReportData,
+			testClient.configMapData,
+		)
```
EdDev, line 86: nit: Can this be an one-liner?

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -53,16 +56,86 @@ func TestReportShouldFailWhen(t *testing.T) {
...
+	})
+	t.Run("on checkup failure", func(t *testing.T) {
```
EdDev, line 88: nit: Can you separate the new test (from the previous one) with a newline?

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -53,16 +56,86 @@ func TestReportShouldFailWhen(t *testing.T) {
...
+		testClient := newConfigMapClientStub(nil)
+		testReporter := reporter.New(testClient, testNamespace, testConfigMapName)
+		checkupStatus := status.Status{
+			FailureReason: []string{"some reason"},
```
EdDev, line 92: It would be better if the correlated data of the reason (`"some reason"`) is expressed as a const and used as input and output expectation.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -53,16 +56,86 @@ func TestReportShouldFailWhen(t *testing.T) {
...
+		assert.Equal(t,
+			expectedReportData,
+			testClient.configMapData,
+		)
+	})
+	t.Run("on checkup multiple failures", func(t *testing.T) {
+		testClient := newConfigMapClientStub(nil)
+		testReporter := reporter.New(testClient, testNamespace, testConfigMapName)
+		checkupStatus := status.Status{
+			FailureReason: []string{"some reason", "some other reason"},
```
EdDev, line 110: same, correlate the input and the output by placing these as constants.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -53,16 +56,86 @@ func TestReportShouldFailWhen(t *testing.T) {
...
 func newConfigMapClientStub(failUpdateConfigMap error) *configMapClientStub {
 	return &configMapClientStub{
 		failUpdateConfigMap: failUpdateConfigMap,
+		configMapData:       map[string]string{},
 	}
 }
 
-func (c configMapClientStub) UpdateConfigMap(_, _ string, _ map[string]string) error {
+func (c *configMapClientStub) UpdateConfigMap(_, _ string, data map[string]string) error {
+	c.configMapData = data
```
EdDev, line 139: Isn't this operation makes the factory initialization of the map redundant (line 134)?

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -53,16 +56,86 @@ func TestReportShouldFailWhen(t *testing.T) {
...
+		}
+
+		assert.Equal(t,
+			expectedReportData,
+			testClient.configMapData,
+		)
+	})
+	t.Run("on checkup multiple failures", func(t *testing.T) {
+		testClient := newConfigMapClientStub(nil)
+		testReporter := reporter.New(testClient, testNamespace, testConfigMapName)
```
EdDev, line 108: This seems to repeat as a common setup. Consider moving it directly under `TestReportShouldSuccessfullyConvertResultValues` before the sub-tests are called.

EdDev (commented):
I had only small comments to add.

>  vm-latency, reporter: Add conversion test 

I think this commit needs reasoning. From the subject line, it is not clear what was missing.
A sentence or two would be nice, explaining why this change was needed.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -36,7 +36,8 @@ const (
 
 func TestReportShouldRunSuccessfullyWhen(t *testing.T) {
 	t.Run("status is initialized", func(t *testing.T) {
-		testReporter := reporter.New(&configMapClientStub{}, testNamespace, testConfigMapName)
+		testClient := newConfigMapClientStub(nil)
```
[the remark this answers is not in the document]
RamLavi: how about 
```suggestion
		var updateConfigMapErr error
		testClient := newConfigMapClientStub(epdateConfigMapErr)
```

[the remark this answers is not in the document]
ormergi: It could make one think we actually inject the error.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: yep :)

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: Done

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: Added
```
	const (
		someFailureReason      = "some reason"
		someOtherFailureReason = "some other reason"
	)
```

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: didn't know it's possible :)
Done

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: Done

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: yep. nice catch :)

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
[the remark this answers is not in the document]
EdDev: My solution is [this](https://github.com/kiagnose/kiagnose/pull/156#discussion_r970344846).
But if you want to keep it, then `var noError error` is what I meant.

RamLavi:
@EdDev fixed [here](https://github.com/kiagnose/kiagnose/compare/540dab940319771c995daa656abc696b544a5fa8..940c721287312d2af6b5fe12af65e9ced5884fcb)

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
EdDev: I do not see it changed.
I would suggest not to resolve conversations before getting the commenter ack, it will make it easier to track it and not assume resolution prematurely.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
EdDev: Which in turn makes the whole factory redundant IMO.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
EdDev: It is not always possible. If the tests mutate the objects in a way it interferes, then it is not possible.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -54,7 +57,73 @@ func TestReportShouldFailWhen(t *testing.T) {
...
 }
 
+func TestReportShouldSuccessfullyConvertResultValues(t *testing.T) {
+	const (
+		someFailureReason      = "some reason"
+		someOtherFailureReason = "some other reason"
+	)
+	var updateConfigMapErr error
+	testClient := newConfigMapClientStub(updateConfigMapErr)
+	testReporter := reporter.New(testClient, testNamespace, testConfigMapName)
```
orelmisan, line 67: Please note that these are shared between tests (the tests that comes after the first test do not receive a fresh object).
Maybe consider:
```go
var testClient configMapClientStub
var testReporter *reporter.Reporter

setup := func() {
	testClient = newConfigMapClientStub(updateConfigMapErr)
	testReporter = reporter.New(testClient, testNamespace, testConfigMapName)
}
```

and than call `setup()` in the beginning of each test.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: ACK :)

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: I won't insist but I think it improves readability.. tell me if you feel strongly against it.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -45,16 +50,91 @@ func TestReportShouldRunSuccessfullyWhen(t *testing.T) {
...
+
+	t.Run("on checkup multiple failures", func(t *testing.T) {
+		checkupStatus := status.Status{
+			FailureReason: []string{someFailureReason, someOtherFailureReason},
+		}
+		assert.NoError(t, testReporter.Report(checkupStatus))
+
+		expectedReportData := map[string]string{
+			"status.succeeded":     strconv.FormatBool(false),
+			"status.failureReason": someFailureReason + ", " + someOtherFailureReason,
```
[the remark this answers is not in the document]
RamLavi: I don't think it's important how we set the string, as long as it is as expected.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: usually I would concur, but in this case (as was already discussed in previous comments) I don't think it's necessary to use a setup function.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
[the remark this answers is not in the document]
orelmisan: My suggestion supports concatenation of arbitrary number of failure reasons.
Its not critical at the moment, thus marked as a nit.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
ormergi: I am against setting test objects (client, reporter, etc..) as globals they are prone to bugs, as the test suite grows and more tests are added it's harder to follow and it makes the reader search where they are initialized and what kind of magic setup() does.
Also, once a test requires a slightly different setup, `setup()` will require a change or removal (when it won't fit all tests).

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
EdDev: I do won't insist, but I have trouble agreeing with your argument.
I do not think you gain anything by using a function.

Lets compare:
`newConfigMapClientStub(nil)`
vs
`&configMapClientStub{}`

`newConfigMapClientStub(UpdateConfigMapErr)`
vs
`&configMapClientStub{failUpdateConfigMap: UpdateConfigMapErr}`

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
EdDev: It is not global, it is limited to the specific tests that have been added here and these do not need different objects.
They are also very close to each other.

I have no string feeling to any other the options.

RamLavi:
@EdDev fixed [here](https://github.com/kiagnose/kiagnose/compare/05d91add5df9f8fe0e138f6ca7e6974d4dac003e..9eaff975ddb6db29faeace0a0ff7e3a1af4382b5)

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
```diff
@@ -36,7 +36,9 @@ const (
 
 func TestReportShouldRunSuccessfullyWhen(t *testing.T) {
 	t.Run("status is initialized", func(t *testing.T) {
-		testReporter := reporter.New(&configMapClientStub{}, testNamespace, testConfigMapName)
+		var noError error
+		testClient := newConfigMapClientStub(noError)
```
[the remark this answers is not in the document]
RamLavi: I have considered. my conclusion hasn't changed.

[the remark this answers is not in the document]
EdDev: There have been 2 other conversations this has been the subject of.
Please continue the discussion there, there is no need to open a 3rd one now.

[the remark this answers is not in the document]
ormergi: I commented here because this is the relevant diff, no intention to make more noise.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
ormergi: Since the last thread is resolved, I am commenting here for visibility:

I think [this](https://github.com/kiagnose/kiagnose/pull/156/files/2bc06754d355f3f5f19f516da461b5b1efe40f55#diff-2d21b1780f545bd92df143499f288130aaf54c3bde3f23c77cb40c1530d3cee2R39-R40) is an example of why this function is redundant, it enforces you to use it where it's not really needed and not in the way it should be used (passing not initialized error just to make it work, not much different than specifying nil).
 
*If this is not the right thread, let me know which one is it.

orelmisan:
[Second](https://github.com/kiagnose/kiagnose/pull/156/commits/9eaff975ddb6db29faeace0a0ff7e3a1af4382b5) commit message:
> Currently the reporter does not check the values are as expected.
Added test to check that the conversion is correct.

I suggest to add:
Currently the reporter **tests** does....

RamLavi:
@orelmisan @ormergi @EdDev removed the factory function. see [here](https://github.com/kiagnose/kiagnose/compare/9eaff975ddb6db29faeace0a0ff7e3a1af4382b5..f01a05d3095036661c0540a6b22db9ace0e91b0e)
In the future let's keep these kind of discussions more to the point, and in one place.

ormergi (approved):
Thanks for the changes!

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
RamLavi: I removed the factory, I'm also convinced it is not needed. resolving this issue

ormergi (approved):

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
orelmisan: This is not a blocker and could happen in a follow-up PR.

checkups/kubevirt-vm-latency/vmlatency/internal/reporter/reporter_test.go
orelmisan: `reporter.reporter` is stateless to its not a major issue at the moment.
We could change it in a follow up PR.

orelmisan (approved):
Thank you for the changes @RamLavi.

EdDev (approved):
Raising

PR #159, opened 2022-09-06
vm latency, reporter: Export Reporter struct

orelmisan:
Currently, the `reporter.reporter` struct is not exported, but is returned from `reporter.New()`.

Export `reporter.Reporter` struct, in order for it to be handled more conveniently (for example in tests).


EdDev (commented):
It is hard to understand from this change the reason.
If it worked so far, why would we need to change it? Perhaps introduce this as a commit in the PR that requires it? It will be easier to follow its reasoning.

orelmisan:
Thank you @EdDev, sounds right.
I'll close this PR.
@RamLavi its probably a better fit to PR #156.

PR #165, opened 2022-09-13
kubevirt-vm-latnecy: Randomize VMs affinity label

ormergi:
Following #155, https://github.com/kiagnose/kiagnose/pull/155#discussion_r969254470

Running more than one instance of kubevirt latency checkup is not possible due to the fact that VMs of the first instance will cause the VMs of the other instances to be stuck on the scheduling phase due to the first instance VMs affinity rules.

Change the VMs label that is used by their Pod anti-affinity to have a random value, so that VMs that were created by one instance of the checkup won't block other instances VMs.

Rebased on #155, only the last commit is relevant.

### DIFF
--- a/checkups/kubevirt-vm-latency/go.sum
+++ b/checkups/kubevirt-vm-latency/go.sum
@@ -143,7 +143,6 @@ github.com/cockroachdb/datadriven v0.0.0-20200714090401-bf6692d28da5/go.mod h1:h
 github.com/cockroachdb/errors v1.2.4/go.mod h1:rQD95gz6FARkaKkQXUksEje/d9a6wBJoCr5oaCLELYA=
 github.com/cockroachdb/logtags v0.0.0-20190617123548-eb05cc24525f/go.mod h1:i/u985jwjWRlyHXQbwatDASoW0RMlZ/3i9yJHE2xLkI=
 github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd/go.mod h1:sE/e/2PUdi/liOCUjSTXgM1o87ZssimdTWN964YiIeI=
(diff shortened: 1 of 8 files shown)


orelmisan:
@ormergi Could you please rebase?

RamLavi (approved):
/lgtm

ormergi:
[Rebased](https://github.com/kiagnose/kiagnose/compare/36fee9a781d77c521772025567633779dd25043d..39c3c627ab8d80e45e02dfd005a054ce808ca5cf)

RamLavi (approved):
/lgtm

orelmisan (commented):
Following our offline discussion, IMO we should have a label that has the value of the checkup's name, which is unique in a namespace and set by Kiagnose.
This will enable a user to filter VMI objects by the checkup name:
```bash
kubectl get vmis -l <label-name>=<checkup-name>
```

The same label could be used as well for pod anti affinity.

The con of using a UID IMO is that it cannot be related to the checkup name, so a user could not tell which VMI belongs to which checkup.

ormergi:
> The con of using a UID IMO is that it cannot be related to the checkup name, so a user could not tell which VMI belongs to which checkup.

Note that the con that you mention is a different problem than what this PR is meant to resolve.
This PR addresses the problem when running more than one instance of the latency checkup VMs wont block each other due to their label.
What you describe is being able to correlate the user-supplier ConfigMap to the VMs it creates.

Following offline discussion, it was concluded that for now its good enough to piggy-back this label to solve both issues,
where we will use the user-supplier ConfigMap UUID as the checkup identifier.
Once we have those bits I will adapt this PR to use it instead of generating the UID in place.

EdDev:
@ormergi , is this change still relevant?

ormergi:
> @ormergi , is this change still relevant?

nope. 

/close

PR #167, opened 2022-09-14
vmlatency, launcher, tests: Fix a call to assert.Equal()

orelmisan:
assert.Equal() expects the expected results to be the first argument, and the actual argument second.

Fix a call which was in the reverse order.


ormergi (approved):

RamLavi (approved):
/lgtm

EdDev (approved):

PR #169, opened 2022-09-19
vmlatency, config: Refactor

orelmisan:
Currently, it is not trivial to consume a new field in the API between the framework and the checkup due to the complexity of `config.New()`:

1. Currently, there is a distinction between the states where a mandatory environment variable is:
  - Missing
  - Invalid

In order to reduce the number of possible emitted errors by a factor of two, and to simplify the parsing code - treat all missing mandatory environment variables as invalid.

2. Simplify the parsing and validation of the checkup's configuration:
  - Read all string environment variables in one go.
  - Parse all string environment variables in one go.
  - Replace multiple errors related to source and target nodes with one error.

  Keep the parsing and validation of integer fields as-is.

### DIFF
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
@@ -54,13 +54,9 @@ type Config struct {
 
 var (
 	ErrInvalidEnv                       = errors.New("environment is invalid")
-	ErrResultsConfigMapNameMissing      = fmt.Errorf("%q environment variable is missing", ResultsConfigMapNameEnvVarName)
 	ErrInvalidResultsConfigMapName      = fmt.Errorf("%q environment variable is invalid", ResultsConfigMapNameEnvVarName)
-	ErrResultsConfigMapNamespaceMissing = fmt.Errorf("%q environment variable is missing", ResultsConfigMapNamespaceEnvVarName)
 	ErrInvalidResultsConfigMapNamespace = fmt.Errorf("%q environment variable is invalid", ResultsConfigMapNamespaceEnvVarName)
-	ErrNetworkNameMissing               = fmt.Errorf("%q environment variable is missing", NetworkNameEnvVarName)
 	ErrInvalidNetworkName               = fmt.Errorf("%q environment variable is invalid", NetworkNameEnvVarName)
-	ErrNetworkNamespaceMissing          = fmt.Errorf("%q environment variable is missing", NetworkNamespaceEnvVarName)
 	ErrInvalidNetworkNamespace          = fmt.Errorf("%q environment variable is invalid", NetworkNamespaceEnvVarName)
 	ErrSourceNodeNameMissing            = fmt.Errorf("%q environment variable is missing", SourceNodeNameEnvVarName)
 	ErrInvalidSourceNodeName            = fmt.Errorf("%q environment variable is invalid", SourceNodeNameEnvVarName)
@@ -78,34 +74,22 @@ func New(env map[string]string) (Config, error) {
 		return Config{}, ErrInvalidEnv
 	}
 
-	resultsConfigMapName, exists := env[ResultsConfigMapNameEnvVarName]
-	if !exists {
-		return Config{}, ErrResultsConfigMapNameMissing
-	}
+	resultsConfigMapName := env[ResultsConfigMapNameEnvVarName]
 	if resultsConfigMapName == "" {
 		return Config{}, ErrInvalidResultsConfigMapName
 	}
 
-	resultsConfigMapNamespace, exists := env[ResultsConfigMapNamespaceEnvVarName]
-	if !exists {
-		return Config{}, ErrResultsConfigMapNamespaceMissing
-	}
+	resultsConfigMapNamespace := env[ResultsConfigMapNamespaceEnvVarName]
 	if resultsConfigMapNamespace == "" {
 		return Config{}, ErrInvalidResultsConfigMapNamespace
 	}
 
-	networkName, exists := env[NetworkNameEnvVarName]
-	if !exists {
-		return Config{}, ErrNetworkNameMissing
-	}
+	networkName := env[NetworkNameEnvVarName]
 	if networkName == "" {
 		return Config{}, ErrInvalidNetworkName
 	}
 
-	networkNamespace, exists := env[NetworkNamespaceEnvVarName]
-	if !exists {
-		return Config{}, ErrNetworkNamespaceMissing
-	}
+	networkNamespace := env[NetworkNamespaceEnvVarName]
 	if networkNamespace == "" {
 		return Config{}, ErrInvalidNetworkNamespace
 	}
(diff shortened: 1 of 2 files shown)


orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/1a41f39340d93aac31113fa41840fc740c2c2075..afcdb0aad4256e8115dd73563930bab740ae8df4): Replaced `fmt.Errorf()` with `errors.New()`.

ormergi (commented):
HI Orel, thanks for the PR.

I think that having the "missing env var" errors has values as it provides a more accurate failure reason.
Regarding the commit message, the possible error count is not actually reduced, it's just that the error message would be the same for different reasons (missing and invalid parameters).

The rest of the changes look good to me.

RamLavi (approved):
I think this can make the code a bit lighter to read..
/lgtm

orelmisan:
> I think that having the "missing env var" errors has values as it provides a more accurate failure reason.

I share your opinion that a more accurate error message is better, but in the balance between code readability and a bit more accuracy in the error message - IMO the code readability wins.

> Regarding the commit message, the possible error count is not actually reduced, it's just that the error message would be the same for different reasons (missing and invalid parameters).

I've changed the phrasing to:
> In order to reduce the number of possible emitted errors by a factor of two...

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/afcdb0aad4256e8115dd73563930bab740ae8df4..b71025c8a596e79a5bc84a492b38c58e0f7c7c5d): Edited [first](https://github.com/kiagnose/kiagnose/pull/169/commits/765434623cc05b5e8bd03ec5750bba68abe9678d) commit message in order to answer @ormergi's [comment](https://github.com/kiagnose/kiagnose/pull/169#pullrequestreview-1113500656).

ormergi:
> I share your opinion that a more accurate error message is better, but in the balance between code readability and a bit more accuracy in the error message - IMO the code readability wins.

I dont see how redability is improved sure you have less code now so it looks better, but in the cost of informative error messages, I think it's wrong.
It also won't scale, if tomorrow there will be more parameters to handle the code will ends up in the same state as now.

If you want to improve readability you could extract a batch of related parameters validation to functions, similar to `validateNodeNames()`.

orelmisan:
> > I share your opinion that a more accurate error message is better, but in the balance between code readability and a bit more accuracy in the error message - IMO the code readability wins.
> 
> I dont see how redability is improved sure you have less code now so it looks better, but in the cost of informative error messages, I think it's wrong. It also won't scale, if tomorrow there will be more parameters to handle the code will ends up in the same state as now.
> 

Every string field had at minimum two `if` statements:
1. Check that it exists.
2. Check that it has a legal value.

Now there is only the latter, thus reducing the `if` statements by a factor of two.

RamLavi (approved):
/lgtm

RamLavi (dismissed):
/lgtm

ormergi:
> Every string field had at minimum two `if` statements:
> 
>     1. Check that it exists.
> 
>     2. Check that it has a legal value.
> 
> 
> Now there is only the latter, thus reducing the `if` statements by a factor of two.

As I said earlier the problem stays once there will be more parameters to handle.
If I understand you correctly, you agree that the error messages should stay as they are, what I suggested will enable keeping them and improve the code readability (extract parameters to functions).

orelmisan:
> > Every string field had at minimum two `if` statements:
> > ```
> > 1. Check that it exists.
> > 
> > 2. Check that it has a legal value.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Now there is only the latter, thus reducing the `if` statements by a factor of two.
> 
> As I said earlier the problem stays once there will be more parameters to handle. If I understand you correctly, you agree that the error messages should stay as they are, what I suggested will enable keeping them and improve the code readability (extract parameters to functions).

I prefer to have more accurate error messages. The cost of having them at the moment is to have a lot of `if` statements.
In this specific case, having accurate error messages has less priority than having a shorter and clearer code.

@ormergi how about we'll hold this PR, so you could suggest an alternative change?

orelmisan:
Hi @ormergi, ten days had passed since our last discussion on this PR.
Is there an alternative PR you wish to suggest instead of this one? If not, I suggest we move forward with this PR.
An alternative change could also be suggested after this PR is merged.

ormergi:
> Hi @ormergi, ten days had passed since our last discussion on this PR. Is there an alternative PR you wish to suggest instead of this one? If not, I suggest we move forward with this PR. An alternative change could also be suggested after this PR is merged.

Ten days ago I suggested an alternative to the changes I asked https://github.com/kiagnose/kiagnose/pull/169#issuecomment-1254662243, but you did not respond to it, I am not sure what you expect.

orelmisan:
Summary:
We both agreed that having a more accurate error messages is beneficial.
It was suggested by @ormergi to keep the "missing" environment variables and to split `New()` into multiple sub-functions, each containing the two `if` statements.
I disagree because I find the complexity not worth it (and would probably add another `if` to check if the sub-function had failed).
I claim that the cost of having two possible errors for each environment variable is too high and does not scale.
@EdDev WDYT?

checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
```diff
@@ -101,48 +89,44 @@ func New(env map[string]string) (Config, error) {
...
+
+	if c.NetworkAttachmentDefinitionName == "" {
+		return ErrInvalidNetworkName
+	}
+
+	if c.NetworkAttachmentDefinitionNamespace == "" {
+		return ErrInvalidNetworkNamespace
+	}
+
+	if c.SourceNodeName == "" && c.TargetNodeName != "" {
```
EdDev, line 126: The condition is strange (I mean, using if/else).

How about one condition:
```
if c.SourceNodeName == "" && c.TargetNodeName != "" || c.SourceNodeName != "" && c.TargetNodeName == "" {
    return ErrIllegalSourceAndTargetNodesCombination
}
```

EdDev (approved):
The main change I have seen is dropping the `missing` errors, merging them with the `invalid` ones.
It indeed reduces the reported error accuracy, but I do not think it is dramatic.
One could count of a missing key or a missing value (i.e. empty string) as the same thing, in both there is a missing/invalid data).

I do not know if there is any correct answer here. Seems like a tie score to me.
I think we can take this in and if a more structured validation is presented in the future, we can re-evaluate it.

checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
orelmisan: I agree it looks much better.
Done.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/b71025c8a596e79a5bc84a492b38c58e0f7c7c5d..9f98bd37400088016fb60eb673e2979c550dc6fb): Answered @EdDev [comment](https://github.com/kiagnose/kiagnose/pull/169#discussion_r985402970).

EdDev (approved):
Thanks!

Issue #164, opened 2022-09-12
VM latency checkup: Cannot simultaneously run multiple VM latency checkups in the same namespace


orelmisan:
It is not possible to run more than one instance of the VM latency checkup simultaneously in the same namespace.
The checkup creates the VMIs with hard coded:
1. Names.
2. MAC addressed.
3. IP addresses.

PR #171, opened 2022-09-19
vmlatency, checkup: Name VMIs after the checkup's name

orelmisan:
Currently, the VMIs created by the VM latency checkup has hard-coded names.
This prevents execution of multiple instances of this checkup in parallel (for more details see issue #164).

1. Add a new `CHECKUP_NAME` environment variable in the API between the framework and the checkup.
2. Use it to create the names of the VMIs in VM latency checkup.

Depends on PR #169.

### DIFF
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
@@ -48,6 +48,7 @@ type checker interface {
 type checkup struct {
 	client    vmi.KubevirtVmisClient
 	namespace string
+	name      string
 	params    config.CheckupParameters
 	results   status.Results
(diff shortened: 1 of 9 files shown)


orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/307712768ddd0ccb522575af416e256433a280c0..7c987c478c2fa991583353ad7ab74ee8feb91ec2): Changed the VMI name to `<checkup-name>-<suffix>` in order to make a list of VMI objects to have better readability when sorted.

orelmisan:
After an offline discussion, it was decided to use the UID of the user-supplied ConfigMap object instead of its name.
The user-supplied ConfigMap object could be deleted and recreated with the same name, but with other UID.
This was deemed not good enough.

PR #173 Replaces this PR.

PR #172, opened 2022-09-20
Bump Base image version

orelmisan:
Bump base image version of:
- Kiagnose
- Echo checkup
- VM latency checkup


RamLavi (approved):
/lgtm

EdDev (approved):
Seems like an overkill to create a commit per each file, but it is not worth discussing beyond this small comment.

Thanks!

PR #166, opened 2022-09-14
vmlatency, config, tests: refactor

orelmisan:
Refactor VM latency's config tests, in order to improve code readability and enable simpler addition of tests in the future.

PR #173, opened 2022-09-20
vmlatency: Use checkup UID as label value

orelmisan:
Currently, the source and target VMIs have a label `latency-check-vm` with a hard-coded empty string value.
This prevents execution of multiple instances of this checkup in parallel (for more details see issue #164).

1. Add a new `CHECKUP_UID` environment variable in the API between the framework and the checkup.
2. Change the label name to "latency-check/uid" to better reflect its value.
3. Add the checkup's UID as the value of the "latency-check/uid" label.

To get checkup's UID:
```bash
kubectl get cm -n <target-namespace> <cm-name> -o jsonpath='{.metadata.uid}'
```

To list VMI objects by checkup's UID:
```bash
kubectl get vmis -n <target-namespace> -l latency-check-vm=<checkup-UID>
```

~~Depends on PR #169.~~
Replaces PR #171.

### DIFF
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
@@ -46,6 +46,7 @@ type checker interface {
 
 type checkup struct {
 	client    vmi.KubevirtVmisClient
+	uid       string
 	namespace string
 	params    config.CheckupParameters
 	results   status.Results
@@ -54,9 +55,10 @@ type checkup struct {
 	checker   checker
 }
 
-func New(c vmi.KubevirtVmisClient, namespace string, params config.CheckupParameters, checker checker) *checkup {
+func New(c vmi.KubevirtVmisClient, uid, namespace string, params config.CheckupParameters, checker checker) *checkup {
 	return &checkup{
 		client:    c,
+		uid:       uid,
 		namespace: namespace,
 		params:    params,
 		checker:   checker,
@@ -68,9 +70,9 @@ func (c *checkup) Preflight() error {
 }
 
 const (
-	SourceVmiName       = "latency-check-source"
-	TargetVmiName       = "latency-check-target"
-	LabelLatencyCheckVM = "latency-check-vm"
+	SourceVmiName        = "latency-check-source"
+	TargetVmiName        = "latency-check-target"
+	LabelLatencyCheckUID = "latency-check/uid"
 )
 
 func (c *checkup) Setup(ctx context.Context) error {
@@ -92,8 +94,8 @@ func (c *checkup) Setup(ctx context.Context) error {
 		return fmt.Errorf("%s: %v", errMessagePrefix, err)
 	}
 
-	sourceVmi := newLatencyCheckVmi(SourceVmiName, c.params.SourceNodeName, netAttachDef, sourceVmiMac, sourceVmiCidr)
-	targetVmi := newLatencyCheckVmi(TargetVmiName, c.params.TargetNodeName, netAttachDef, targetVmiMac, targetVmiCidr)
+	sourceVmi := newLatencyCheckVmi(c.uid, SourceVmiName, c.params.SourceNodeName, netAttachDef, sourceVmiMac, sourceVmiCidr)
+	targetVmi := newLatencyCheckVmi(c.uid, TargetVmiName, c.params.TargetNodeName, netAttachDef, targetVmiMac, targetVmiCidr)
 
 	if err = vmi.Start(c.client, c.namespace, sourceVmi); err != nil {
 		return fmt.Errorf("%s: %v", errMessagePrefix, err)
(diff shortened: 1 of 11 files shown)


ormergi:
It seems that this PR now overlaps with #165 in the sense of randomizing the node affinity label.

orelmisan:
> It seems that this PR now overlaps with #165 in the sense of randomizing the node affinity label.

That is right.
This was the result of our offline discussion - instead of using the user-supplied ConfigMap name as an identifier for the checkup, use its UID.
We've also agreed that this UID shall be used as the value for the label on the VMIs.

ormergi:
> This was the result of our offline discussion - instead of using the user-supplied ConfigMap name as an identifier for the checkup, use its UID.

I mean for the last commit, where the UID is being used for the affinity label, this overlaps with this PR.

RamLavi:
the PR desc does not mention the motive. can you specify that?

orelmisan:
> the PR desc does not mention the motive. can you specify that?

From the PR description:
> Currently, the source and target VMIs have a label latency-check-vm with a hard-coded empty string value.
This prevents execution of multiple instances of this checkup in parallel (for more details see issue https://github.com/kiagnose/kiagnose/issues/164).

RamLavi (commented):
I may be missing something, but where are we actually consuming this? all I see is insertion and validation. where do we add it as a the label value?

orelmisan:
> I may be missing something, but where are we actually consuming this? all I see is insertion and validation. where do we add it as a the label value?

https://github.com/kiagnose/kiagnose/pull/173/files#diff-d60aa090110dbe4421acf2db6a61ce9c88ae82264bd0e66ddc999917378b3c22R144

RamLavi (approved):
ahh, missed the last commit. 
/lgtm

ormergi (changes requested):
I think PR should not depend on #166, it can stand on its own.
The last commit overlaps with #165, you could remove it and let #165 do the integration as we concluded offline.

orelmisan:
> I think PR should not depend on #166, it can stand on its own. The last commit overlaps with #165, you could remove it and let #165 do the integration as we concluded offline.

I believe you are referring to PR #169 and not PR #166 which was merged.
Without the refactor to `config.New()`, the linter fails because of high cyclomatic complexity.

The last commit suggests an alternative mechanism to set the value of the VMI label to the UID from the user-supplied CM, instead of generating it on the fly, as was suggested in PR #165.
This solves the [issue](https://github.com/kiagnose/kiagnose/pull/165#pullrequestreview-1113312737) raised in PR #165.

Do you have any comments regarding the code?

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/251685b4590b4cea46976296c8a29c073a15a493..8d1adc108953a0c02e0775a901c12a5628d6b037): Rebase

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -138,7 +141,7 @@ func newLatencyCheckVmi(
 		vmiInterface = vmi.NewInterface(networkName, vmi.WithMacAddress(macAddress), vmi.WithBridgeBinding())
 	}
 
-	vmLabel := vmi.Label{Key: LabelLatencyCheckVM, Value: ""}
+	vmLabel := vmi.Label{Key: LabelLatencyCheckVM, Value: uid}
```
ormergi, line 144: Please change the label name and variable name to `LabelLatencyCheckUID = "latency-check/uid"` so it describes it correctly.

ormergi (changes requested):

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/4b2406dcf1bc7230c1e3707de519a460a0d08e38..88e8593186ecdd6a525473d5d3e1c5bcf47ce18b): Rebase.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
orelmisan: Done.

RamLavi (approved):
/lgtm

ormergi (approved):

checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
```diff
@@ -107,6 +111,10 @@ func New(env map[string]string) (Config, error) {
 }
 
 func (c Config) validate() error {
+	if c.CheckupUID == "" {
+		return ErrInvalidCheckupUID
```
EdDev, line 115: While this is a minor incident at this stage, this should not have been overseen:
The change breaks compatibility with the previous kiagnose core. One could have upgraded on but not the other.

At the minimum this needs to be mentioned in the change and in the changelog.
A better solution would have been to have an alternative (randomizing an UID?).
For the long run, we need to focus on such cases and provide guidelines or tooling to handle them.

EdDev (approved):
I had one comment which I would have considered a blocker, but at the stage of the project, it is unlikely that we have any usage/user who can be affected.

Nevertheless, I suggest to be careful and be explicit when breaking compatibility.
It is also a good opportunity to add a section into the documentation and future designs about these scenarios.

PR #174, opened 2022-09-21
checkup, Collect logs

RamLavi:
Currently checkup job's logs are not collected by kiagnose. 
This PR collects the checkup's logs and print them in kiagnose's logs in order to make them more visible for de facto debugging.

### DIFF
--- a/kiagnose/internal/checkup/checkup.go
+++ b/kiagnose/internal/checkup/checkup.go
@@ -122,9 +122,11 @@ func (c *Checkup) Run() error {
 		return fmt.Errorf("%s: %v", errPrefix, err)
 	}
 
-	if c.job, err = job.WaitForJobToFinish(c.client, c.job, c.jobTimeout); err != nil {
+	var updatedJob *batchv1.Job
+	if updatedJob, err = job.WaitForJobToFinish(c.client, c.job, c.jobTimeout); err != nil {
 		return fmt.Errorf("%s: %v", errPrefix, err)
 	}
+	c.job = updatedJob
 
 	return nil
 }
--- a/kiagnose/internal/checkup/checkup.go
+++ b/kiagnose/internal/checkup/checkup.go
@@ -22,6 +22,7 @@ package checkup
 import (
 	"errors"
 	"fmt"
+	"log"
 	"strings"
 	"time"
 
@@ -133,6 +134,20 @@ func (c *Checkup) Run() error {
 
 func (c *Checkup) Results() (results.Results, error) {
 	return results.ReadFromConfigMap(c.client, c.resultConfigMap.Namespace, c.resultConfigMap.Name)
+}
+
+func (c *Checkup) Logs() error {
+	if c.job == nil {
+		return fmt.Errorf("job is nil")
+	}
+
+	var err error
+	var logs string
+	if logs, err = job.GetLogs(c.client, c.job); err != nil {
+		return err
+	}
+	log.Printf("checkup job %q Logs:\n%s\n", c.job.Name, logs)
+	return nil
 }
 
 func (c *Checkup) SetTeardownTimeout(duration time.Duration) {
--- a/kiagnose/internal/checkup/job/job.go
+++ b/kiagnose/internal/checkup/job/job.go
@@ -113,6 +113,33 @@ func DeleteAndWait(client kubernetes.Interface, job *batchv1.Job, timeout time.D
 	return err
 }
 
+func GetLogs(client kubernetes.Interface, job *batchv1.Job) (string, error) {
+	jobPodName, err := getPodNameByJob(client, job.Name, job.Namespace)
+	if err != nil {
+		return "", err
+	}
+	rawLogs, err := client.CoreV1().Pods(job.Namespace).GetLogs(jobPodName, &corev1.PodLogOptions{}).DoRaw(context.Background())
+	if err != nil {
+		return "", fmt.Errorf("failed to find job's %s/%s pod: %v", job.Namespace, job.Name, err)
+	}
+
+	return string(rawLogs), nil
+}
+
+func getPodNameByJob(client kubernetes.Interface, jobName, jobNamespace string) (string, error) {
+	const JobNameLabel = "job-name"
+	jobLabel := k8slabels.Set{JobNameLabel: jobName}
+	podList, err := client.CoreV1().Pods(jobNamespace).List(context.Background(), metav1.ListOptions{LabelSelector: jobLabel.String()})
+	if err != nil {
+		return "", err
+	}
+
+	if len(podList.Items) != 1 {
+		return "", fmt.Errorf("failed to find job's pod: found %d pods for Job %s/%s", len(podList.Items), jobNamespace, jobName)
+	}
+	return podList.Items[0].Name, nil
+}
+
 func batchJobFinished(event k8swatch.Event) (bool, error) {
 	j, ok := event.Object.(*batchv1.Job)
 	if !ok {
--- a/kiagnose/internal/launcher/launcher.go
+++ b/kiagnose/internal/launcher/launcher.go
@@ -21,6 +21,7 @@ package launcher
 
 import (
 	"errors"
+	"log"
 	"strings"
 	"time"
 
@@ -32,6 +33,7 @@ type workload interface {
 	Setup() error
 	Run() error
 	Results() (results.Results, error)
+	Logs() error
 	Teardown() error
 }
 
@@ -81,6 +83,10 @@ func (l Launcher) Run() (runErr error) {
 	}
 
 	defer func() {
+		if err := l.checkup.Logs(); err != nil {
+			log.Printf("failed to collect logs: %v", err)
+		}
+
 		if err := l.checkup.Teardown(); err != nil {
 			errorPool = append(errorPool, err)
 		}
--- a/kiagnose/internal/launcher/launcher_test.go
+++ b/kiagnose/internal/launcher/launcher_test.go
@@ -256,6 +256,10 @@ func (s checkupStub) Results() (results.Results, error) {
 	return s.results, nil
 }
 
+func (s checkupStub) Logs() error {
+	return nil
+}
+
 func (s checkupStub) Teardown() error {
 	return s.failTeardown
 }
--- a/kiagnose/internal/checkup/checkup_test.go
+++ b/kiagnose/internal/checkup/checkup_test.go
@@ -369,7 +369,7 @@ func TestCheckupRunShouldFailWhen(t *testing.T) {
 
 		assert.NoError(t, testCheckup.Setup())
 		assert.ErrorContains(t, testCheckup.Run(), "initial RV \"\" is not supported")
-		assert.NoError(t, testCheckup.Teardown())
+		assert.ErrorContains(t, testCheckup.Teardown(), "initial RV \"\" is not supported")
 	})
 
 	t.Run("Job wont finish on time", func(t *testing.T) {
@@ -385,6 +385,7 @@ func TestCheckupRunShouldFailWhen(t *testing.T) {
 
 		assert.NoError(t, testCheckup.Setup())
 		assert.ErrorContains(t, testCheckup.Run(), wait.ErrWaitTimeout.Error())
+		testClient.injectWatchWithJobDeleteEvent(testCheckupName)
 		assert.NoError(t, testCheckup.Teardown())
 	})
 
@@ -405,6 +406,7 @@ func TestCheckupRunShouldFailWhen(t *testing.T) {
 
 		assert.NoError(t, testCheckup.Setup())
 		assert.ErrorContains(t, testCheckup.Run(), wait.ErrWaitTimeout.Error())
+		testClient.injectWatchWithJobDeleteEvent(testCheckupName)
 		assert.NoError(t, testCheckup.Teardown())
 	})
 }
--- a/kiagnose/internal/checkup/checkup_test.go
+++ b/kiagnose/internal/checkup/checkup_test.go
@@ -171,7 +171,7 @@ func TestTeardownInTargetNamespaceShouldSucceed(t *testing.T) {
 	testClient.injectJobWatchEvent(newJobWithCondition(checkupJobName, completeTrueJobCondition))
 	assert.NoError(t, testCheckup.Run())
 
-	testClient.injectWatchWithJobDeleteEvent(testTargetNs, checkupJobName)
+	testClient.injectWatchWithJobDeleteEvent(checkupJobName)
 	assert.NoError(t, testCheckup.Teardown())
 
 	assertNamespaceExists(t, testClient.Clientset, testTargetNs)
@@ -279,7 +279,7 @@ func TestCheckupRunShouldCreateAJob(t *testing.T) {
 			actualJob.ResourceVersion = ""
 			assert.Equal(t, actualJob, expectedJob)
 
-			testClient.injectWatchWithJobDeleteEvent(testTargetNs, checkupJobName)
+			testClient.injectWatchWithJobDeleteEvent(checkupJobName)
 			assert.NoError(t, testCheckup.Teardown())
 		})
 	}
@@ -317,7 +317,7 @@ func TestCheckupRunShouldSucceed(t *testing.T) {
 			assert.NoError(t, testCheckup.Setup())
 			assert.NoError(t, testCheckup.Run())
 
-			testClient.injectWatchWithJobDeleteEvent(testTargetNs, checkupJobName)
+			testClient.injectWatchWithJobDeleteEvent(checkupJobName)
 			assert.NoError(t, testCheckup.Teardown())
 		})
 	}
@@ -510,13 +510,13 @@ func (c *testsClient) injectJobWatchEvent(job *batchv1.Job) {
 	c.PrependWatchReactor(jobResource, watchReactionFn)
 }
 
-func (c *testsClient) injectWatchWithJobDeleteEvent(namespace, name string) {
+func (c *testsClient) injectWatchWithJobDeleteEvent(name string) {
 	watchReactionFn := func(action clienttesting.Action) (bool, watch.Interface, error) {
 		watcher := watch.NewRaceFreeFake()
 		watcher.Delete(&batchv1.Job{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:            name,
-				Namespace:       namespace,
+				Namespace:       testTargetNs,
 				ResourceVersion: "123",
 			},
 		})
--- a/manifests/kiagnose.yaml
+++ b/manifests/kiagnose.yaml
@@ -41,6 +41,14 @@ rules:
       - create
       - delete
       - watch
+  - apiGroups: [ "" ]
+    resources: [ "pods" ]
+    verbs:
+      - list
+  - apiGroups: [ "" ]
+    resources: [ "pods/log" ]
+    verbs:
+      - get
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding


ormergi:
> /hold we may want to consider using a more specific label selector in case there are more than 1 checkups of the same kind running on the namespace.

I think that the current state is good enough as the checkup object hold a refrenace of its Job object.
There is no scenario where a checkup prints the wrong job logs, right?

kiagnose/internal/launcher/launcher.go
```diff
@@ -103,6 +104,11 @@ func (l Launcher) Run() (runErr error) {
 		statusData.FailureReason = append(statusData.FailureReason, resultData.FailureReason)
 	}
 
+	if err = l.checkup.Logs(); err != nil {
```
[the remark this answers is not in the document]
orelmisan: I support this way of collecting logs as a distinct step, similar to results collection.
IIUC, the logs well be collected only if the checkup ran successfully - will the logs be collected (for example) in case of a timeout?

kiagnose/internal/launcher/launcher_test.go
```diff
@@ -256,6 +256,10 @@ func (s checkupStub) Results() (results.Results, error) {
 	return s.results, nil
 }
 
+func (s checkupStub) Logs() error {
+	return nil
```
orelmisan, line 260: Please consider testing the case when logs collection fails.

orelmisan (commented):
Thank you for the PR @RamLavi.

kiagnose/internal/launcher/launcher.go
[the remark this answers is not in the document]
RamLavi: I think simplifying reasoning is not relevant here. The reason I put it here is that we want the logs to be collected even if the Run was not successful.
I did get the position wrong - should be in the defer function if we want it to report in case of failure..

kiagnose/internal/checkup/job/job.go
```diff
@@ -113,6 +114,40 @@ func DeleteAndWait(client kubernetes.Interface, job *batchv1.Job, timeout time.D
 	return err
 }
 
+func PrintLogs(client kubernetes.Interface, job *batchv1.Job) (string, error) {
```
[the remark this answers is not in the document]
RamLavi: changed to `GetLogs`, this complies with the pod's `GetLogs` API

kiagnose/internal/launcher/launcher_test.go
RamLavi: that is a very good point, but I think it depends on what's the desired behavior.
the failure unit test also checks that error returned (/injected) is represented on the `FailureReason`.
However, do we want failure to fetch the logs to be considered as test failure? IMO the answer is no.
This is similar to the behavior of Report() failure that is currently added to the "run" error but is not added as a FailureReason... 
But indeed this deserves discussing, so WDYT?

orelmisan: `Report()` is different from `Results()` and `Logs()` in a way that it is the final action kiagnose takes before exiting the program.
This is why a failure in `Report()` could only be logged.

IMO, a failure to fetch the result or the logs should be treated the same.

RamLavi:
> > /hold we may want to consider using a more specific label selector in case there are more than 1 checkups of the same kind running on the namespace.
> 
> I think that the current state is good enough as the checkup object hold a refrenace of its Job object. There is no scenario where a checkup prints the wrong job logs, right?

I don't think so, BUT in another topic, I see that the pod could die before the collection is made... so we may need to do one of these options:
1. add a finalizer to the pod's job. which. we. do. NOT. want.
2. maybe do a continuous stream during the test lifecycle.. need to check if that's a thing..

orelmisan:
> > > /hold we may want to consider using a more specific label selector in case there are more than 1 checkups of the same kind running on the namespace.
> > 
> > 
> > I think that the current state is good enough as the checkup object hold a refrenace of its Job object. There is no scenario where a checkup prints the wrong job logs, right?
> 
> I don't think so, BUT in another topic, I see that the pod could die before the collection is made... so we may need to do one of these options:
> 
> 1. add a finalizer to the pod's job. which. we. do. NOT. want.
> 2. maybe do a continuous stream during the test lifecycle.. need to check if that's a thing..

`kubectl logs job.batch/<Kiagnose-job-name> -n kiagnose` is able to fetch the logs after the Job had finished.
It it interesting to see how it does it.

kiagnose/internal/launcher/launcher.go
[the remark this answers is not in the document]
ormergi: > I support this way of collecting logs as a distinct step, similar to results collection. IIUC, the logs well be collected only if the checkup ran successfully - will the logs be collected (for example) in case of a timeout?

Unlike checkup result, which is considered as an API, collecting the checkup logs is eventually a log message and should be treated as a non-disruptive operation (i.e: should not return an error)

> I think simplifying reasoning is not relevant here. The reason I put it here is that we want the logs to be collected even if the Run was not successful. I did get the position wrong - should be in the defer function if we want it to report in case of failure..

This can be achieved also if it's done at `checkup.Run()`, the deferred function is already bloated let's avoid adding more to it when its possible.

Also, you can a do a nice improvement by following the checkup logs and printing them "live".

kiagnose/internal/checkup/job/job.go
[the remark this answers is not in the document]
ormergi: note that `job` is another abstraction level on top of k8s client it doesn't have to follow the original function names.
I suggested calling it just "Logs" since "Get" doesn't give more information other than the obvious :slightly_smiling_face: (still a nit though..)

kiagnose/internal/launcher/launcher_test.go
ormergi: Note that when Report() fails it means we could not update the results configmap thus there is no need to add it to `failureReason` because it won't reflect at the results configmap.
It is printed to stderr instead.

Printing the checkup logs should be non-disruptive operation, when it fails it should not fail the checkup (no need to add the error to failureReason). 
The error can be printed on the logs as well.

ormergi:
> I don't think so, BUT in another topic

I checked the code, the checkup holds the create the Job object, no way for it to print other job logs.

I see that the pod could die before the collection is made... so we may need to do one of these options:
> 
>     1. add a finalizer to the pod's job. which. we. do. NOT. want.
>
>     2. maybe do a continuous stream during the test lifecycle.. need to check if that's a thing..

I am not sure what you mean by die as long it exists we can access the logs.
When the checkup job is completed (failed or not), pods are not disposed of until it's `activeDeadlineSeconds` reach (and since restartPolicy=Never is set).

kiagnose/internal/launcher/launcher.go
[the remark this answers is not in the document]
RamLavi: > > I support this way of collecting logs as a distinct step, similar to results collection. IIUC, the logs well be collected only if the checkup ran successfully - will the logs be collected (for example) in case of a timeout?
> 
> Unlike checkup result, which is considered as an API, collecting the checkup logs is eventually a log message and should be treated as a non-disruptive operation (i.e: should not return an error)

I agree that it should not be disrupting. Can you explain why doesn't `Logs` deserve to be a proper API?
> 
> > I think simplifying reasoning is not relevant here. The reason I put it here is that we want the logs to be collected even if the Run was not successful. I did get the position wrong - should be in the defer function if we want it to report in case of failure..
> 
> This can be achieved also if it's done at `checkup.Run()`, the deferred function is already bloated let's avoid adding more to it when its possible.

We may be talking about different deffers, I mean the collect the logs just before the teardown. That deffer is not bloated at all. Moreover, I think that Logs should be separate from Run() cause it is logically separate - in terms that it should should regardless of what happens on Run(). Therefor imo it should be outside of it
> 
> Also, you can a do a nice improvement by following the checkup logs and printing them "live".

I could, but I don't see a good reason to do so. we have all the means to collect the logs just before the teardown, so why would we want to complicate things?

kiagnose/internal/launcher/launcher_test.go
RamLavi: I agree, thus there is no reason to add the unit test for it.

kiagnose/internal/checkup/job/job.go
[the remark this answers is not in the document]
RamLavi: that's true, but out abstraction sort of adds a new function `GetLogs` that is not in the original API, but behaves the same as Pod.GetLogs(). So I think the name is suiting. I hope it doesn't break any rules..

RamLavi:
> > I don't think so, BUT in another topic
> 
> I checked the code, the checkup holds the create the Job object, no way for it to print other job logs.
> 
> > I see that the pod could die before the collection is made... so we may need to do one of these options:
> > ```
> > 1. add a finalizer to the pod's job. which. we. do. NOT. want.
> > 
> > 2. maybe do a continuous stream during the test lifecycle.. need to check if that's a thing..
> > ```
> 
> I am not sure what you mean by die as long it exists we can access the logs. When the checkup job is completed (failed or not), pods are not disposed of until it's `activeDeadlineSeconds` reach (and since restartPolicy=Never is set).

you're right. I was seeing something else that looked like the job died before collection:
When we the checkup job times-up - the the job is lost [here](https://github.com/kiagnose/kiagnose/blob/main/kiagnose/internal/checkup/job/job.go#L66) - I think we should no return nil in this case - WDYT?

RamLavi:
This [change](https://github.com/kiagnose/kiagnose/compare/33958bf02d923eef3c93b9bae43fda534915b5fa..f1e7095f83736fe3f46acfc34e0e064d6cfd5726) mainly introduces a new commit that makes sure the job object is not lost in case of timeout

/unhold

RamLavi:
not sure why unit test lane `checks / Unit Test (1.19) (pull_request) ` fails, it passed locally

orelmisan:
> not sure why unit test lane `checks / Unit Test (1.19) (pull_request) ` fails, it passed locally

https://github.com/kiagnose/kiagnose/actions/runs/3149223951/jobs/5120610529#step:4:112

kiagnose/internal/launcher/launcher.go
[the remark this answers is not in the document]
ormergi: > I agree that it should not be disrupting. Can you explain why doesn't Logs deserve to be a proper API?

By "results" I meant the actual results fields and not `Results` function.
Logs are nice to have they should not make the checkup fail (i.e: exit with an error).

> We may be talking about different deffers, I mean the collect the logs just before the teardown. That deffer is not bloated at all.

I think it is, it has logic that is not easy to grasp.

>  Moreover, I think that Logs should be separate from Run() cause it is logically separate - in terms that it should should regardless of what happens on Run(). Therefor imo it should be outside of it

Collecting logs must consider Run() result and be part of it, there is the case where Run() failed and there is nothing to collect (e.g: Job did not started).

> I could, but I don't see a good reason to do so. we have all the means to collect the logs just before the teardown, so why would we want to complicate things?

It could give the user immediate feedback regarding the checkup status instead of waiting until the checkup is finished, failed, or timed out to get these logs.
If you think that it takes more time and work we can do it later, it should not complicate things.

kiagnose/internal/launcher/launcher_test.go
ormergi: > I agree, thus there is no reason to add the unit test for it.
If you have no objections, I'll close this thread

From what I understand for the code, as long as the current implementation stays, when Logs() fails the checkup also fails (return with succeeded=false), which IMO should have a unit test.
But I think we agreed that collecting the Job logs should not cause the checkup to fail, so I guess this is going to change.

kiagnose/internal/checkup/checkup.go
```diff
@@ -117,14 +117,17 @@ func (c *Checkup) Setup() error {
...
+		c.job = nil
 		return fmt.Errorf("%s: %v", errPrefix, err)
 	}
+	c.job = updateJob
 
-	if c.job, err = job.WaitForJobToFinish(c.client, c.job, c.jobTimeout); err != nil {
+	if updateJob, err = job.WaitForJobToFinish(c.client, c.job, c.jobTimeout); err != nil {
 		return fmt.Errorf("%s: %v", errPrefix, err)
 	}
+	c.job = updateJob
```
ormergi, line 131: I would change `job.WaitForJobToFinish` to not return nil or change it to not return an updated Job object + fetching the Job at the caller.

kiagnose/internal/checkup/job/job.go
```diff
@@ -113,6 +114,41 @@ func DeleteAndWait(client kubernetes.Interface, job *batchv1.Job, timeout time.D
...
+		return "", err
+	}
+	defer podLogs.Close()
+
+	buf := new(bytes.Buffer)
+	_, err = buf.ReadFrom(podLogs)
+	if err != nil {
+		return "", err
+	}
+	return buf.String(), nil
```
ormergi, line 135: Would that work https://github.com/ormergi/k8s-checkup-framework/blob/main/cmd/checkup-launcher.go#L167-L16 instead of working with Stream()?

kiagnose/internal/checkup/job/job.go
```diff
@@ -113,6 +114,41 @@ func DeleteAndWait(client kubernetes.Interface, job *batchv1.Job, timeout time.D
...
+}
+
+func getPodNameByJob(client kubernetes.Interface, jobName, jobNamespace string) (string, error) {
+	const JobNameLabel = "job-name"
+	jobLabel := k8slabels.Set{JobNameLabel: jobName}
+	podList, err := client.CoreV1().Pods(jobNamespace).List(context.Background(), metav1.ListOptions{LabelSelector: jobLabel.String()})
+	if err != nil {
+		return "", err
```
ormergi, line 143: nit: It wpuld help debugging if this error will be prefixed with `failed to find job's <job-ns/job-name> pod`.

kiagnose/internal/launcher/launcher.go
```diff
@@ -81,6 +82,10 @@ func (l Launcher) Run() (runErr error) {
 	}
 
 	defer func() {
+		if err := l.checkup.Logs(); err != nil {
```
ormergi, line 85: Note that since Logs() error is appended to `errorPool` will cause the checkup to fail as well.

kiagnose/internal/launcher/launcher.go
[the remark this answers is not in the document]
RamLavi: >Logs are nice to have they should not make the checkup fail (i.e: exit with an error).

fixed.

>>We may be talking about different deffers, I mean the collect the logs just before the teardown. That deffer is not bloated at all.

>I think it is, it has logic that is not easy to grasp.

Again - I don't think you are looking in the right defer. Please take another look as the location is changed.

>Collecting logs must consider Run() result and be part of it, there is the case where Run() failed and there is nothing to collect (e.g: Job did not started).

I am already checking the scenario of Run() failing, but I don't see why it should be coupled inside Run(). please provide reasons.

>>I could, but I don't see a good reason to do so. we have all the means to collect the logs just before the teardown, so why would we want to complicate things?

>It could give the user immediate feedback regarding the checkup status instead of waiting until the checkup is finished, failed, or timed out to get these logs.
If you think that it takes more time and work we can do it later, it should not complicate things.

I don't think it's complicated to actually do. I think it's unnecessarily complicated to have. Worse case scenario the logs are collected after the timeout.

kiagnose/internal/launcher/launcher_test.go
RamLavi: I stand corrected. when I said it is done after - it was when the Logs was located after `FailureReason` was set, but that is no longer the scenario. changed it to only report a log

kiagnose/internal/checkup/checkup.go
```diff
@@ -117,14 +117,17 @@ func (c *Checkup) Setup() error {
...
 	const errPrefix = "run"
 	var err error
-
-	if c.job, err = job.Create(c.client, c.job); err != nil {
+	var updateJob *batchv1.Job
+	if updateJob, err = job.Create(c.client, c.job); err != nil {
+		c.job = nil
 		return fmt.Errorf("%s: %v", errPrefix, err)
 	}
+	c.job = updateJob
```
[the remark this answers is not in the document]
RamLavi: I disagree. and I think this is a good approach in general. We should not assume the we the job is return with a defined value if there is an error.

kiagnose/internal/checkup/checkup.go
RamLavi: I think the behavior of a function returning nil if err!=nil should be consistent throughout. we can control what we do with this object, in an explicit manner, like here.

kiagnose/internal/checkup/job/job.go
RamLavi: it can be done either way, but I like yours better. thanks!

kiagnose/internal/checkup/job/job.go
RamLavi: I would prefer this description to come out of the `List` err.. but since we don't control it - this may have some value - fixed

kiagnose/internal/launcher/launcher.go
RamLavi: as mentioned, fixed, thank!

kiagnose/internal/checkup/checkup.go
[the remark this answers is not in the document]
ormergi: It is up to how we implement `job.Create` (if an error occurred it returns nil) there are no assumptions, in this case this approach add noise we can avoid.

kiagnose/internal/checkup/checkup.go
ormergi: I agree on keeping it consistent, this is why I suggested changing `job.WaitForJobToFinish` to not fetch the updated object, and instead fetch the Job again at the caller.

Initially, it made sense that `job.WaitForJobToFinish` would return the updated Job instance, but looking at this again it seems there is a bug in the current implementation where the job won't be cleaned up when there is a timeout.

kiagnose/internal/checkup/checkup.go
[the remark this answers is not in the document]
RamLavi: I disagree

kiagnose/internal/checkup/checkup.go
RamLavi: when I think of it, it might be a bug
cause if the job timeouted, but returns nil, how is it removed on teardown?

RamLavi: nah it's all good

kiagnose/internal/checkup/checkup.go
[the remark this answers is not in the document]
ormergi: Lemme rephrase, `job.Create()` either return the job object or nil + error, there is no additional state.

I do not insist on this change.

kiagnose/internal/checkup/checkup.go
ormergi: You fixed it by assigning the job to an intermediate variable :slightly_smiling_face:

ormergi (dismissed):
Thanks for the changes.

kiagnose/internal/launcher/launcher.go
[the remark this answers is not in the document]
ormergi: > Again - I don't think you are looking in the right defer. Please take another look as the location is changed.
> 
I got it wrong sorry for the noise.

> I am already checking the scenario of Run() failing, but I don't see why it should be coupled inside Run(). please provide reasons.
> 
There is coupling in any case since its a method addressing its object field.

What I suggest is to fetch and print the checkup logs as part of Run() after at the time the Job is finished (before handling the wait error).
It will simplify the code as it won't be necessary to add Logs() methods to the launcher and handle its error.
And also won't print an error to logging in the scenario the Job doesn't exist.

Although I do not agree that the current implementation is the way to go, I do not insist on this change.

kiagnose/internal/checkup/checkup.go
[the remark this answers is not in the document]
RamLavi: Having both you and @orelmisan think it's the right thing to do, I'll fix it on Create() case..

kiagnose/internal/checkup/checkup.go
RamLavi: yep, I made it explicit..

ormergi (approved):

orelmisan (commented):
Thank you for the changes @RamLavi.

orelmisan (dismissed):
Feel free to address these nits in a follow-up PR.

kiagnose/internal/checkup/checkup.go
```diff
@@ -122,9 +122,12 @@ func (c *Checkup) Run() error {
 		return fmt.Errorf("%s: %v", errPrefix, err)
 	}
 
-	if c.job, err = job.WaitForJobToFinish(c.client, c.job, c.jobTimeout); err != nil {
+	var updateJob *batchv1.Job
```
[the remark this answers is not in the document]
RamLavi: done

kiagnose/internal/checkup/checkup.go
```diff
@@ -136,6 +137,21 @@ func (c *Checkup) Results() (results.Results, error) {
 	return results.ReadFromConfigMap(c.client, c.resultConfigMap.Namespace, c.resultConfigMap.Name)
 }
 
+func (c *Checkup) Logs() error {
+	var err error
+	var logs string
```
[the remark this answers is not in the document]
RamLavi: done

kiagnose/internal/checkup/job/job.go
```diff
@@ -113,6 +113,33 @@ func DeleteAndWait(client kubernetes.Interface, job *batchv1.Job, timeout time.D
...
+	}
+	rawLogs, err := client.CoreV1().Pods(job.Namespace).GetLogs(jobPodName, &corev1.PodLogOptions{}).DoRaw(context.Background())
+	if err != nil {
+		return "", fmt.Errorf("failed to find job's %s/%s pod: %v", job.Namespace, job.Name, err)
+	}
+
+	return string(rawLogs), nil
+}
+
+func getPodNameByJob(client kubernetes.Interface, jobName, jobNamespace string) (string, error) {
```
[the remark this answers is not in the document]
RamLavi: seems redundant, there is only one pod.

kiagnose/internal/checkup/job/job.go
```diff
@@ -113,6 +113,33 @@ func DeleteAndWait(client kubernetes.Interface, job *batchv1.Job, timeout time.D
 	return err
 }
 
+func GetLogs(client kubernetes.Interface, job *batchv1.Job) (string, error) {
+	jobPodName, err := getPodNameByJob(client, job.Name, job.Namespace)
+	if err != nil {
+		return "", err
+	}
+	rawLogs, err := client.CoreV1().Pods(job.Namespace).GetLogs(jobPodName, &corev1.PodLogOptions{}).DoRaw(context.Background())
```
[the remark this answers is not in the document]
RamLavi: future PR is fine by me

ormergi (approved):

orelmisan (approved):
Thank you @RamLavi

EdDev (approved):
Thanks!

Raising.

PR #177, opened 2022-09-29
docs, proposals: Kiagnose as an operator

ormergi:
A design proposal to transform Kiagnose to Kubernetes operator and transition to using a single checkup CRD and controller.

Rendered doc https://github.com/ormergi/kiagnose/blob/a0b41ebc9d005c2403bde3d2df712ba649dd67c1/docs/kiagnose-operator.md


orelmisan (commented):
Hi @ormergi, there are two kiagnose-operator.md files.
One under docs, and the other under docs/design-proposals.

docs/design-proposals/kiagnose-operator.md
```diff
@@ -0,0 +1,446 @@
...
+Given the following Namespace and ServiceAccount:
+```yaml
+---
+apiVersion: v1
+kind: Namespace 
+metadata:
+  name: echo-test
+---
+apiVersion: batch/v1
+kind: ServiceAccount
```
phoracek, line 130: Is a service account needed here? I can imagine that many checkups would not need to do any operation with the kube API (e.g. they just exercise services through their HTTP endpoints). In that case, `serviceAccountName` could be left out.

EdDev:

PR #178, opened 2022-10-03
docs, proposals: Kiagnose as an operator with mutli checkups CRDs

ormergi:
A design proposal to transform Kiagnose to Kubernetes operator and transition to using dedicated checkup CRD and controller for each type of checkup.

Rendered doc https://github.com/ormergi/kiagnose/blob/c4367bd20e0a328b3abbb2169165070fc28d05be/docs/proposals/kiagnose-operator-multi-checkups-crds.md

Signed-off-by: Or Mergi [ormergi@redhat.com](mailto:ormergi@redhat.com)


EdDev:

PR #181, opened 2022-10-18
design: checkup specification definition

EdDev:
Define a kiagnose *checkup* specification and a certification process, to standardize and assist in creating and consuming checkups.

### DIFF
--- a/docs/design-proposals/checkup_definition.md
+++ b/docs/design-proposals/checkup_definition.md
@@ -1,0 +1,169 @@
+Kiagnose Checkup Specification
+=
+
+Author: [Edward Haas](https://github.com/eddev)
+
+# Summary
+Define a kiagnose *checkup* specification and a certification process,
+to standardize and assist in creating and consuming checkups.
+
+# Motivation
+The kiagnose project has started off as a simple job that ran a checkup.
+
+The data inputted by the user and the results can be split in two categories:
+- Generic, common to all checkups.
+- Extended, unique and specific per each checkup.
+
+In order to allow kiganose checkups to integrate in the existing Kubernetes and OpenShift
+ecosystem, multiple approaches have been raised.
+The existing approach and the new suggestions focus on providing a running application (and controllers),
+either as a reference or as a service.
+
+Classic services for checkup execution & control encounter issues of permission management and elevation concerns.
+On the other hand, focusing on specific checkups as a stand-alone product, lack the ability to impose a common/generic
+API, which in turn prevents clients from operating on checkups independently of the checkup kind/type and require
+frequent intervention from the cluster admin.
+
+There is a need to find a solution which can impose hard requirements, like a common API but
+at the same time provide flexibility for vendors to create checkups as they see fit and per
+their ability.
+Placing a low entry bar for creating a checkup, opens up kiagnose for a larger community of checkup creators
+and in turn to a wider consumption potential.
+
+## Goal
+- Minimize the required knowledge to compose a checkup.
+- Provide as much independence to checkup writers as possible.
+- Assure consumers can identify and consume checkups regardless of type.
+- Minimize (or even eliminate) the need for a cluster-admin to approve each checkup.
+
+# Proposal
+
+## Definition of Personas
+- Kiagnose checkup users.
+- Kiagnose checkup authors.
+
+## User Stories
+As a checkup user:
+- I want to identify available checkups for me to run.
+- I want to run an individual checkup with default parameters on a specific target namespace.
+- I want to run an individual checkup with non-default parameters on a specific target namespace.
+- I want to observe the checkup progress and its final result.
+- I want to collect all relevant logs and/or artifacts after the execution.
+
+As a checkup author:
+- I want to have a specification that defines what I must conform to.
+- I want to have references to existing checkups.
+- I want to have tutorials for writing a checkup.
+- I want to use tooling and libraries that assist in writing a checkup.
+- I want to certify my checkup once (I think) it is ready.
+- I want to make my checkup available for deployment on K8S based clusters.
+
+## Solution Overview
+The proposal approach is focusing on defining a checkup through specification, allowing
+checkup authors to implement checkups per hard requirements and at the same time to give
+freedom on how they are implemented and extended.
+
+Checkup consumers on the other hand, can use the specifications to interact with checkups
+in a generic manner.
+
+De facto, the specification defines a minimum standard that is common to all checkups.
+To assure the checkup is conforming to the specification, a certification process is
+proposed.
+
+While specifications are enough to guide the creation of checkups, it would also be
+beneficial to provide tooling to assist in the checkup development.
+Such tooling may include libraries, scripts and tests.
+
+The implementation and its details are left to the checkup author.
+Details about using a controller, operator or any other means is left to the author
+discretion, as long as the solution conforms to the specification.
+
+The specification includes mandatory and optional fields.
+Mandatory fields must be implemented, while optional fields are left to the checkup author
+discretion.
+
+> **_NOTE:_**  The proposal has been influenced by the [CNI](https://github.com/containernetworking/cni)
+> project.
+
+## Specification
+
+### Version
+The kiagnose checkup specification is versioned in order to allow checkup vendors to express
+to which specification their checkups conform to.
+Consumer clients that interact with checkups may use this information to access and operate checkups.
+
+### Format
+Data to and from a checkup will be based on JSON.
+
+To allow for a standard layer of communication, a Custom Resource (CR) is chosen to pass the data.
+
+### Custom Resource Definition
+A CRD will be provided to specify the fields that the specification requires.
+Checkup authors are free to extend the data with additional fields.
+
+> **_NOTE:_** The usage of a CRD does not imply that there is a need for a controller
+> when implementing the checkup. The CRD is used here to define an API, nothing more.
+
+> **_NOTE:_** Checkups may use their own CRD/s for their operation. The CRD specified
+> in this section is to be used to interact with the checkup clients.
+
+Checkup deployments should make sure the CRD exists and deploy it if necessary.
+
+```yaml
+apiVersion: kiagnose.io/v1alpha1
+kind: Checkup
+metadata:
+  name: CheckupExample
+spec:
+  version: <Checkup kind name and version (e.g. `foo/v1`)>
+  image: <checkup image name>
+  timeout: <timeout to wait for checkup to finish [min]>
+  serviceAccountName: <SA that provides the required permissions to run the checkup>
+  schema: <Optional JSON-schema definition that describes the extended fields under the `param` field>
+  param: <Optional key/value fields that extends the base CRD, specific per checkup kind>
+```
+
+> **_NOTE:_** The optional `schema` field value contains a JSON-schema that describe the structure of the extension
+> fields a specific checkup introduces. The format is the same as the one used in a CRD manifest, at the 
+> [openAPIV3Schema](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#jsonschemaprops-v1-apiextensions-k8s-io)
+> field.
+> Clients may use this schema to learn what are the supported extension fields.
+
+## Tooling
+Tools that assist in creating checkups highly depends on how the checkup is implemented.
+In case one wants to develop a controller that monitors a specific checkup kind, scripts and libraries
+can be provided to assist authors in processing common tasks (e.g. reconcile loop template, watching the CR, validating
+base fields, enforcing timeout).
+
+E2E tests which validate the checkup specification can be included, allowing authors to validate their checkups before
+proceeding to certify them.
+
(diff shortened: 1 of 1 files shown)


docs/design-proposals/checkup_definition.md
```diff
@@ -0,0 +1,169 @@
...
+at the same time provide flexibility for vendors to crate checkups as they see fit and per
+their ability.
+Placing a low entry bar for creating a checkup, opens up kiagnose for a larger community of checkup creators
+and in turn to a wider consumption potential.
+
+## Goal
+- Minimize the required knowledge to compose a checkup.
+- Provide as much independence to checkup writers as possible.
+- Assure consumers can identify and consume checkups regardless of type.
+- Minimize (or even eliminate) the need for a cluster-admin to approve each checkup.
```
phoracek, line 37: We must aim to eliminate it. It is unrealistic to think that a cluster admin of a big private cloud would spend time reviewing checkups

docs/design-proposals/checkup_definition.md
```diff
@@ -0,0 +1,169 @@
...
+Checkup deployments should make sure the CRD exists and deploy it if necessary.
+
+```yaml
+apiVersion: kiagnose.io/v1alpha1
+kind: Checkup
+metadata:
+  name: CheckupExample
+spec:
+  version: <Checkup kind name and version (e.g. `foo/v1`)>
+  image: <checkup image name>
```
phoracek, line 119: Follow-up from our conversation: This image could be any garbage from the internet (by design, we don't want to keep this flexible). In a way it is like `image` passed to a Pod or ISO to a VM. We can have a centralized trusted operator requiring admin to deploy it (like kubevirt), but then every checkup (think VM or pod) can be instantiated by untrusted users from unstrusted sources. The only thing keeping evil at bay is RBAC.

EdDev: Yes, I also think this is a security issue.
It is also a maintenance problem, e.g. upgrading a checkup application requires a change in the image value, left for the one that executes the checkup. But such an upgrade should be left (at least) for the project admin and not for the executor.

I tend here to remove this image field, but then the problem of where to place this information is opening up.
We cannot place it in a control-plane namespace (i.e. at the cluster level) because then we force the cluster-admin to be involved in a checkup deployment.
We could create a configmap or a new CR for this information, at the target namespace, where each checkup must be registered in with its version (name and version) and image.
Well, we could even avoid completely from specifying where or how this information is collected & consumed, we can just mention that it is up to the checkup to make it available somehow (and give examples).

docs/design-proposals/checkup_definition.md
```diff
@@ -0,0 +1,169 @@
...
+> **_NOTE:_** The usage of a CRD does not imply that there is a need for a controller
+> when implementing the checkup. The CRD is used here to define an API, nothing more.
+
+> **_NOTE:_** Checkups may use their own CRD/s for their operation. The CRD specified
+> in this section is to be used to interact with the checkup clients.
+
+Checkup deployments should make sure the CRD exists and deploy it if necessary.
+
+```yaml
+apiVersion: kiagnose.io/v1alpha1
```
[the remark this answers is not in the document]
phoracek: Design pattern section sounds great to me, thanks.

docs/design-proposals/checkup_definition.md
phoracek: +1 for adding it as a recommendation.

docs/design-proposals/checkup_definition.md
phoracek: I would argue that the image should be passed directly by the executor. That way we treat Checkup just like we would treat a Pod. If we eventually see a need to manage lifecycle of checkups, we could introduce something akin to OLM. Where OLM manages the lifecycle of pods, this higher level tool would manage checkup primitives.

docs/design-proposals/checkup_definition.md
```diff
@@ -0,0 +1,169 @@
...
+metadata:
+  name: CheckupExample
+spec:
+  version: <Checkup kind name and version (e.g. `foo/v1`)>
+  image: <checkup image name>
+  timeout: <timeout to wait for checkup to finish [min]>
+  serviceAccountName: <SA that provides the required permissions to run the checkup>
+  schema: <Optional JSON-schema definition that describes the extended fields under the `param` field>
+  param: <Optional key/value fields that extends the base CRD, specific per checkup kind>
+```
```
EdDev, line 124: The `status` section is missing.

docs/design-proposals/checkup_definition.md
EdDev: Allowing the execution of a pod directly, including the option to allow passing in the image indirectly, is in my view a very high permission. I think that most applications/solutions abstract the pod and do not allow such a direct approach.

The project-admin cannot control in such a case what is going to run, i.e. the user that triggered the checkup may actually run a malicious application. Why would we want to expose such a hole?

And to be clear here, if it is fine for the checkup executor to run a pod/job directly, then we do not need anything beyond what already exists today.

docs/design-proposals/checkup_definition.md
```diff
@@ -0,0 +1,169 @@
...
+The existing approach and the new suggestions focus on providing a running application (and controllers),
+either as a reference or as a service.
+
+Classic services for checkup execution & control encounter issues of permission management and elevation concerns.
+On the other hand, focusing on specific checkups as a stand-alone product, lack the ability to impose a common/generic
+API, which in turn prevents clients from operating on checkups independently of the checkup kind/type and require
+frequent intervention from the cluster admin.
+
+There is a need to find a solution which can impose hard requirements, like a common API but
+at the same time provide flexibility for vendors to crate checkups as they see fit and per
```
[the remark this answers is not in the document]
EdDev: Done

docs/design-proposals/checkup_definition.md
```diff
@@ -0,0 +1,169 @@
...
+their ability.
+Placing a low entry bar for creating a checkup, opens up kiagnose for a larger community of checkup creators
+and in turn to a wider consumption potential.
+
+## Goal
+- Minimize the required knowledge to compose a checkup.
+- Provide as much independence to checkup writers as possible.
+- Assure consumers can identify and consume checkups regardless of type.
```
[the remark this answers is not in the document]
EdDev: I intended for checkup type/kind/version. What wording would you suggest here to express it better?

docs/design-proposals/checkup_definition.md
phoracek: > Allowing the execution of a pod directly, including the option to allow passing in the image indirectly, is in my view a very high permission. I think that most applications/solutions abstract the pod and do not allow such a direct approach.

Applications would run Deployments, which also expose the image directly. They may manage their lifecycle with external tooling, but within the cluster, it would still be a Deployment creating a Pod. If we want higher level tools, that's fine for me, but we should not skip the primitives.

> 
> The project-admin cannot control in such a case what is going to run, i.e. the user that triggered the checkup may actually run a malicious application. Why would we want to expose such a hole?

IIUIC if they want to prevent people from running untrusted images, they can restrict registries that can be used. Anything the user runs would be still bound to the namespace an its RBAC, so it should not concern the cluster admin. The project admin must decide whether they trust the image they are passing in. Again, this is not different from Pods.

> 
> And to be clear here, if it is fine for the checkup executor to run a pod/job directly, then we do not need anything beyond what already exists today.

It is ok to do that, if we define an interface for checkup images. It would be better however, to have a specialized API (a CR) serving as a frontend for these pods.

orelmisan: I'm looking at the Job API [specification](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#job-v1-batch), and according to this thread, it sounds like a reasonable API.
The following fields could be specified:
- Image (even more than one if we really need).
- Timeout (activeDeadlineSeconds).
- ServiceAccount name.
- Arbitrary input parameters (as environment variables / files on the file system).

The only non straightforward issue is how to get arbitrary outputs when the checkup is finished.

Do we really need another abstraction for simple checkups?

EdDev: > Applications would run Deployments, which also expose the image directly. Users may manage their lifecycle with external tooling, but within the cluster, it would still be a Deployment creating a Pod. If we want higher level tools, that's fine for me, but we should not skip the primitives.

If the application has an API (e.g. CR), then I would not expect from it to expose the image.
One creates a CR, the application (usually as a controller) creates the deployment/job/pod, abstracting what it runs.
There is a good reason to do that, beyond permissions... it allows upgrades (i.e. use a newer image in the background).

I hold the opinion that primitives are for higher abstractions to use, not to be used directly.
Once permissions are given for an application (e.g. a checkup), it would be safer to control what it uses beyond the restrictions you mentioned. Be it because it is more secure, but also because it allows its operational lifecycle to be managed (automatically).

To be clear, I am not against using images directly, that is up to the checkup author/vendor to decide.
I just do not want to restrict it to do so. Placing the image as part of the checkup API/interface, implies one is expected to provide the image each time an execution is triggered. But I think, if the image is to be provided, then it should not be provided through the checkup API, it can be passed through the means it is triggered (e.g. the pod/job that was directly created to run it).

> > The project-admin cannot control in such a case what is going to run, i.e. the user that triggered the checkup may actually run a malicious application. Why would we want to expose such a hole?
> 
> IIUIC if they want to prevent people from running untrusted images, they can restrict registries that can be used. Anything the user runs would be still bound to the namespace an its RBAC, so it should not concern the cluster admin. The project admin must decide whether they trust the image they are passing in. Again, this is not different from Pods.

Yes, all this can be done, but I think it is complimentary and provides protections at different layers of the solution.
I also have trouble understanding the reasoning of defining a CR and allowing arbitrary images to run there. Unlike a pod (or a Deployment) this CR is specific and is expected to be processed by a specific image, not arbitrary ones.

> > And to be clear here, if it is fine for the checkup executor to run a pod/job directly, then we do not need anything beyond what already exists today.
> 
> It is ok to do that, if we define an interface for checkup images. It would be better however, to have a specialized API (a CR) serving as a frontend for these pods.

The value of having the CR is its API, the API in turn allows to structure well and abstract the data needed to run a checkup. Exposing the image hurts the abstraction, disallowing solutions that include a operational lifecycle management (Operators) and therefore I prefer it out.

If the checkup is designed to be triggered by jobs/pods, then the opportunity to input the image is there.
If the checkup is designed to be automatically triggered by a controller that monitors the CR/s, then which image to use can be embedded into that controller logic.
If the last option uses a generic controller, that does not have this image information, then I would expect it to access some external registry/DB and acquire that information.

I wander how a client, e.g. a UI client, that is suppose to identify checkup CR/s is expected to help trigger individual checkups in a generic manner. If the image information is required, then the UI frontend will be a simple YAMEL editor, in which the user that wants to trigger the checkup will gave to input the image (that he is suppose to copy from somewhere?). I have hard time to image it this way.

EdDev: > I'm looking at the Job API [specification](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#job-v1-batch), and according to this thread, it sounds like a reasonable API. The following fields could be specified:
> 
>     * Image (even more than one if we really need).
> 
>     * Timeout (activeDeadlineSeconds).
> 
>     * ServiceAccount name.
> 
>     * Arbitrary input parameters (as environment variables / files on the file system).
> 
> 
> The only non straightforward issue is how to get arbitrary outputs when the checkup is finished.
> 
> Do we really need another abstraction for simple checkups?

An interface, i.e. API, is aimed to abstract lower primitives.
A pod is a primitive that is aimed to run arbitrary workloads, when we define a CRD we want usually to run a more specific workload, which in this case is a checkup.
I expressed why I think we should not expose the image field in my previous answer.

orelmisan: > > I'm looking at the Job API [specification](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#job-v1-batch), and according to this thread, it sounds like a reasonable API. The following fields could be specified:
> > ```
> > * Image (even more than one if we really need).
> > 
> > * Timeout (activeDeadlineSeconds).
> > 
> > * ServiceAccount name.
> > 
> > * Arbitrary input parameters (as environment variables / files on the file system).
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The only non straightforward issue is how to get arbitrary outputs when the checkup is finished.
> > Do we really need another abstraction for simple checkups?
> 
> An interface, i.e. API, is aimed to abstract lower primitives. A pod is a primitive that is aimed to run arbitrary workloads, when we define a CRD we want usually to run a more specific workload, which in this case is a checkup. I expressed why I think we should not expose the image field in my previous answer.

I understand it.
This is why CRDs go hand in hand with a custom-controller.
If the trigger is different from a custom-controller, a Pod or a Job should point to the CR, and IMO the abstraction looses its strength because the user should define many parameters in the Job as well.

EdDev:

PR #183, opened 2022-10-23
design: Deploy & run kiagnose checkups in target namespace

EdDev:
Reduce permissions for Kiagnose deployment and execution to a namespace-admin (instead of a cluster-admin).

### DIFF
--- a/docs/design-proposals/kiagnose-per-namespace.md
+++ b/docs/design-proposals/kiagnose-per-namespace.md
@@ -1,0 +1,176 @@
+Kiagnose Per Namespace
+=
+
+# Summary
+Reduce permissions for Kiagnose checkup deployment and execution to a namespace-admin (instead of
+a cluster-admin).
+
+# Motivation
+Currently, Kiagnose requires a cluster-admin to deploy its core component and execute individual checkups.
+
+The requirement for a cluster-admin (or a user with pod execution on the kiagnose control namespace)
+to execute a checkup is too limiting and creates a strong dependency on the cluster-admin for each
+checkup.
+
+In addition, the two steps execution that exists today, where a kiagnose application runs a checkup
+application becomes redundant. Once all components have the same permissions (e.g. namespace-admin),
+there is no strong benefit to split them as independent applications.
+
+## Goal
+- Deployment and execution of checkups should require only namespace-admin permissions and below
+  (i.e. eliminate the need for a cluster-admin intervention).
+- Simplify and unify API definition for checkups and their clients.
+- Minimize the required knowledge to compose a checkup.
+
+## Non-Goal
+- Handle Kiagnose checkup life-cycle (operation).
+
+# Proposal
+
+## Definition Of Personas
+- Kiagnose checkup users.
+- Kiagnose checkup authors.
+
+## User Stories
+As a checkup user (that is **not** a cluster-admin):
+- I want to run an individual checkup with optional parameters on a specific target namespace.
+- I want to observe the checkup progress and its final result.
+- I want to collect all relevant logs and/or artifacts after the execution.
+
+As a checkup author:
+- I want to have an API specification that defines what I must conform to.
+- I want to have references to existing checkups.
+- I want to have tutorials for writing a checkup.
+- I want to use tooling and libraries that assist in writing a checkup.
+- I want to certify my checkup once (I think) it is ready.
+- I want to make my checkup available for deployment on K8s based clusters.
+
+## Solution Overview
+The described proposal suggests to deploy and execute checkups in the same namespace, where the
+checkups are aimed to run in (AKA *target* namespace).
+
+Instead of deploying kiagnose framework in a dedicated control plane namespace and starting
+a checkup execution by defining a job in that same namespace, the deployment and execution
+will move to the target namespace, which in turn will drop the need for a cluster-admin intervention.
+
+The solution includes the simplification of the execution flow and the removal of intermediate API
+definitions which existed so far. It is suggested to convert the existing kiagnose application
+(i.e. the framework) to a library that checkup authors may use to compose their application.
+By doing so, instead of two applications with two API definitions, a single application, which is
+the checkup, will be required (checkup authors may decide to expand to many applications per their need).
+
+As a subsequence, the results-configmap which has been used by the checkup to output the results, is removed.
+Checkups will directly write the results to the user-facing configmap.
+
+## API Changes
+
+The existing solution involved two API layers:
+- User facing API: Exposed to checkup users that execute them through the kiagnose framework.
+- Checkup API: The API checkups interact with the kiagnose framework.
+
+> **_NOTE:_** The user facing API format is left as a configmap.
+
+The checkup API is dropped and checkups are expected to directly interact with the user facing API.
+As with the existing solution, the API includes basic common fields that all checkups **must** implement
+and checkup-specific fields which extend it.
+
+In order to further simplify the API and take advantage of the new execution flow (single application),
+some base fields may be dropped now in the configmap:
+- spec.image
+- spec.serviceAccountName
+
+### Input Fields
+The user can configure the following under the `data` field:
+
+| Property                | Description                                                                                                                 | Mandatory | Remarks                               |
+|-------------------------|-----------------------------------------------------------------------------------------------------------------------------|-----------|---------------------------------------|
+| spec.timeout            | After how much time should Kiagnose stop the running checkup                                                                | Yes       | 5m, 1h etc                            |
+| spec.param.*            | Arbitrary strings that will be passed to the checkup as input parameters                                                    | No        | [0..N]                                |
+
+### Output Fields
+
+| Property                   | Description                                         | Mandatory | Remarks |
+|----------------------------|-----------------------------------------------------|-----------|---------|
+| status.succeeded           | Has the checkup succeeded                           | Yes       |         |
+| status.failureReason       | Failure reason in case of a failure                 | Yes       |         |
+| status.startTimestamp      | Checkup start timestamp                             | Yes       |         |
+| status.completionTimestamp | Checkup completion timestamp                        | Yes       |         |
+| status.result.*            | Arbitrary strings that were reported by the checkup | No        | [0..N]  |
+
+## Operational Flow
+The following main steps are expected to occur in order to use a checkup:
+- Deploy base kiagnose checkup permissions (in the target namespace).
+- Deploy checkup specific permissions (in the target namespace).
+- Create a checkup instance configmap (with relevant input data).
+- Create a Kubernetes Job to start the checkup execution.
+- Examine the checkup progress and results by viewing the initial configmap.
+- Delete the configmap once all data is collected for the specific checkup run.
+- Optionally, delete the base kiagnose and specific checkup permissions.
+
+> **_NOTE:_** Any permissions (e.g. SA, Role, RoleBinding) are under the namespace-admin
+> responsibility to examine, approve and deploy.
+
+### Checkup Execution Example
+
+- ConfigMap (input):
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: example-checkup-config
+data:
+  spec.timeout: 5m
+  spec.param.param_key_1: "value 1"
+  spec.param.param_key_2: "value 2"
+```
+
+- Jod (execution):
+```yaml
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: example-checkup
+spec:
+  backoffLimit: 0
+  activeDeadlineSeconds: 500
+  template:
+    spec:
+      serviceAccount: example-checkup-sa
+      restartPolicy: Never
+      containers:
+        - name: checkup
+          image: quay.io/kiagnose/example-checkup:main
+          imagePullPolicy: Always
+          env:
+            - name: CONFIGMAP_NAMESPACE
+              value: <target namespace>
+            - name: CONFIGMAP_NAME
+              value: example-checkup-config
+```
+
+- ConfigMap (output):
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: example-checkup-config
+data:
+  spec.timeout: 5m
+  spec.param.param_key_1: "value 1"
+  spec.param.param_key_2: "value 2"
+
+  status.succeeded: "true"
+  status.failureReason: ''
+  status.startTimestamp: "2022-05-25T11:53:49Z"
+  status.completionTimestamp: "2022-05-25T11:54:46Z"
+  status.result.key1: "result 1"
+  status.result.key2: "result 2"
+```
+
+> **_NOTE:_** It is recommended to split the role of the base Kiagnose and the checkup extensions.
+> E.G. the base Kiagnose role includes rules to access configmaps for read and write.
+
+> **_NOTE:_** It is recommended to specify a timeout in the Job spec (`activeDeadlineSeconds`),
+> assuring the Job finishes eventually.
+> When specified, it should be higher than the timeout inputted in the ConfiMap (`spec.timeout`),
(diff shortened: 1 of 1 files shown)


docs/design-proposals/kiagnose-per-namespace.md
```diff
@@ -0,0 +1,164 @@
...
+some base fields may be dropped now in the configmap:
+- spec.image
+- spec.serviceAccountName
+
+### Input Fields
+The user can configure the following under the `data` field:
+
+| Property                | Description                                                                                                                 | Mandatory | Remarks                               |
+|-------------------------|-----------------------------------------------------------------------------------------------------------------------------|-----------|---------------------------------------|
+| spec.timeout            | After how much time should Kiagnose stop the running checkup                                                                | Yes       | 5m, 1h etc                            |
```
orelmisan, line 84: I suggest to remove this field as well.
IMO it should go to the Job `spec.activeDeadlineSeconds` field, in order to get an automatic timeout feature "for free".

docs/design-proposals/kiagnose-per-namespace.md
```diff
@@ -0,0 +1,164 @@
+Kiagnose Per Namespace
+=
+
+# Summary
+Reduce permissions for Kiagnose deployment and execution to a namespace-admin (instead of
```
orelmisan, line 5: This sentence suggest that we are going to keep Kiagnose.
How about:

```suggestion
Reduce permissions for execution of Kiagnose-compliant checkup to a namespace-admin (instead of
```
?

docs/design-proposals/kiagnose-per-namespace.md
```diff
@@ -0,0 +1,164 @@
...
+a cluster-admin).
+
+# Motivation
+Kiagnose requires a cluster-admin to deploy its core component and execute individual checkups.
```
orelmisan, line 9:
```suggestion
Currently, Kiagnose requires a cluster-admin to deploy its core component and execute individual checkups.
```

docs/design-proposals/kiagnose-per-namespace.md
```diff
@@ -0,0 +1,164 @@
...
+
+The requirement for a cluster-admin (or a user with pod execution on the kiagnose control namespace)
+to execute a checkup is too limiting and creates a strong dependency on the cluster-admin for each
+checkup.
+
+In addition, the two steps execution that exists today, where a kiagnose application runs a checkup
+application becomes redundant. Once all components have the same permissions (e.g. namespace-admin),
+there is no strong benefit to split them as independent applications.
+
+, allowing all kiagnose components, including the checkups to require only namespace-admin
```
orelmisan, line 19: This sentence starts with a comma.

docs/design-proposals/kiagnose-per-namespace.md
```diff
@@ -0,0 +1,164 @@
...
+permissions, in both deployment and execution, is
```
orelmisan, line 20: This sentence doesn't end.

docs/design-proposals/kiagnose-per-namespace.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Goal
+- Deployment and execution of checkups should require only namespace-admin permissions and below
+  (i.e. eliminate the need for a cluster-admin intervention).
+- Simplify and unify API definition for checkups and their clients.
+- Minimize the required knowledge to compose a checkup.
```
orelmisan, line 26: How is this goal achieved?

docs/design-proposals/kiagnose-per-namespace.md
```diff
@@ -0,0 +1,164 @@
...
+
```
orelmisan, line 27: IMO we should add a non-goal:
"Handle checkup life-cycle"

docs/design-proposals/kiagnose-per-namespace.md
```diff
@@ -0,0 +1,164 @@
...
+- I want to observe the checkup progress and its final result.
+- I want to collect all relevant logs and/or artifacts after the execution.
+
+As a checkup author:
+- I want to have an API specification that defines what I must conform to.
+- I want to have references to existing checkups.
+- I want to have tutorials for writing a checkup.
+- I want to use tooling and libraries that assist in writing a checkup.
+- I want to certify my checkup once (I think) it is ready.
+- I want to make my checkup available for deployment on K8S based clusters.
```
orelmisan, line 46: nit:
```suggestion
- I want to make my checkup available for deployment on K8s based clusters.
```

docs/design-proposals/kiagnose-per-namespace.md
```diff
@@ -0,0 +1,164 @@
...
+|----------------------------|-----------------------------------------------------|-----------|---------|
+| status.succeeded           | Has the checkup succeeded                           | Yes       |         |
+| status.failureReason       | Failure reason in case of a failure                 | Yes       |         |
+| status.startTimestamp      | Checkup start timestamp                             | Yes       |         |
+| status.completionTimestamp | Checkup completion timestamp                        | Yes       |         |
+| status.result.*            | Arbitrary strings that were reported by the checkup | No        | [0..N]  |
+
+## Operational Flow
+The following main steps are expected to occur in order to use a checkup:
+- Deploy base kiagnose checkup permissions (in the target namespace).
```
orelmisan, line 99: Why is this necessary?

docs/design-proposals/kiagnose-per-namespace.md
```diff
@@ -0,0 +1,164 @@
...
+- Deploy checkup specific permissions (in the target namespace).
+- Create a checkup instance configmap (with relevant input data).
+- Create a kubernetes job to start the checkup execution.
```
orelmisan, line 102: nit:
```suggestion
- Create a Kubernetes Job to start the checkup execution.
```

docs/design-proposals/kiagnose-per-namespace.md
```diff
@@ -0,0 +1,164 @@
...
+- Examine the checkup progress and results by viewing the initial configmap.
```
orelmisan, line 103: We need to wait for the Job to finish first.

docs/design-proposals/kiagnose-per-namespace.md
```diff
@@ -0,0 +1,164 @@
...
+kind: ConfigMap
+metadata:
+  name: example-checkup-config
+data:
+  spec.timeout: 5m
+  spec.param.param_key_1: "value 1"
+  spec.param.param_key_2: "value 2"
+```
+
+- Pod (execution):
```
orelmisan, line 121: Job

orelmisan (commented):
Thank you for the PR @EdDev.

docs/design-proposals/kiagnose-per-namespace.md
EdDev: We expect SA, Roles and RoleBindings to have been deployed for the checkup to run.
While we do not provide them, they must be in place for the checkup to operate.

The `base` refers to permissions all checkups need and we can provide them as a manifest for authors to use (can be classified as tooling). I wanted to differentiate between the base and specific ones intentionally.

In any case, to avoid confusion, I added a note about permissions, so it will be clear who is responsible for them.

docs/design-proposals/kiagnose-per-namespace.md
EdDev: Done

docs/design-proposals/kiagnose-per-namespace.md
EdDev: I think it is should be understood implicitly from the sentence. It is also not exactly correct that the job needs to finish before the results are posted. There may be results, and the job may still be around for some moments before marked as finished/done/deleted.

docs/design-proposals/kiagnose-per-namespace.md
EdDev: Done

EdDev:
[change](https://github.com/kiagnose/kiagnose/compare/c7e3b4539a20d64c73ff990e8772b67014f556b2..eaf47b1b384a81a7246bc1f6364b16a0de1b49bc): Answered review comments.

docs/design-proposals/kiagnose-per-namespace.md
orelmisan: > I think that the checkup application needs to control this internally, so it can perform the necessary teardown steps (e.g. update the configmap).

In case the Job controller is controlling the timeout, the application should receive a SIGTERM and handle it.
Furthermore, the user shall have certainty the the checkup had terminated, even if it was buggy and didn't terminate by its own.

The checkup could be written an in arbitrary language, so providing tooling to handle the timeout from the inside is more complicated.

> Also, the Job is not a Kiagnose API, and we cannot enforce to have it.

The Job is a part of Kiagnose API because mandatory data is provided through it:
1. Checkup image.
2. ServiceAccountName.
3. Two environment variables to link to the user-supplied ConfigMap.

docs/design-proposals/kiagnose-per-namespace.md
orelmisan: The intermediate API was simpler to consume because it was only environment variables (which is language agnostinc).
The new checkup API requires a call to the K8s API and more complicated parsing of the input.
IMO the required knowledge to compose a checkup had increased a little.

docs/design-proposals/kiagnose-per-namespace.md
orelmisan: IIUC the consequences, you are suggesting to:
1. Have a base Role with the permissions to read/write the user-supplied ConfigMap.
2. Have a specific Role with the permissions the checkup requires.

Assuming the checkup is providing the ServiceAccount object, two RoleBinding objects will have to bind the ServiceAccount to their Roles.

IMO this could be simplified by requiring checkup authors to have the ConfigMap read/write permissions in their specific Role.

docs/design-proposals/kiagnose-per-namespace.md
orelmisan: BTW, the reasoning behind using a Job instead of a Pod was the ability to have a builtin timeout and cleanup.

EdDev: > In case the Job controller is controlling the timeout, the application should receive a SIGTERM and handle it. 

It would be "impolite" and risky to perform any calls to the kubernetes API one such a signal is received.
If the garbage collector controllers take care of something in the background automatically, that is fine. But if you want to run a long term I/O action in the signal handler, IMO it will not result in anything good.

> Furthermore, the user shall have certainty the the checkup had terminated, even if it was buggy and didn't terminate by its own.

And that is the purpose of the timeout set on the Job.
The responsibility for it to finish or not is on the one that started it and hopefully added such a timeout.
It is also on the client/user the responsibility to put the right values in the timeout in the configmap and job.

I can add such a note to the design and it should reach the documentation eventually.
But we have no way to validate it.

> The checkup could be written an in arbitrary language, so providing tooling to handle the timeout from the inside is more complicated.

I think that handling a signal is more complicated.

> > Also, the Job is not a Kiagnose API, and we cannot enforce to have it.
> 
> The Job is a part of Kiagnose API because mandatory data is provided through it:
> 
>     1. Checkup image.
> 
>     2. ServiceAccountName.
> 
>     3. Two environment variables to link to the user-supplied ConfigMap.

This is very borderline IMO and as part of the Kiagnose specification I do not want to be bound to it.
Someone may decide to create a controller instead of a Job, or one can use a Pod.
I have no interest to limit ourself to this method, and although you are right this is an API at the moment, it is not one that we are interested to enforce to be a must.

docs/design-proposals/kiagnose-per-namespace.md
EdDev: I am not forcing to use it in a specific way or the other, it is a recommendation.

My recommendation is to to use roles per what they functionally provide and not as all-in-one rules for a checkup. That way, you can specify a role for managing VMI/s, reading NAD/s, reading/writing configmap/s.
Having something modular is also more descriptive of the intention and also allows re-use.

It also makes sense to me to have a role that is the bare minimum needed for a checkup, from which, it can be extended.
If the author decides he wants to merge all in one role, I won't stop him, but I also won't be interested to encourage to do it.

EdDev:
[change](https://github.com/kiagnose/kiagnose/compare/eaf47b1b384a81a7246bc1f6364b16a0de1b49bc..d01bbe3f977208ed3518800d3e05d8aaf39d0014): Added a Job timeout/deadline and recommended to use it in a note with appropriate values.

docs/design-proposals/kiagnose-per-namespace.md
EdDev: > BTW, the reasoning behind using a Job instead of a Pod was the ability to have a builtin timeout and cleanup.

Pods have it as well (`activeDeadlineSeconds`).

orelmisan (dismissed):
Thank you @EdDev.

docs/design-proposals/kiagnose-per-namespace.md
EdDev: > Another argument is that if our recommendation is to use a Job, there are potentially two places where you could place your timeout, and the first to expire wins.

I have added the needed note.
Generally speaking, one timeout is for the checkup application to handle and the other timeout is for the Job execution to be protected. These are two different things.

docs/design-proposals/kiagnose-per-namespace.md
```diff
@@ -0,0 +1,173 @@
...
+| status.failureReason       | Failure reason in case of a failure                 | Yes       |         |
+| status.startTimestamp      | Checkup start timestamp                             | Yes       |         |
+| status.completionTimestamp | Checkup completion timestamp                        | Yes       |         |
+| status.result.*            | Arbitrary strings that were reported by the checkup | No        | [0..N]  |
+
+## Operational Flow
+The following main steps are expected to occur in order to use a checkup:
+- Deploy base kiagnose checkup permissions (in the target namespace).
+- Deploy checkup specific permissions (in the target namespace).
+- Create a checkup instance configmap (with relevant input data).
```
RamLavi, line 101: Do we plan to merge the user-supplied configmap and the results configmap?
it looks like it from the example, and if so we need to say it explicitly.

RamLavi (commented):
looks good, added some notes

docs/design-proposals/kiagnose-per-namespace.md
EdDev: It was already mentioned:
> The checkup API is dropped and checkups are expected to directly interact with the user facing API.

RamLavi: >The checkup API is dropped and checkups are expected to directly interact with the user facing API.

Well this is what is implied, but otherwise it is not mentioned explicitly that we are merging these 2 Configmaps. This should be explicit, and not only by this line/the example

RamLavi: I would add it in the first bullets in `Solution Overview`

docs/design-proposals/kiagnose-per-namespace.md
```diff
@@ -0,0 +1,173 @@
...
+- I want to have an API specification that defines what I must conform to.
+- I want to have references to existing checkups.
+- I want to have tutorials for writing a checkup.
+- I want to use tooling and libraries that assist in writing a checkup.
+- I want to certify my checkup once (I think) it is ready.
+- I want to make my checkup available for deployment on K8s based clusters.
+
+## Solution Overview
+The described proposal suggests to deploy and execute checkups in the same namespace, where the
+checkups is aimed to run in (AKA *target* namespace).
```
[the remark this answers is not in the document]
EdDev: Done

EdDev:
[change](https://github.com/kiagnose/kiagnose/compare/d01bbe3f977208ed3518800d3e05d8aaf39d0014..70eaacae8a96ef14e269ee1c7c8f89696766507a): Answered review comments.

orelmisan (approved):

PR #184, opened 2022-10-24
checkups, vm-latency: Drop unused preflight step

EdDev:
The preflight step, in which some validation was planned to be executed, is not being used.

One of the main reasons has to do with the realization that the validations would have required elevated permissions in the cluster/NS. Moreover, such checks may just be raceful due to the K8s operation mode (e.g. the existence of an entity may say nothing about its actual state 1msec later).

Therefore, this step is removed from the checkup.

PR #190, opened 2022-10-26
[release-0.2] checkups, vm-latency: Drop unused preflight step

EdDev:
Backport #184 to stable branch release 0.2.

The preflight step, in which some validation was planned to be executed, is not being used.

One of the main reasons has to do with the realization that the validations would have required elevated permissions in the cluster/NS. Moreover, such checks may just be raceful due to the K8s operation mode (e.g. the existence of an entity may say nothing about its actual state 1msec later).

Therefore, this step is removed from the checkup.


orelmisan (approved):
Thank you!

PR #191, opened 2022-10-26
checkups: Temporary remove the `echo` checkup

EdDev:
As part of the unification changes of kiagnose into the checkups, the `echo` checkup is temporary removed to allow the kiagnose module to be cleaned.

A following change will re-introduce the `echo` checkup in a standalone form (i.e. without the kiagnose application).


orelmisan (approved):
Thank you

PR #199, opened 2022-10-30
vmlatency, e2e: Upgrade K8s version to v1.25.3

orelmisan:
Currently, we are checking the VM latency checkup against K8s v1.24.

Upgrade kind to the latest version v1.17.0 - thus upgrading K8s cluster to current stable v1.25.3.
Upgrade kubectl utility to v1.25.3 to match.

The k8s [client-go](https://github.com/kubernetes/client-go) package will be upgraded in a follow-up PR.
We should consider pining and mirroring kind [node images](https://hub.docker.com/r/kindest/node/), which are hosted on hub.docker.com.


EdDev:
> We should consider pining and mirroring kind [node images](https://hub.docker.com/r/kindest/node/), which are hosted on hub.docker.com.

Agreed, the images should be pinned to assure we explicitly do the upgrade.
Although, I think that if we pin the kind version, we in practice pin the image as well (implicitly).

@ormergi , WDYT?

EdDev (approved):
Thanks!

We need to fix the local cache we use today.
We currently check the existence of a file, which does not include the version. This keeps old versions locally, avoiding their update.
We should have a follow up PR that changes the name of the d/l file, to include a version (and to compare against the expected file).

orelmisan:
> Thanks!
> 
> We need to fix the local cache we use today. We currently check the existence of a file, which does not include the version. This keeps old versions locally, avoiding their update. We should have a follow up PR that changes the name of the d/l file, to include a version (and to compare against the expected file).

Agreed.

ormergi:
> > We should consider pining and mirroring kind [node images](https://hub.docker.com/r/kindest/node/), which are hosted on hub.docker.com.
> 
> Agreed, the images should be pinned to assure we explicitly do the upgrade. Although, I think that if we pin the kind version, we in practice pin the image as well (implicitly).
> 
> @ormergi , WDYT?

That seems right, IMO the image should be explicitly set if one would like to use its own custom image or spin older version of k8s (e.g: 1.22 using the KIND release images).

ormergi (approved):
LGTM

PR #201, opened 2022-11-02
vmlatency, Dockerfile: Upgrade base image version

orelmisan:
In order to address security issues discovered in the base image, upgrade it to version `8.6-994`.


EdDev (approved):
Thanks.

I think we should solve it for the long run and stop investing in these upgrades.
The long run solution is to drop the ubi completely and use scratch.

PR #202, opened 2022-11-03
vmlatency: Randomize VMI names

orelmisan:
Currently, source and target VMI names are hard-codded. This prevents simultaneous execution of multiple checkup instances in the same namespace.

VMIs names will have a random string prefix:
`latency-check-source-<random string>`
`latency-check-target-<random string>`

Partially solves issue #164.

~~Depends on PR #205.~~

### DIFF
--- a/checkups/kubevirt-vm-latency/vendor/k8s.io/apimachinery/pkg/util/rand/rand.go
+++ b/checkups/kubevirt-vm-latency/vendor/k8s.io/apimachinery/pkg/util/rand/rand.go
@@ -1,0 +1,127 @@
+/*
+Copyright 2015 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package rand provides utilities related to randomization.
+package rand
+
+import (
+	"math/rand"
+	"sync"
+	"time"
+)
+
+var rng = struct {
+	sync.Mutex
+	rand *rand.Rand
+}{
+	rand: rand.New(rand.NewSource(time.Now().UnixNano())),
+}
+
+// Int returns a non-negative pseudo-random int.
+func Int() int {
+	rng.Lock()
+	defer rng.Unlock()
+	return rng.rand.Int()
+}
+
+// Intn generates an integer in range [0,max).
+// By design this should panic if input is invalid, <= 0.
+func Intn(max int) int {
+	rng.Lock()
+	defer rng.Unlock()
+	return rng.rand.Intn(max)
+}
+
+// IntnRange generates an integer in range [min,max).
+// By design this should panic if input is invalid, <= 0.
+func IntnRange(min, max int) int {
+	rng.Lock()
+	defer rng.Unlock()
+	return rng.rand.Intn(max-min) + min
+}
+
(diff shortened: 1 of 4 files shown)


RamLavi (dismissed):
/lgtm

checkups/kubevirt-vm-latency/vmlatency/internal/namegenerator/namegenerator.go
```diff
@@ -0,0 +1,40 @@
...
+
+func (ng NameGenerator) Name(prefix string) string {
+	const randomStringLen = 5
+
+	return fmt.Sprintf("%s-%s", prefix, k8srand.String(randomStringLen))
+}
+
+type Stub struct{}
+
+func (ngs Stub) Name(prefix string) string {
```
RamLavi, line 38: why do you put the stub here and not on each test module?

orelmisan: The stub is being used at `checkup_test.go` and `launcher_test.go`.
I'm open to suggestions if you think there is a better place to put it.

EdDev: - I do not think it is common to create stubs in production code, even if it means duplication.
- I would prefer to avoid this stub completely. The information of what VMI is created is saved in the results field and can be retrieved using `Results()`. Can you use that instead to determine the name of the VMI?

EdDev (changes requested):

checkups/kubevirt-vm-latency/vmlatency/internal/namegenerator/namegenerator.go
orelmisan: > * I do not think it is common to create stubs in production code, even if it means duplication.

I'll remove it from here.

> * I would prefer to avoid this stub completely. The information of what VMI is created is saved in the results field and can be retrieved using `Results()`. Can you use that instead to determine the name of the VMI?

This name stub is used to randomize the VMI name upon its creation.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/23bda5831a6624b8f87bd2b005a10ca6f7d6af86..8be2d31c540972459492e163422a5322faaa3775): Moved `nameGeneratorStub` to `checkup_test.go` and `launcher_test.go`.

checkups/kubevirt-vm-latency/vmlatency/internal/namegenerator/namegenerator.go
EdDev: I do not think we are interested in testing the randomization.
And the name of the VMI you can get from the results, you do not need a stub. What am I missing here?

orelmisan: > I do not think we are interested in testing the randomization.

We are not testing the randomization.

> And the name of the VMI you can get from the results, you do not need a stub. What am I missing here?

The results do not contain the VMIs names.

We use the VMIs names in order to check the pod anti affinity or node affinity.

EdDev: > The results do not contain the VMIs names.

Yes, you are right. I got confused with the checkup members.
I think you can still solve it without complicating the production code with an object (to allows stubbing in tests).
I can think now of two options:
- Add getters for the VMI/s from the checkup object.
- Take advantage of the known prefix, list the VMI/s, filter the one relevant and use it.

The second seems less intrusive and moves the work to the tests. I think that would be my favorite.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/8be2d31c540972459492e163422a5322faaa3775..693a1b02cbc8361704033b9c1d80503f1bece347): Removed `nameGenerator` dependency from checkup.

EdDev (commented):
How come `go.mod` has not changed?
You added `k8s.io/apimachinery/pkg/util/rand` but I see only things removed (not added).

EdDev (dismissed):
Thanks!

PR #206, opened 2022-11-08
Update Go version to 1.19

ormergi:
In order to upgrade k8s dependencies it is necessary to upgrade Go version to 1.19.
This PR updates kiagnose and kubevirt-vm-latnecy to use Go 1.19 version.

Updating Go version requires upgrading the linter as well, golangci-lint is upgraded to 1.50.1.


orelmisan (dismissed):
Thank you @ormergi.
@EdDev could you please have a look?

EdDev (commented):
I'm trying to understand why the root `go.sum` file has not changed while a lot of vendored files changed.

Also, have you verified that running `tidy` before the bump has not changed anything?

ormergi:
> Also, have you verified that running `tidy` before the bump has not changed anything?

No, because I was under the impression that everything is aligned.
I will check.

ormergi:
> I'm trying to understand why the root `go.sum` file has not changed while a lot of vendored files changed.

@EdDev note that root (kiagnose) vendor dir is not tracked, I added it because I thought we do.

ormergi:
[Changes](https://github.com/kiagnose/kiagnose/compare/2c025409227542783db86ad4428c461e79d2d5bb..a406730e8b03b571e714ec6b29c72571e715932c): removed root vendor dir as it was before, it's not clear for me if it's necessary or not as Kiagnose does not produce a binary.

ormergi:
@EdDev please take another look.

orelmisan (approved):

EdDev:
> it's not clear for me if it's necessary or not as Kiagnose does not produce a binary.

Because it is a library, there is no need to vendor.

PR #208, opened 2022-11-08
vmlatency: Randomize VMs MAC and IP addresses

ormergi:
Currently its not possible to run multiple instances of the checkup simultaneously, one of the reasons is due to hard-coded IP and MAC addresses.
VMs from one checkup instance conflict with VMs from  another checkup instance
and may cause false checkup results.

Randomize the MAC and IP addresses in order prevent such conflicts.

Fixes https://github.com/kiagnose/kiagnose/issues/164

### DIFF
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
@@ -77,11 +77,6 @@ func (c *checkup) Setup(ctx context.Context) error {
 		errMessagePrefix = "setup"
 
 		defaultSetupTimeout = time.Minute * 10
-
-		sourceVmiMac  = "02:00:00:01:00:01"
-		sourceVmiCidr = "192.168.100.10/24"
-		targetVmiMac  = "02:00:00:02:00:02"
-		targetVmiCidr = "192.168.100.20/24"
 	)
 
 	netAttachDef, err := c.client.GetNetworkAttachmentDefinition(
@@ -94,8 +89,8 @@ func (c *checkup) Setup(ctx context.Context) error {
 	sourceVMIName := randomizeName(SourceVMINamePrefix)
 	targetVMIName := randomizeName(TargetVMINamePrefix)
 
-	sourceVmi := newLatencyCheckVmi(c.uid, sourceVMIName, c.params.SourceNodeName, netAttachDef, sourceVmiMac, sourceVmiCidr)
-	targetVmi := newLatencyCheckVmi(c.uid, targetVMIName, c.params.TargetNodeName, netAttachDef, targetVmiMac, targetVmiCidr)
+	sourceVmi := newLatencyCheckVmi(c.uid, sourceVMIName, c.params.SourceNodeName, netAttachDef)
+	targetVmi := newLatencyCheckVmi(c.uid, targetVMIName, c.params.TargetNodeName, netAttachDef)
 
 	if err = vmi.Start(c.client, c.namespace, sourceVmi); err != nil {
 		return fmt.Errorf("%s: %v", errMessagePrefix, err)
@@ -120,11 +115,8 @@ func (c *checkup) Setup(ctx context.Context) error {
 }
 
 func newLatencyCheckVmi(
-	uid,
-	name,
-	nodeName string,
-	netAttachDef *netattdefv1.NetworkAttachmentDefinition,
-	macAddress, cidr string) *kvcorev1.VirtualMachineInstance {
+	uid, name, nodeName string,
+	netAttachDef *netattdefv1.NetworkAttachmentDefinition) *kvcorev1.VirtualMachineInstance {
 	const networkName = "net0"
 
 	vmLabel := vmi.Label{Key: LabelLatencyCheckUID, Value: uid}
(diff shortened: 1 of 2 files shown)


checkups/kubevirt-vm-latency/vmlatency/internal/vmi/spec.go
```diff
@@ -207,6 +210,26 @@ func WithMacAddress(macAddress string) interfaceOption {
...
 
+// RandomMACAddress return random global unicast MAC address
+func RandomMACAddress() string {
+	prefix := []byte{0x00, 0x16, 0x3e}
+	const randomBytesCount = 3
+	suffix := make([]byte, randomBytesCount)
+	_, _ = rand.Read(suffix)
+
+	return net.HardwareAddr(append(prefix, suffix...)).String()
+}
```
orelmisan, line 221: IMO we should use [unicast](https://en.wikipedia.org/wiki/MAC_address#Unicast_vs._multicast_(I/G_bit)) and [locally administered addresses](https://en.wikipedia.org/wiki/MAC_address#Universal_vs._local_(U/L_bit)), instead of using universally administered addresses with a well known [OUI](https://hwaddress.com/company/xensource-inc/).
This means that we could:
1. Create only one six-byte slice.
2. Fill it with six random bytes.
3. Ensure that the left-most byte is x2, x6, xA or xE.
4. Convert the byte slice to string.

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/spec.go
```diff
@@ -207,6 +210,26 @@ func WithMacAddress(macAddress string) interfaceOption {
...
+
+// RandomIPAddress return random private IPv4 CIDR notation address from range 10.0.0.0/8
+func RandomIPAddress() string {
+	prefix := []byte{10}
+	const randomOctetsCount = 3
+	suffix := make([]byte, randomOctetsCount)
+	_, _ = rand.Read(suffix)
+	cidr := net.IPNet{IP: append(prefix, suffix...), Mask: []byte{255, 0, 0, 0}}
```
orelmisan, line 229: Please consider this small optimization suggestion:
1. Create a single four byte slice.
2. Set the left-most byte to `10`.
3. Fill the other three bytes with random numbers.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -135,6 +127,7 @@ func newLatencyCheckVmi(
 		affinity = &k8scorev1.Affinity{PodAntiAffinity: vmi.NewPodAntiAffinity(vmLabel)}
 	}
 
+	macAddress := vmi.RandomMACAddress()
```
orelmisan, line 130: Please consider inlining this method call, so it will be consistent with the call to `vmi.RandomIPAddress()`.

orelmisan (commented):
Thank you for the PR @ormergi :)

Could you please reference issue #164 in the commit message and PR description?
Have you tried to execute locally two simultaneous VM latency checkups?

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/spec.go
orelmisan: This will emphasize to an external observer that the MAC addresses are generated.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
ormergi: It cannot be inlined, we need to set the interface with a custom MAC (line 137) and specify it in cloud-init config as a matching address (144)

orelmisan: Thank you for the clarification.

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/spec.go
ormergi: Done

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/spec.go
ormergi: I changed the flow to be similar to how MACs are generated.

ormergi:
@orelmisan please take another look.

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/spec.go
```diff
@@ -207,6 +210,25 @@ func WithMacAddress(macAddress string) interfaceOption {
 	}
 }
 
+// RandomMACAddress return random local unicast MAC address
+func RandomMACAddress() string {
+	const octetsCount = 6
+	address := make([]byte, octetsCount)
+	_, _ = rand.Read(address)
+	address[0] = 0x02
```
orelmisan, line 218: In this byte we only care that the least significant bits are 10 (locally managed address and unicast).
The rest of the bits could be random.

```suggestion
	address[0] = (address[0] | 2) & 0xFE
```

ormergi: I think that setting 0x02 directly is easier to understand than what you suggest, there are many ways to solve this problem lets keep it simple as possible.

ormergi:
I verified and with these PR changes, it's possible (and works!) to run multiple instances of the checkup simultaneously.

orelmisan:
> Could you please reference issue #164 in the commit message and PR description?

ormergi:
> > Could you please reference issue #164 in the commit message and PR description?

Done

orelmisan:
> > > Could you please reference issue #164 in the commit message and PR description?
> 
> Done

Could you please update the commit message as well?

ormergi:
> 

Done

ormergi:
> Could you please update the commit message as well?

@orelmisan Done

orelmisan (approved):
Thank you @ormergi!

PR #211, opened 2022-11-13
vmlatency, checkup: Add OwnerReference to VMIs

orelmisan:
Currently, if a problem occurs during the teardown process, one or more VMIs could stay in the namespace as leftovers.

Add OwnerReference to source and target VMIs, in order to use the automatic K8s garbage collector, so they will be
garbage-collected when the checkup Pod is deleted.

The ownerReference is only created if the Pod's name and UID are known: The name should always be present, the UID is dependent on the user's configuration of downward API.

### DIFF
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
@@ -89,8 +89,8 @@ func (c *checkup) Setup(ctx context.Context) error {
 	sourceVMIName := randomizeName(SourceVMINamePrefix)
 	targetVMIName := randomizeName(TargetVMINamePrefix)
 
-	sourceVmi := newLatencyCheckVmi(c.uid, sourceVMIName, c.params.SourceNodeName, netAttachDef)
-	targetVmi := newLatencyCheckVmi(c.uid, targetVMIName, c.params.TargetNodeName, netAttachDef)
+	sourceVmi := newLatencyCheckVmi(c.uid, sourceVMIName, c.params.SourceNodeName, c.params.PodName, c.params.PodUID, netAttachDef)
+	targetVmi := newLatencyCheckVmi(c.uid, targetVMIName, c.params.TargetNodeName, c.params.PodName, c.params.PodUID, netAttachDef)
 
 	if err = vmi.Start(c.client, c.namespace, sourceVmi); err != nil {
 		return fmt.Errorf("%s: %v", errMessagePrefix, err)
@@ -115,7 +115,7 @@ func (c *checkup) Setup(ctx context.Context) error {
 }
 
 func newLatencyCheckVmi(
-	uid, name, nodeName string,
+	uid, name, nodeName, ownerName, ownerUID string,
 	netAttachDef *netattdefv1.NetworkAttachmentDefinition) *kvcorev1.VirtualMachineInstance {
 	const networkName = "net0"
 
@@ -129,6 +129,7 @@ func newLatencyCheckVmi(
 
 	macAddress := vmi.RandomMACAddress()
 	return vmi.NewAlpine(name,
+		vmi.WithOwnerReference(ownerName, ownerUID),
 		vmi.WithLabels(vmLabel),
 		vmi.WithAffinity(affinity),
 		vmi.WithMultusNetwork(networkName, netAttachDef.Namespace+"/"+netAttachDef.Name),
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
@@ -38,6 +38,8 @@ const (
 )
 
 type Config struct {
+	PodName                              string
+	PodUID                               string
 	NetworkAttachmentDefinitionName      string
 	NetworkAttachmentDefinitionNamespace string
 	TargetNodeName                       string
@@ -64,6 +66,8 @@ func New(baseConfig kconfig.Config) (Config, error) {
 	}
 
 	newConfig := Config{
+		PodName:                              baseConfig.PodName,
+		PodUID:                               baseConfig.PodUID,
 		NetworkAttachmentDefinitionName:      baseConfig.Params[NetworkNameParamName],
 		NetworkAttachmentDefinitionNamespace: baseConfig.Params[NetworkNamespaceParamName],
 		TargetNodeName:                       baseConfig.Params[TargetNodeNameParamName],
--- a/checkups/kubevirt-vm-latency/README.md
+++ b/checkups/kubevirt-vm-latency/README.md
@@ -155,6 +155,10 @@ spec:
               value: <target-namespace>
             - name: CONFIGMAP_NAME
               value: kubevirt-vm-latency-checkup-config
+            - name: POD_UID
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.uid
 EOF
 ```
 
--- a/checkups/kubevirt-vm-latency/automation/e2e.sh
+++ b/checkups/kubevirt-vm-latency/automation/e2e.sh
@@ -271,6 +271,10 @@ spec:
               value: ${TARGET_NAMESPACE}
             - name: CONFIGMAP_NAME
               value: ${VM_LATENCY_CONFIGMAP}
+            - name: POD_UID
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.uid
 EOF
 
     ${KUBECTL} wait --for=condition=complete --timeout=10m job.batch/${CHECKUP_JOB} -n ${TARGET_NAMESPACE}
--- a/checkups/kubevirt-vm-latency/vendor/github.com/kiagnose/kiagnose/kiagnose/config/config.go
+++ b/checkups/kubevirt-vm-latency/vendor/github.com/kiagnose/kiagnose/kiagnose/config/config.go
@@ -37,6 +37,7 @@ var (
 type Config struct {
 	ConfigMapNamespace string
 	ConfigMapName      string
+	PodName            string
 	UID                string
 	Timeout            time.Duration
 	Params             map[string]string
@@ -63,6 +64,7 @@ func Read(client kubernetes.Interface, rawEnv map[string]string) (Config, error)
 	return Config{
 		ConfigMapNamespace: env.ConfigMapNamespace,
 		ConfigMapName:      env.ConfigMapName,
+		PodName:            env.PodName,
 		UID:                cmSettings.UID,
 		Timeout:            cmSettings.Timeout,
 		Params:             cmSettings.Params,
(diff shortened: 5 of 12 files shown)


orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/a87e120685c5237f5c445616382f85121bb023d1..41cee3b1cf6c5fe71c168e79b75d754961609410): Updated commit message and PR description

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -144,6 +146,7 @@ func newLatencyCheckVmi(
 				vmi.WithMatchingMAC(macAddress),
 			),
 		),
+		vmi.WithOwnerReference(uid, configMapName),
```
ormergi, line 149: nit: Please more it to be the first option of the builder (line 134).

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -46,23 +46,25 @@ type checker interface {
...
+	configMapName string
+	params        config.Config
+	results       status.Results
+	sourceVM      *kvcorev1.VirtualMachineInstance
+	targetVM      *kvcorev1.VirtualMachineInstance
+	checker       checker
 }
 
-func New(c vmi.KubevirtVmisClient, uid, namespace string, params config.Config, checker checker) *checkup {
+func New(c vmi.KubevirtVmisClient, uid, namespace, configMapName string, params config.Config, checker checker) *checkup {
```
ormergi, line 60: Please consider renaming `configMapName` field to something like `owner`/`ownerName`/`ownerObject`.
Having it like that is a bit confusing and doesn't help me understand that the owner is the checkup config-map (it can be understood from mainflow.go line 55).

ormergi (changes requested):
@orelmisan how did you verify that it actually works? (I am asking because we don't have e2e tests yet).

orelmisan:
> @orelmisan how did you verify that it actually works? (I am asking because we don't have e2e tests yet).

I've tested it locally:
1. Disabled the teardown.
2. Listed the VMIs to see they are still there.
3. Removed the user-supplied ConfigMap,
4. Listed the VMIs to see they are removed.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
orelmisan: The problem is that the owner'st kind is hard-codded to "ConfigMap".
So it will look weird IMO.

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
orelmisan: Done.

orelmisan:
Change: Get the ownerReference as an input to checkup.New(), so checkup will not care what object is in the owner reference.
@ormergi could you please take another look?

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
orelmisan: This parameter was removed.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/c2a5782482675395b93ddf209a0c95e0901719ab..7480c40cd29cf9d66ab29e3eda9bc1f4a56ede7e): Removed unused explicit field initialization.

ormergi (dismissed):
Thanks for the changes, nice work!

With this change, we won't need to worry about leftovers when we will introduce the "retain" flag for debugging as the VMs will be deleted once the config-map is deleted.

RamLavi (approved):
/lgtm

ormergi (changes requested):
I prefer how it was before, the UID should be explicitly defined in the checkup object.
Taking it from another object is actually moving the responsibility of the UID to the checkup.

Also, IMO having the code more compact doesn't worth it in this case.
The amount of parameters you need to pass to functions (`checkup.New` and `newLatencyCheckVmi`) is reduced but it doesn't improve readability.

orelmisan:
> I prefer how it was before, the UID should be explicitly defined in the checkup object.
Taking it from another object is actually moving the responsibility of the UID to the checkup.

> Also, IMO having the code more compact doesn't worth it in this case.
The amount of parameters you need to pass to functions (checkup.New and newLatencyCheckVmi) is reduced but it doesn't improve readability.

The UID is used in two places:
1. OwnerReference.
2. Value of the label that is used for podAntiAffinity.

Why is it important to explicitly get it?
The value is the same.

Several interfaces where cleaned, why is it a bad thing in your opinion?

orelmisan:
Rebase

checkups/kubevirt-vm-latency/vmlatency/mainflow.go
```diff
@@ -51,8 +54,15 @@ func Run(env map[string]string, namespace string) error {
 		return err
 	}
 
+	ownerReference := metav1.OwnerReference{
```
EdDev, line 57: I would like to move the mainflow skeleton to the kiagnose package so it can serve other checkups.
By adding this ownership here, it will cause the skeleton to be too specific and not generic. Using the configmap as the owner is not always correct, I can think of other checkup authors that will prefer to do something completely different.

As this is specific to a checkup, I think you should move this to the checkup itself to do.
If you need to place additional info for the checkup to compose this configuration, add it to the config struct, which is checkup specific.

EdDev (changes requested):

EdDev:
> > I prefer how it was before, the UID should be explicitly defined in the checkup object.
> > Taking it from another object is actually moving the responsibility of the UID to the checkup.
> 
> > Also, IMO having the code more compact doesn't worth it in this case.
> > The amount of parameters you need to pass to functions (checkup.New and newLatencyCheckVmi) is reduced but it doesn't improve readability.
> 
> The UID is used in two places:
> 
>     1. OwnerReference.
> 
>     2. Value of the label that is used for podAntiAffinity.
> 
> 
> Why is it important to explicitly get it? The value is the same.
> 
> Several interfaces were cleaned, why is it a bad thing in your opinion?

The cleanup came on the expense of an nontrivial dependency.
The skeleton has explicitly used the UID of the configmap, but it is a generic decision.
The suggested change says that the UID used is up to the checkup package (not the mainflow skeleton) itself to decide and the decision is to take it from the owner object.

Considering that other checkups will be written, we can either provide a default UID (of the configmap) through a skeleton lib or not provide one at all. If we provide it, the checkup writer could still override it.
If we will not provide it, it means the responsibility is completely on the checkup writer and we cannot assist her with this thing.

I prefer we do provide the UID from above explicitly, and it is up to the specific checkup to use it or use something else.

And as I commented before, the owner object as an input is the real problem IMO.

orelmisan:
In an offline meeting with @EdDev we've concluded that its more logical to use the checkup's Job underlying Pod as the owner, because it is most probably be removed, unlike the user-supplied ConfigMap which could linger.

One option discussed to get the Pod's name and UID is using downward API as environment variables.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/db9c910cc4070f6aee0474b906a0d68128b8baaf..cdf8aa3d1eeae5150015f7d0561e44622f8a190b): Create the ownerReference inside the checkup, based on the Pod's UID and name.

[Change 2](https://github.com/kiagnose/kiagnose/compare/cdf8aa3d1eeae5150015f7d0561e44622f8a190b..4a5303a56274b58575f130d23d81f102fad5db6f): Added a test to check that ownerReference is not created when Pod's UID is unknown.

EdDev (commented):
Second commit:
> in order to get the Pod's name

I guess you meant ID not name, right?

orelmisan:
> Second commit:
> 
> > in order to get the Pod's name
> 
> I guess you meant ID not name, right?

Yes, thank you.

orelmisan:
Changed the [second](https://github.com/kiagnose/kiagnose/pull/211/commits/3f891dddc8988e368c0d8139e650edeb0ffbc003) commit's message in order to answer @EdDev's [comment](https://github.com/kiagnose/kiagnose/pull/211#pullrequestreview-1189901802).

checkups/kubevirt-vm-latency/vmlatency/mainflow.go
```diff
@@ -51,8 +52,11 @@ func Run(env map[string]string, namespace string) error {
 		return err
 	}
 
+	ownerUID := environment.PodUID(env)
+	ownerName := environment.PodName(env)
```
EdDev, line 56: I would have prefered this to come in through the `cfg`, which is specific to the checkup.

Can we add it to the baseConfig? Then it will pass to the checkup config which may or may not make use of it.
WDYT?

orelmisan: We could add it to the base config.
It would require some changes to the `config.go` file in kiagnose and to `config.New()` in the checkup.

orelmisan: I've published PR #224 to complete the changes required to answer this comment.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/55b61f7859324ddd7fd1ed140cd11cadd9c146c3..915c45878630e1dca5e8f2b431066932b5a0b556): Inserted the Pod's name and UID to the checkup's config.
This is an answer to @EdDev's [comment](https://github.com/kiagnose/kiagnose/pull/211#discussion_r1029256028).

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/b49b0ae055e767435fdfa29c6653615c234e6078..a4969656ecf9576a8a62ed871f5c41c81b817d00): Aligned the order of the Pod name and UID to be uniform.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/a4969656ecf9576a8a62ed871f5c41c81b817d00..ef5ed4f05d719aaff47cda075d50bf7b78155bb1): Fixed a bug in `TestCheckupSetupShouldNotCreateOwnerReferenceWhenPodUIDIsEmpty()`.

checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
```diff
@@ -38,6 +38,8 @@ const (
 )
 
 type Config struct {
+	PodName                              string
+	PodUID                               string
```
EdDev, line 42: nit: I think the conversion to `types.UID` should have been already been done here.

Not that critical though.

EdDev (approved):
Very nice, thank you!

checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
orelmisan: I'll create a follow-up PR.
Thanks.

PR #212, opened 2022-11-13
Move ConfigMap access Role to readme

ormergi:
Currently, there is only one manifest, the checkup config-map access role,
which is used as a reference but it's out of context.

Add checkup config-map access role manifest to readme instead of a
standalone file.


README.md
```diff
@@ -101,6 +101,21 @@ spec:
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: kiagnose-configmap-access
+rules:
+  - apiGroups: [ "" ]
+    resources: [ "configmaps" ]
+    verbs: ["get", "update"]
+```
+
```
orelmisan, line 118: WDYT about moving it under `Installation Steps`?
I think it could be nice to also show the example ServiceAccount and RoleBinding.

orelmisan (commented):
Thank you for the PR @ormergi 
Please see the inline comment.

README.md
ormergi: Done

README.md
```diff
@@ -29,7 +29,35 @@ Kiagnose checkups are expecting the namespace admin to supply the ServiceAccount
...
+    metadata:
+      name: kiagnose-configmap-access
+    subjects:
+    - kind: ServiceAccount
+      name: example-sa
+    roleRef:
+      kind: Role
+      apiGroup: rbac.authorization.k8s.io
+      name: kiagnose-configmap-access
+   ```
```
orelmisan, line 60: Could you please align the text to the left?

README.md
orelmisan: Thanks

README.md
ormergi: Its indented to be aligned with section 3.

orelmisan (dismissed):

EdDev:
I am not yet convinced this is a better direction.
Accessing the CM is the single manifest we consider common between all checkups and it makes sense to me to use it as a tool/lib.

It is harder to test a readme, but a manifest file can be used, tested and consumed by others easily.

I also find the details added to the root readme an overkill. The instructions there are more of an overview over exact steps/details. I would prefer there an explanation to what the manifest contains and used for.

RamLavi (approved):
/lgtm

ormergi:
Having it as a standalone file requires updating it manually as there is not generated automatically (no release process),
and the user still needs to create the service-account and the role-binding it doesn't reflect the "full picture".
Relying on that file in e2e tests requires each script to set a project path and derive the manifests dir path from it, 
it makes the scripts bound to the project files tree, I don't like this pattern and prefer to avoid it when possible.

What I had in mind is to combine that manifest in the readme and to improve the readme because it is a little bit vague about what is necessary for working with Kiagnose.
In general, I think that a readme should include snippets one could copy-paste and actually get started quickly and have something that works easily.

ormergi:
rebased

EdDev:
> Having it as a standalone file requires updating it manually as there is not generated automatically (no release process)

What do we need to update manually and why?

> , and the user still needs to create the service-account and the role-binding it doesn't reflect the "full picture". Relying on that file in e2e tests requires each script to set a project path and derive the manifests dir path from it, it makes the scripts bound to the project files tree, I don't like this pattern and prefer to avoid it when possible.

You can argue the same for not having any kiagnose libraries dependencies at all.
And no one enforces their usage, it is there for convenience.
In this repo, it makes sense to me to re-use this manifest and by that hint this is something common.
And we can maintain it by just using it in the tests, making sure it is correct.

> What I had in mind is to combine that manifest in the readme and to improve the readme because it is a little bit vague about what is necessary for working with Kiagnose. In general, I think that a readme should include snippets one could copy-paste and actually get started quickly and have something that works easily.

You could solve this by generating the manifests for the user based on some minimal input.

PR #213, opened 2022-11-13
github workflows: Add go-mod-tidy action

ormergi:
Add [go-tidy-check](https://github.com/marketplace/actions/go-tidy-check?version=v1.0.2) to github worklow.

It will improve the project integrity by blocking PRs that cause not-committed go module changes.


ormergi:
It seems that vmlatency checkup modules are not tidy https://github.com/kiagnose/kiagnose/actions/runs/3455511751/jobs/5767585557, I've posted a PR to fix it #214.

ormergi:
rebased

ormergi:
rebased

orelmisan (commented):

The `katexochen/go-tidy-check` github action is a quite new and doesn't have a large user base.
I suggest to implement its functionality in the two `automation/make.sh` scripts using `go mod tidy` and `git diff` for example.

@EdDev WDYT?

ormergi:
> The `katexochen/go-tidy-check` github action is a quite new and doesn't have a large user base. I suggest to implement its functionality in the two `automation/make.sh` scripts using `go mod tidy` and `git diff` for example.
> 

@orelmisan how can you tell how many users a github action has?
The fact that it's not popular doesn't mean its not good.

orelmisan:
> > The `katexochen/go-tidy-check` github action is a quite new and doesn't have a large user base. I suggest to implement its functionality in the two `automation/make.sh` scripts using `go mod tidy` and `git diff` for example.
> 
> @orelmisan how can you tell how many users a github action has? The fact that it's not popular doesn't mean its not good.

I haven't said it is not good.
I'm afraid that we use it and it won't be maintained for a long time.

ormergi:
> I haven't said it is not good. I'm afraid that we use it and it won't be maintained for a long time.

Well, if it will breaks we can always use somthing else.
When I searched for go-mod-tidy action, this one seemed the most reliable it also tested.

EdDev (approved):
I am not worried about the new action item, it seems to be writen in Go so if there are problems we could contribute and fix it.

What I am missing here is checking other stuff, like vendoring results.
(that we pushed the vendored content)
Maybe the job should be named `go-mod` so we can place there other module checks. But I guess we can do that later as well.

PR #217, opened 2022-11-15
vmlatency: Define user at Dockerfile

orelmisan:
Currently, the vm latency Dockerfile does not specify the user running the kubevirt-vm-latency, thus it is implicitly running as root.

This required the checkup Job to specify the container's user.

Add a `USER` stanza to the vm latency checkup's Dockerfile.
Remove `runAsUser` from the container's securityContext.

Install `shadow-utils` in the image in order to use the `adduser` command, and remove it after creating the user.

### DIFF
--- a/checkups/kubevirt-vm-latency/Dockerfile
+++ b/checkups/kubevirt-vm-latency/Dockerfile
@@ -1,5 +1,12 @@
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6-994
 
+RUN microdnf install shadow-utils && \
+    adduser --system --no-create-home -u 1001 vm-latency && \
+    microdnf remove shadow-utils && \
+    microdnf clean all
+
 COPY ./bin/kubevirt-vm-latency /usr/bin
+
+USER 1001
 
 CMD kubevirt-vm-latency
--- a/checkups/kubevirt-vm-latency/README.md
+++ b/checkups/kubevirt-vm-latency/README.md
@@ -144,7 +144,6 @@ spec:
         - name: vm-latency-checkup
           image: quay.io/kiagnose/kubevirt-vm-latency:main
           securityContext:
-            runAsUser: 1000
             allowPrivilegeEscalation: false
             capabilities:
               drop: ["ALL"]
--- a/checkups/kubevirt-vm-latency/automation/e2e.sh
+++ b/checkups/kubevirt-vm-latency/automation/e2e.sh
@@ -260,7 +260,6 @@ spec:
         - name: vm-latency-checkup
           image: ${CHECKUP_IMAGE}
           securityContext:
-            runAsUser: 1000
             allowPrivilegeEscalation: false
             capabilities:
               drop: ["ALL"]


orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/cacae3c35da5b8428cb9225034cbbfccde63f5bd..515c974b47d196ec99b3d35368dbc7f2fb560c40): Commit message.

checkups/kubevirt-vm-latency/Dockerfile
```diff
@@ -2,4 +2,6 @@ FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6-994
 
 COPY ./bin/kubevirt-vm-latency /usr/bin
 
+USER nobody
```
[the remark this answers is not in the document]
orelmisan: It is one of the users that exists in our base image:
```bash
$ podman run --rm -it registry.access.redhat.com/ubi8/ubi-minimal cat /etc/passwd
root:x:0:0:root:/root:/bin/bash
bin:x:1:1:bin:/bin:/sbin/nologin
daemon:x:2:2:daemon:/sbin:/sbin/nologin
adm:x:3:4:adm:/var/adm:/sbin/nologin
lp:x:4:7:lp:/var/spool/lpd:/sbin/nologin
sync:x:5:0:sync:/sbin:/bin/sync
shutdown:x:6:0:shutdown:/sbin:/sbin/shutdown
halt:x:7:0:halt:/sbin:/sbin/halt
mail:x:8:12:mail:/var/spool/mail:/sbin/nologin
operator:x:11:0:operator:/root:/sbin/nologin
games:x:12:100:games:/usr/games:/sbin/nologin
ftp:x:14:50:FTP User:/var/ftp:/sbin/nologin
nobody:x:65534:65534:Kernel Overflow User:/:/sbin/nologin
```

We could also create a dedicated user and use it instead.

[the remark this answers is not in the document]
ormergi: I looked it up, it seems that `nobody` is standard user in Linux systems with minimal permissions, here is a nice explanation https://unix.stackexchange.com/a/186581
See also https://refspecs.linuxbase.org/LSB_3.0.0/LSB-PDA/LSB-PDA/usernames.html

I think creating a dedicated user is better.

checkups/kubevirt-vm-latency/README.md
```diff
@@ -144,11 +144,9 @@ spec:
         - name: vm-latency-checkup
           image: quay.io/kiagnose/kubevirt-vm-latency:main
           securityContext:
-            runAsUser: 1000
             allowPrivilegeEscalation: false
             capabilities:
               drop: ["ALL"]
-            runAsNonRoot: true
```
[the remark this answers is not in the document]
orelmisan: It was not working when used a mnemonic name.
I've used a number user id and returned it.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/515c974b47d196ec99b3d35368dbc7f2fb560c40..dbacfc00d9aacaab22d320237aa6ccf937ceef9e): Use a dedicated user for the checkup.

checkups/kubevirt-vm-latency/Dockerfile
```diff
@@ -1,5 +1,9 @@
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6-994
 
+RUN microdnf install shadow-utils
+RUN  adduser --system --no-create-home -u 1001 vm-latency
+USER 1001
```
ormergi, line 5: I would remove the uid and let it pick the number, we can pass the user name to USER right?

orelmisan: We can pass the user name to USER, but then we will get an error from K8s for using `runAsNonRoot` with a mnemonic user name and not a number

orelmisan:
```
container has runAsNonRoot and image has non-numeric user (vm-latency), cannot verify user is non-root (pod: "kubevirt-vm-latency-checkup-r5jvb_target-ns(426d9f73-fbe9-464d-a31c-aa301b5710fa)", container: vm-latency-checkup)
```

orelmisan: I guess they want to verify the the user id is greater than zero, to disable another mnemonic name for the root user.

checkups/kubevirt-vm-latency/Dockerfile
```diff
@@ -1,5 +1,9 @@
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6-994
 
+RUN microdnf install shadow-utils
```
ormergi, line 3: Please add `&& microdnf clear all` in order to remove the cached installation fails.
Also format the file (newline after each command).

orelmisan: Done.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/dbacfc00d9aacaab22d320237aa6ccf937ceef9e..221e4a4eb79e39b2b23e7a718b30d5f01c466f9a): Answered a [comment](https://github.com/kiagnose/kiagnose/pull/217#discussion_r1022689946).

ormergi (approved):
Thanks for the changes, LGTM.

EdDev (commented):
I do not understand why this is the correct action to take.
Specifying with which user to run in the job seems more secure to me.

Depending for the image to have a user is less secure, because it may just miss that part in general.
Could you please explain why this is consider more secure?
And also explain what does not work? (alternatively just ref an issue/bug)

I also do not like that the image adds a binary in it, I would hope to see this done as part of the build, and not leave it in the container.

orelmisan:
> I do not understand why this is the correct action to take. Specifying with which user to run in the job seems more secure to me.
Depending for the image to have a user is less secure, because it may just miss that part in general. Could you please explain why this is consider more secure? And also explain what does not work? (alternatively just ref an issue/bug)

Please see [this](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user) link to the dockerfile best practices regarding the USER stanza.

K8s API reference - [securityContext](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#securitycontext-v1-core):
`runAsUser`: The UID to run the entrypoint of the container process. Defaults to user specified in image metadata if unspecified.

`runAsNonRoot`: Indicates that the container must run as a non-root user. If true, the Kubelet will validate the image at runtime to ensure that it does not run as UID 0 (root) and fail to start the container if it does. If unset or false, no such validation will be performed.

In summary: we specify the user in the Dockerfile, so we are not dependent on the user specifying it in the checkup Job manifest.

> I also do not like that the image adds a binary in it, I would hope to see this done as part of the build, and not leave it in the container.

The image adds the `shadow-utils` package, because without it, there is no `adduser` command.
This is a tiny package.

EdDev:
> `runAsNonRoot`: Indicates that the container must run as a non-root user. If true, the Kubelet will validate the image at runtime to ensure that it does not run as UID 0 (root) and fail to start the container if it does. If unset or false, no such validation will be performed.

If we specify this, then I guess it protects against wrong usage and improper image creation.

Thanks for the explanation.

orelmisan:
> > The image adds the `shadow-utils` package, because without it, there is no `adduser` command.
> > This is a tiny package.
> 
> Why can't you do this from outside or perform a two step build?

We need to add a user to the image we are going to run.
This package is installed only because the ubi image is tiny and very minimal.

orelmisan:
[Change](https://github.com/kiagnose/kiagnose/compare/221e4a4eb79e39b2b23e7a718b30d5f01c466f9a..9b41408f9c9f051889c3f9e2d4d844178bdbb292): Remove the `shadow-utils` package after creating the user.

ormergi (approved):
LGTM

PR #219, opened 2022-11-17
vmlatency, config: Change environment to params

orelmisan:
Since PR #189, the checkup receives its parameters from the user-supplied ConfigMap instead from environment variables.

Change the terminology of environment variables to params in order to reflect the new status.

Note: This changes error messages seen by the user.

Signed-off-by: Orel Misan <omisan@redhat.com>

PR #189, opened 2022-10-26
kiagnose, checkup: Embed Kiagnose in the vm-latency checkup

EdDev:
The execution of a checkup, included two steps:
- Executing the Kiagnose Job (by the user).
- Executing the Checkup Job (by the Kiagnose application).

The vm-latency checkup is changed now to use Kiagnose as an external
library and to execute in a one step.
The logic that existed in the Kiagnose application is now imported
and used directly in the checkup.

With this change, the user is expected to execute the checkup Job
directly.
The checkup is now directly accessing the user-faced ConfigMap and the
usage of the internal ConfigMap is dropped.

The following main changes included in this PR:
- Expose kiagnose module packages in order to allow external checkups to use them.
- Modify the kubevirt-vm-latency checkup to use the kiagnose module packaged and operate as one application.
- Update the automation and CI scripts/configuration to successfully test the modified checkup.
- Update the documentation to reflect the new operation flow.

> **_NOTE_**: The current documented diagram is removed for now, until an updated one is introduced.

This change implements the design proposal described in #183 .

PR #220, opened 2022-11-17
config: Remove conversion to EnvVars and back

orelmisan:
Since PR https://github.com/kiagnose/kiagnose/pull/189, the checkup receives its parameters
from the user-supplied ConfigMap instead from environment variables.

Currently, the read params are converted to type EnvVar, and then back to map[string]string.

Remove these redundant conversions.
Remove some dependencies as a side effect.

Note: This change breaks the current interface provided by the kiagnose library.

~~Depends on PR #219.~~

### DIFF
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
@@ -27,12 +27,12 @@ import (
 )
 
 const (
-	NetworkNamespaceParamName              = "NETWORK_ATTACHMENT_DEFINITION_NAMESPACE"
-	NetworkNameParamName                   = "NETWORK_ATTACHMENT_DEFINITION_NAME"
-	SampleDurationSecondsParamName         = "SAMPLE_DURATION_SECONDS"
-	SourceNodeNameParamName                = "SOURCE_NODE"
-	TargetNodeNameParamName                = "TARGET_NODE"
-	DesiredMaxLatencyMillisecondsParamName = "MAX_DESIRED_LATENCY_MILLISECONDS"
+	NetworkNamespaceParamName              = "network_attachment_definition_namespace"
+	NetworkNameParamName                   = "network_attachment_definition_name"
+	SampleDurationSecondsParamName         = "sample_duration_seconds"
+	SourceNodeNameParamName                = "source_node"
+	TargetNodeNameParamName                = "target_node"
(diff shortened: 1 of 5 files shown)


checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
```diff
@@ -27,12 +27,12 @@ import (
...
-	SampleDurationSecondsParamName         = "SAMPLE_DURATION_SECONDS"
-	SourceNodeNameParamName                = "SOURCE_NODE"
-	TargetNodeNameParamName                = "TARGET_NODE"
-	DesiredMaxLatencyMillisecondsParamName = "MAX_DESIRED_LATENCY_MILLISECONDS"
+	NetworkNamespaceParamName              = "network_attachment_definition_namespace"
+	NetworkNameParamName                   = "network_attachment_definition_name"
+	SampleDurationSecondsParamName         = "sample_duration_seconds"
+	SourceNodeNameParamName                = "source_node"
+	TargetNodeNameParamName                = "target_node"
+	DesiredMaxLatencyMillisecondsParamName = "max_desired_latency_milliseconds"
```
ormergi, line 35: The status fields and params fields have different formats, the params are snake case and the status fields are camel case.
I think this is an opportunity to align them to be the same format.
I am favor of the camel case, its pretty common in k8s eco-system.

ormergi:
> Note: This change breaks the current interface provided by the kiagnose library.

How does it breaks?

checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
orelmisan: The change suggested by this PR, does not break the user-facing API.

In principle I agree with you, but IMO it should come in a time that is acceptable by the project's stakeholders (and in a dedicated PR).

orelmisan:
> > Note: This change breaks the current interface provided by the kiagnose library.
> 
> How does it breaks?

It changes the [name and the type](https://github.com/kiagnose/kiagnose/pull/220/commits/a20490c698b46eda0c82110188b5938a95ec9296#diff-90979d03084d8a6305182107a57289abe017d1855a59ee1a47ed81138e3febbdL42) of `config.Config{}`, which is an interface between the checkup and the library.

orelmisan:
Rebase.

ormergi (approved):

RamLavi (approved):
/lgtm

EdDev (approved):
Thanks

PR #225, opened 2022-11-28
vmlatency, main: Fix variable shadowing

orelmisan:
Currently, the second `err` variable is shadowing the first. Use a single `err` variable instead of two.


ormergi (approved):
Please note that when `err` is assigned the program will exit immediately and so the correct error message will propagate, there will be no actual shadowing.

Nevertheless, it seems to make the code more consistent.

RamLavi (approved):
/lgtm

EdDev (approved):
I think it is common to shadow `err`, it does not raise any questions about logic when it is done.

While I think this is a redundant change, I also do not mind if it gets in.

PR #226, opened 2022-11-28
vmlatency, Dockerfile: Use Entrypoint instead of CMD

orelmisan:
Currently, the Dockerfile is using the `CMD` stanza, which is supposed to provide defaults for an executing container [1].

Use the `ENTRYPOINT` stanza instead, which is used to configure a container that will run as an executable [2].

[1] https://docs.docker.com/engine/reference/builder/#cmd 
[2] https://docs.docker.com/engine/reference/builder/#entrypoint


ormergi (approved):

RamLavi (approved):
/lgtm

EdDev (approved):

PR #227, opened 2022-11-28
vmlatency, config: Fix variable shadowing

orelmisan:
Currently, the `err` variable which holds the returned value from `newConfig.validate()` is shadowing a previous declaration.

Fix the variable shadowing.

### DIFF
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
@@ -91,7 +91,8 @@ func New(baseConfig kconfig.Config) (Config, error) {
 	}
 	newConfig.DesiredMaxLatencyMilliseconds = desiredMaxLatency
 
-	if err := newConfig.validate(); err != nil {
+	err = newConfig.validate()
+	if err != nil {
 		return Config{}, err
 	}
 


checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
```diff
@@ -91,7 +91,8 @@ func New(baseConfig kconfig.Config) (Config, error) {
 	}
 	newConfig.DesiredMaxLatencyMilliseconds = desiredMaxLatency
 
-	if err := newConfig.validate(); err != nil {
+	err = newConfig.validate()
+	if err != nil {
```
ormergi, line 95:
```suggestion
	if err = newConfig.validate(); err != nil {
```

ormergi (approved):
Please note that when `err` is assigned `New()` returns immediately and so the correct error message will propagate, there will be no actual shadowing.

checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
orelmisan: This was my first choice.
The linter fails it.

RamLavi (approved):
/lgtm

EdDev (approved):
This is redundant, the original code was fine.
Said that, I do not really mind, it is a nit.

PR #230, opened 2022-12-07
vmlatency, client: Add context to client layer

orelmisan:
Currently, functions from the client layer does not accept a context as a parameter, thus making them un-cancelable from the outside.

Add a context parameter to applicable functions from the client layer, in order to be able to cancel them from the outside.

This will be used in a follow-up PR to enforce the timeout parameter specified by the user.

### DIFF
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
@@ -117,6 +117,8 @@ func (c *Client) SerialConsole(namespace, vmiName string, timeout time.Duration)
 	return c.KubevirtClient.VirtualMachineInstance(namespace).SerialConsole(vmiName, &kubecli.SerialConsoleOptions{ConnectionTimeout: timeout})
 }
 
-func (c *Client) GetNetworkAttachmentDefinition(namespace, name string) (*netattdefv1.NetworkAttachmentDefinition, error) {
-	return c.K8sCniCncfIoV1Interface.NetworkAttachmentDefinitions(namespace).Get(context.Background(), name, metav1.GetOptions{})
+func (c *Client) GetNetworkAttachmentDefinition(
+	ctx context.Context,
+	namespace, name string) (*netattdefv1.NetworkAttachmentDefinition, error) {
+	return c.K8sCniCncfIoV1Interface.NetworkAttachmentDefinitions(namespace).Get(ctx, name, metav1.GetOptions{})
 }
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
@@ -57,8 +57,25 @@ func New() (*Client, error) {
 	return &Client{c, cniClient}, nil
 }
 
-func (c *Client) GetVirtualMachineInstance(namespace, name string) (*kvcorev1.VirtualMachineInstance, error) {
-	return c.KubevirtClient.VirtualMachineInstance(namespace).Get(name, &metav1.GetOptions{})
+func (c *Client) GetVirtualMachineInstance(ctx context.Context, namespace, name string) (*kvcorev1.VirtualMachineInstance, error) {
+	type resultWrapper struct {
+		vmi *kvcorev1.VirtualMachineInstance
+		err error
+	}
+
+	resultCh := make(chan resultWrapper, 1)
+
+	go func() {
+		vmi, err := c.KubevirtClient.VirtualMachineInstance(namespace).Get(name, &metav1.GetOptions{})
+		resultCh <- resultWrapper{vmi, err}
+	}()
+
+	select {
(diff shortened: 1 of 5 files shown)


RamLavi (commented):
question, why is this important to us?
why not wait for https://github.com/kubevirt/kubevirt/issues/8885 to be approved?

orelmisan:
> question, why is this important to us? why not wait for [kubevirt/kubevirt#8885](https://github.com/kubevirt/kubevirt/issues/8885) to be implemented?

Currently, the user-supplied `timeout` field it not respected.
In order to respect it, we need to pass a context to long running functions.

There is an open PR of the required changes to VMI methods in KubeVirt's client-go: https://github.com/kubevirt/kubevirt/pull/8908

The change proposed in this PR is a temporary workaround, until the clean fix will be merged and released so we can consume it.

orelmisan:
Rebase

checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
```diff
@@ -79,9 +79,22 @@ func (c *Client) GetVirtualMachineInstance(ctx context.Context, namespace, name
...
+		createdVMI, err := c.KubevirtClient.VirtualMachineInstance(namespace).Create(vmi)
+		resultCh <- resultWrapper{createdVMI, err}
+	}()
+
+	select {
+	case result := <-resultCh:
+		return result.vmi, result.err
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	}
```
ormergi, line 97: I see this logic reaping, how about reusing it by extracting it to function?
For example: 
```
func execWithContext(ctx context.Context, fn func() (any, error)) (any, error) {
	type result struct {
		value any
		err   error
	}

	resultCh := make(chan result, 1)

	go func() {
		v, err := fn()
		resultCh <- resultWrapper{v, err}
	}()

	select {
	case result := <-resultCh:
		return result.value, result.err
	case <-ctx.Done():
		return nil, ctx.Err()
	}
}
```
Usage:
```
func (c *Client) CreateVirtualMachineInstance(ctx context.Context, namespace string, vmi *kvcorev1.VirtualMachineInstance) (*kvcorev1.VirtualMachineInstance, error) {
return execWithContext(ctx, func() (any, error) {
	return c.KubevirtClient.VirtualMachineInstance(namespace).Create(vmi)
})
```

orelmisan: This is a nice suggestion, thank you.
The parameter types of `Get()` and `Create()` are different.
Do you have a suggestion on how to bypass this problem?

EdDev: If I am not mistaken (please correct me here if I am), the usage of `any` is not for generics, but for pure interfaces. A solution that will not be type safe and require conversion at the caller.

orelmisan: `any` is an alias for `interface{}`.

checkups/kubevirt-vm-latency/vmlatency/internal/vmi/operation.go
```diff
@@ -35,16 +35,19 @@ import (
...
 
 type KubevirtVmisClient interface {
-	GetVirtualMachineInstance(namespace, name string) (*kvcorev1.VirtualMachineInstance, error)
-	CreateVirtualMachineInstance(namespace string, vmi *kvcorev1.VirtualMachineInstance) (*kvcorev1.VirtualMachineInstance, error)
-	DeleteVirtualMachineInstance(namespace, name string) error
+	GetVirtualMachineInstance(ctx context.Context, namespace, name string) (*kvcorev1.VirtualMachineInstance, error)
+	CreateVirtualMachineInstance(
+		ctx context.Context,
+		namespace string,
+		vmi *kvcorev1.VirtualMachineInstance) (*kvcorev1.VirtualMachineInstance, error)
```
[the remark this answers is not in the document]
orelmisan: Thanks for the suggestion.
I personally prefer to have the parameters names in order to have more context.
This multi-line declaration is indeed ugly, but in the balance between pretty and less informative, and ugly and informative, I choose the later.

EdDev (approved):
Very nice, thank you!

checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
EdDev: Yes, and it was aimed for generics and not interfaces. Otherwise, it is hard to follow the intention.
Bottom line, use `any` when intending to use generics and `interface{}` when intending to use interfaces to reduce confusions.

ormergi: >  Do you have a suggestion on how to bypass this problem?

I need to think about it more, lets's continue it on a follow-up PR.

PR #228, opened 2022-11-29
vmlatency:  Align 'spec' and 'status' field names format 

ormergi:
Currently, the status fields and params fields are in different formats, params fields are snake case and the status fields are camel case.

Change the params fields to be aligned with status fields camel case format, its common format in k8s eco-system.

The current changes do not remove the old form parameters completely and they **can** be used.
The old form parameters should be removed when deprecated completely.

PR #232, opened 2022-12-08
vmlatnecy, e2e: Use parameters new form

ormergi:
Following #228, update the e2e tests to use the new parameters form.


orelmisan (commented):
Could you please add a little commit message and PR description with a reference to the relevant PR you are following up? (so our future selves will have some context when reading this commit)

orelmisan (dismissed):
Thanks.

orelmisan (approved):

PR #236, opened 2022-12-08
vmlatency, client: Export context wrap code to func

ormergi:

### DIFF
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
@@ -38,11 +38,6 @@ type Client struct {
 	netattdefclient.K8sCniCncfIoV1Interface
 }
 
-type resultWrapper struct {
-	vmi *kvcorev1.VirtualMachineInstance
-	err error
-}
-
 func New() (*Client, error) {
 	kubeconfig, err := rest.InClusterConfig()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


checkups/kubevirt-vm-latency/vmlatency/internal/client/client.go
```diff
@@ -63,54 +58,25 @@ func New() (*Client, error) {
...
-	select {
-	case err := <-resultCh:
-		return err
-	case <-ctx.Done():
-		return ctx.Err()
-	}
+	_, err := execWithContext(ctx, func() (*kvcorev1.VirtualMachineInstance, error) {
+		return nil, c.KubevirtClient.VirtualMachineInstance(namespace).Delete(name, &metav1.DeleteOptions{})
+	})
+	return err
```
orelmisan, line 79: The VirtualMachineInstance().Delete() method returns only one value.

Using `execWithContext()` on it seems forced to me.

orelmisan (commented):
The suggested change does make the code shorter, but I think it also makes it harder to understand.
It makes use of two levels of closures, which IMO is harder to follow than the existing implementation that only uses one closure.

@EdDev, @RamLavi WDYT?

EdDev:
> Thank you for the PR @ormergi. The suggested change does make the code shorter, but I think it also makes it harder to understand. It makes use of two levels of closures, which IMO is harder to follow than the existing implementation that only uses one closure.
> 
> @EdDev, @RamLavi WDYT?

I like the direction.

It is a nice challenge to extract the context wrapping logic out of the specific calling function.
The extraction here seems to me incomplete, leaving a strong dependency on what is called.

I am not sure what two levels of closures we have here, I think there is only one.
But I do not think by itself it is a problem, Go likes to wrap things this way and we see it done a lot in helper libs to generalize and decouple logic.

I thought to suggest an alternative here, but it required me to write the whole thing anyway, so I just posted #237 . Maybe you can check it and possibly change this PR in a similar way.

ormergi:
@EdDev 

> It is a nice challenge to extract the context wrapping logic out of the specific calling function. The extraction here seems to me incomplete, leaving a strong dependency on what is called.
> 

I left it like this (coupled with vmi package) since there is no other usages at the moment, but I do like the idea of exporting it to self-contained package.

> I thought to suggest an alternative here, but it required me to write the whole thing anyway, so I just posted #237 . Maybe you can check it and possibly change this PR in a similar way.

Thanks for posting an example, I do like this idea actually I had the exact iteration when working on it.
I didn't like the required "boilerplate code" (defining the variables in advance, handling both context and the delegated func error) and wanted to have the smallest footprint, this is why I changed it eventually.
It seems that the benefits are worth it, and seeing this example suggested convinced me to continue with it among other options.

I can do these changes to this PR if you like, or would you prefer to continue at #237?

PR #238, opened 2023-01-22
configmap: Add Create option to library

RamLavi:
This option will be needed for the kubevirt-dpdk-checkup.


RamLavi:
no need for this yet. closing

PR #239, opened 2023-01-23
vmlatency, checkup, setup: Delete VMI/s if setup fails

EdDev:
In scenarios where the setup step fails, the teardown step will not be triggered.

The setup should handle its own cleanup, i.e. deleting the VMI/s created as a best effort attempt.

~~Depends on #240~~

### DIFF
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
@@ -22,6 +22,7 @@ package checkup
 import (
 	"context"
 	"fmt"
+	"log"
 	"strings"
 	"time"
 
@@ -72,7 +73,7 @@ const (
 	LabelLatencyCheckUID = "latency-check/uid"
 )
 
-func (c *checkup) Setup(ctx context.Context) error {
+func (c *checkup) Setup(ctx context.Context) (setupErr error) {
 	const errMessagePrefix = "setup"
 
 	netAttachDef, err := c.client.GetNetworkAttachmentDefinition(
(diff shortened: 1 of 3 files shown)


checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
```diff
@@ -92,9 +93,33 @@ func (c *checkup) Setup(ctx context.Context) error {
 	if err = vmi.Start(ctx, c.client, c.namespace, sourceVmi); err != nil {
 		return fmt.Errorf("%s: %v", errMessagePrefix, err)
 	}
+	const setupCleanupTimeout = 30 * time.Second
+	defer func() {
```
ormergi, line 97: How about defining the a func for the defered cleanup logic and reusing it for both VMs?
Example:
```
cleanUpVM := func(vmiName string) {}
vmi.Start(ctx, c.client, c.namespace, sourceVmi)
defer func(){ 
    if setupErr != nil {
        cleanUpVMI(sourceVmi.Name)
    }
}
vmi.Start(ctx, c.client, c.namespace, targetVmi)
defer func(){ 
    if setupErr != nil {
        cleanUpVMI(targetVmi.Name)
    }
}
```

test/e2e/vmlatency/vmlatency_test.py
```diff
@@ -60,6 +63,41 @@ def test_successful_run(kclient, checkup_sa, checkup_role_binding, configmap_rol
...
+        name="vmlatency-checkup-test",
+        namespace=namespace,
+        data={
+            "spec.timeout": "2m",
+            "spec.param.networkAttachmentDefinitionNamespace": nad.namespace,
+            "spec.param.networkAttachmentDefinitionName": nad.name,
+            "spec.param.maxDesiredLatencyMilliseconds": "500",
+            "spec.param.sampleDurationSeconds": "5",
+            "spec.param.sourceNode": "kind-worker",
+            "spec.param.targetNode": "no-such-worker",
```
ormergi, line 78: This test expects the target VM creation to fail due to a not existing node, but would 2 minutes timeout exceed the overall time of the setup phase?
In other words, is there a scenario that this test will pass even though the setup phase did not finish?

checkups/kubevirt-vm-latency/vmlatency/internal/checkup/checkup.go
EdDev: Good idea, thanks.
Done.

test/e2e/vmlatency/vmlatency_test.py
EdDev: The test waits for `"false" == cm.instance.data.get("status.succeeded")`, otherwise will fail by raising an exception (`TimeoutError`).

EdDev:
[change](https://github.com/kiagnose/kiagnose/compare/f23d5fa3685ad1309cf86a838965c7ca614fd82e..8b95c0ef516095da45f253239a2139de4cc3b80b): Answered suggestion in the review.

EdDev:
change: Rebase

orelmisan (approved):
Thank you

ormergi (approved):
Thanks for the changes

PR #240, opened 2023-01-24
tests, e2e: Fix the checkup image tag

EdDev:
The e2e tests have been running with the wrong image (the one from main) instead of the one build from the source-code.


orelmisan (approved):
Thank you

PR #243, opened 2023-01-26
vmlatency: Fix latency comparison

orelmisan:
Currently, if the actual measured latency is less than 1ms, it is considered as 0ms.
This created a small bug[1] when the max desired latency was 0ms.

Improve the comparison of the max desired latency and the actual measured latency.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2156392

### DIFF
--- a/checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
+++ b/checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
@@ -25,6 +25,7 @@ import (
 	"log"
 	"math"
 	"strconv"
+	"time"
 
 	kconfig "github.com/kiagnose/kiagnose/kiagnose/config"
 )
@@ -56,7 +57,7 @@ type Config struct {
 	TargetNodeName                       string
 	SourceNodeName                       string
 	SampleDurationSeconds                int
-	DesiredMaxLatencyMilliseconds        int
+	DesiredMaxLatency                    time.Duration
 }
 
 var (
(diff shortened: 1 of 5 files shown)


checkups/kubevirt-vm-latency/vmlatency/internal/config/config.go
```diff
@@ -95,9 +96,13 @@ func New(baseConfig kconfig.Config) (Config, error) {
...
 
 	if v := readConfig(baseConfig.Params, DesiredMaxLatencyMillisecondsParamName, DesiredMaxLatencyMillisecondsDeprecatedParamName); v != "" {
-		if newConfig.DesiredMaxLatencyMilliseconds, err = strconv.Atoi(v); err != nil {
+		var rawDesiredMaxLatencyMilliseconds int
+
+		if rawDesiredMaxLatencyMilliseconds, err = strconv.Atoi(v); err != nil {
 			return Config{}, fmt.Errorf("%q parameter is invalid: %v", DesiredMaxLatencyMillisecondsParamName, err)
 		}
+
+		newConfig.DesiredMaxLatency = time.Duration(rawDesiredMaxLatencyMilliseconds) * time.Millisecond
```
ormergi, line 105: How about using the built-in ParseDuration?
```
cost MillisecondUnit= "ms"
desiredMaxLatency, err := time.ParseDuration(rawDesiredMaxLatencyMilliseconds + MillisecondUnit)
```

orelmisan: This is a great idea (that maybe could be implemented in the other checkups as well).
The problem is that it requires more changes to the unit test.
I prefer to do it in a follow-up PR.

RamLavi: +1 for the follow up PR

RamLavi (approved):
/lgtm

PR #244, opened 2023-03-01
vmlatency, Dockerfile: Update base image to ubi-minimal9

orelmisan:
In order to address security issues discovered in the current base image, and to get the latest ubi image, update the base image to ubi-minimal 9[1].

Also add `-y` flags to microdnf commands, so they will not hang.

[1] https://catalog.redhat.com/software/containers/ubi9/ubi-minimal/615bd9b4075b022acc111bf5


RamLavi (approved):
/lgtm

PR #252, opened 2024-01-08
VM Latency: Bump base image

orelmisan:
Bump the base image to:
`registry.access.redhat.com/ubi9/ubi-minimal:9.3-1475`

which is the current latest.


RamLavi (approved):
/lgtm

